PR #170, opened 2016-07-26
use default Fedora's httpd.conf

ignatenkobrain:


jsilhan:
@j-mracek , what do you think? can it break something?

j-mracek:
@ignatenkobrain @jsilhan I think it should not break anything. I think we need and use default settings of httpd.

Issue #161, opened 2016-06-27
gather test names during the provision


ignatenkobrain:
otherwise if something got removed after provision, but before test job - something will crash

ignatenkobrain:
@j-mracek it's easy to fix. I think it can be useful for you to understand JJB a bit.

ignatenkobrain:
Fixed as part of #171.

PR #171, opened 2016-07-26
add helper to operate on CI

ignatenkobrain:


ignatenkobrain:
as side effect - fixes #161, but still can be improved.

PR #181, opened 2016-08-30
repoquery-1: expect md expiration msg in stderr

ignatenkobrain:
In dnf#a9f70757b7d2f23e48b265662ba95a8dae202730 we moved message about
metadata expiration to the stderr as it's not supposed to be parseable.


ignatenkobrain:
CC @MichaelMraka

P.S. I didn't have time to test it.

ignatenkobrain:
```
      Assertion Failed: The line starting with u'Last metadata expiration check' was found in u'stderr', but should be absent
      Captured stdout:
      Last metadata expiration check: 17043 days, 13:51:07 ago on Thu Jan 01 00:00:00 1970 UTC.
      Cache-only enabled but no cache for 'test-1', disabling.
```
@j-mracek

PR #182, opened 2016-09-22
Enhance local-build performance

j-mracek:
It do not requires to install all dnf components from rpms/


Dockerfile
```diff
@@ -9,10 +9,12 @@ RUN dnf -y update
 # enum34: rpmdb State
 # whichcraft: shutil.which() for py2
 # jinja2: rpmspec template
-RUN dnf -y install httpd python2-behave python2-six python-enum34 https://kojipkgs.fedoraproject.org//packages/python-whichcraft/0.4.0/2.fc24/noarch/python2-whichcraft-0.4.0-2.fc24.noarch.rpm python-jinja2
+RUN dnf -y install dnf-plugins-core python3-dnf-plugins-core python2-dnf-plugins-core httpd python2-behave python2-six python-enum34 https://kojipkgs.fedoraproject.org//packages/python-whichcraft/0.4.0/2.fc24/noarch/python2-whichcraft-0.4.0-2.fc24.noarch.rpm python-jinja2
 COPY dnf-docker-test/repo /var/www/html/repo/
 COPY dnf-docker-test/features /behave/
 
+# Allows to run test with rpms from only single component in rpms/
+RUN dnf -y copr enable rpmsoftwaremanagement/dnf-nightly
```
ignatenkobrain: it's bad idea for CI to do this as we always build full tree of components and if we use COPR repo we can get different results

Dockerfile
```diff
@@ -9,10 +9,12 @@ RUN dnf -y update
 # enum34: rpmdb State
 # whichcraft: shutil.which() for py2
 # jinja2: rpmspec template
-RUN dnf -y install httpd python2-behave python2-six python-enum34 https://kojipkgs.fedoraproject.org//packages/python-whichcraft/0.4.0/2.fc24/noarch/python2-whichcraft-0.4.0-2.fc24.noarch.rpm python-jinja2
+RUN dnf -y install dnf-plugins-core python3-dnf-plugins-core python2-dnf-plugins-core httpd python2-behave python2-six python-enum34 https://kojipkgs.fedoraproject.org//packages/python-whichcraft/0.4.0/2.fc24/noarch/python2-whichcraft-0.4.0-2.fc24.noarch.rpm python-jinja2
```
ignatenkobrain: I think here should go only packages which we explicitly need for testing framework, not for testing itself.

j-mracek:
@ignatenkobrain Igor, please be more flexible to needs of other people. I want to use the framework to test my new PR on dnf. I don't want to build all components, only to test with dnf. Also as you talk about different results, different test approaches can discover more problems.

Dockerfile
j-mracek: It used this approach before, lets think about revert it. It makes life more easy for people that want only test changes in single package locally. I want to increase support for local test functions, because now the functionality is limited.

Dockerfile
j-mracek: This three packages are required for testing (dnf-plugins-core is not part of docker image) and to enable copr repository. It is really important to enlarge capability for local testing. Relay on only copr build option is not strategic.

Dockerfile
ignatenkobrain: When we will install packages for `/rpms/` then it will mess up which is not what we want.

About single packages... You can download latest RPMs from COPR, put them into `rpms/` and replace packages which you want. This way we ensure that nothing will break your test environment.

Dockerfile
ignatenkobrain: Those are required for testing itself, not for testing framework (note that there are plans for testing PackageKit as well and it doesn't need anything like that).

j-mracek:
@ignatenkobrain I refactored it. What do you think?

dnf-testing.sh
```diff
@@ -65,6 +65,17 @@ if [ "$action" = "run" ]; then
         TESTS+=("$arg")
         shift
     done
+elif [ "$action" = "build" ]; then
+    type="local"
+    for arg; do
```
ignatenkobrain: I would prefer to have it in option style (`-t/--type`)

dnf-testing.sh
```diff
@@ -92,6 +103,11 @@ list()
 
 build()
 {
+    if [ "$type" = "jjb" ]; then
+        cp Dockerfile.jjb Dockerfile
```
ignatenkobrain: symlink is better than copy.

Dockerfile.local
```diff
@@ -0,0 +1,27 @@
...
+# whichcraft: shutil.which() for py2
+# jinja2: rpmspec template
+RUN dnf -y install dnf-plugins-core python3-dnf-plugins-core python2-dnf-plugins-core httpd python2-behave python2-six python-enum34 https://kojipkgs.fedoraproject.org//packages/python-whichcraft/0.4.0/2.fc24/noarch/python2-whichcraft-0.4.0-2.fc24.noarch.rpm python-jinja2
+COPY dnf-docker-test/repo /var/www/html/repo/
+COPY dnf-docker-test/features /behave/
+
+# Allows to run test with rpms from only single component in rpms/
+RUN dnf -y copr enable rpmsoftwaremanagement/dnf-nightly
+COPY rpms /rpms/
+RUN dnf -y install /rpms/*.rpm
```
ignatenkobrain: this will fail if there is no RPMs, so it should be conditional

Dockerfile.local
```diff
@@ -0,0 +1,27 @@
...
+
+RUN echo "deltarpm=0" >> /etc/dnf/dnf.conf
+RUN dnf -y update
+# behave: core
+# httpd: old-style repos
+# six: py2/py3
+# enum34: rpmdb State
+# whichcraft: shutil.which() for py2
+# jinja2: rpmspec template
+RUN dnf -y install dnf-plugins-core python3-dnf-plugins-core python2-dnf-plugins-core httpd python2-behave python2-six python-enum34 https://kojipkgs.fedoraproject.org//packages/python-whichcraft/0.4.0/2.fc24/noarch/python2-whichcraft-0.4.0-2.fc24.noarch.rpm python-jinja2
```
ignatenkobrain: I think here should go only test requirements and then we should install all packages from nightlies.

dnf-testing.sh
j-mracek: The human will not use any argument in the option, only jjb.

Dockerfile.local
ignatenkobrain: I meant that container will not be built.

j-mracek: I know and the fail here helps me to understand that my /rpms dir with rpms is not in correct position.

j-mracek:
@ignatenkobrain No idea what and why should be better.

mluscon:
@dnf-bot r+

mluscon:
Merged. The requested improvements will be addressed on the way.

PR #188, opened 2016-10-09
Allow run ci-dnf -stack with dnf-2 and dnf-3

j-mracek:
It works for local and hopefully for jenkins jobs

### DIFF
--- a/dnf-testing.sh
+++ b/dnf-testing.sh
@@ -136,17 +136,23 @@ run()
     [ ${#TESTS[@]} -eq 0 ] && TESTS=("${FEATURES[@]}")
     local failed=0
     local failed_test_name='Failed test(s):'
-    for feature in "${TESTS[@]}"; do
-        if [ -z "$devel" ];then
-            printf "\nsudo docker run --rm "$IMAGE" launch-test "$feature" dnf\n"
-            sudo docker run --rm "$IMAGE" launch-test "$feature" dnf >&2 || \
-            if [ $? -ne 0 ]; then let ++failed && failed_test_name+=" $feature"; fi
-        else
-            printf "\nsudo docker run --rm -v "$devel" "$IMAGE" launch-test "$feature" dnf\n"
-            sudo docker run --rm -v "$devel" "$IMAGE" launch-test "$feature" dnf >&2 || \
-            if [ $? -ne 0 ]; then let ++failed && failed_test_name+=" $feature"; fi
-        fi
-    done
+    if [ -z "$devel" ];then
+        for feature in "${TESTS[@]}"; do
+            for command in dnf-2 dnf-3; do
+                printf "\nsudo docker run --rm "$IMAGE" launch-test "$feature" $command\n"
+                sudo docker run --rm "$IMAGE" launch-test "$feature" $command >&2 || \
+                if [ $? -ne 0 ]; then let ++failed && failed_test_name+=" $feature-$command"; fi
+            done
+        done
+    else
+        for feature in "${TESTS[@]}"; do
+            for command in dnf-2 dnf-3; do
+                printf "\nsudo docker run --rm -v "$devel" "$IMAGE" launch-test "$feature" $command\n"
+                sudo docker run --rm -v "$devel" "$IMAGE" launch-test "$feature" $command >&2 || \
+                if [ $? -ne 0 ]; then let ++failed && failed_test_name+=" $feature-$command"; fi
+            done
+        done
+    fi
     if [ "$failed" != 0 ]; then
         >&2 echo "$failed_test_name"
     fi
--- a/jobs/github-testers-macros.yaml
+++ b/jobs/github-testers-macros.yaml
@@ -261,12 +261,13 @@
           #!/bin/bash
           set -xeuo pipefail
 
+          PREFIXES=('dnf-2' 'dnf-3')
           mkdir test-results/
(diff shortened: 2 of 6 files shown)


dnf-testing.sh
```diff
@@ -110,8 +110,10 @@ run()
     [ ${#TESTS[@]} -eq 0 ] && TESTS=("${FEATURES[@]}")
     local failed=0
     for feature in "${TESTS[@]}"; do
-        sudo docker run --rm "$IMAGE" launch-test "$feature" dnf >&2 || :
-        [ $? -ne 0 ] && let ++failed
+        for command in dnf-2 dnf-3; do
```
ignatenkobrain: basically it will run dnf2 and dnf3 in one container which will double time for testing. We should spawn 2 containers insted.

ignatenkobrain:
create a new job in JJB for the py2/py3 stuff

dnf-testing.sh
j-mracek: @ignatenkobrain I think that the loops run 
sudo docker run --rm "$IMAGE" launch-test "$feature" dnf-2
then:
sudo docker run --rm "$IMAGE" launch-test "$feature" dnf-3
Therefore it is not run in single container. Am I wrong?

j-mracek: I takes twice as time because it is not run in parallel.

j-mracek:
@ignatenkobrain Please can you create a new job in JJB for the py2/py3 stuff? No xp with it.

dnf-testing.sh
ignatenkobrain: @j-mracek yes, exactly. If you do 2 different jenkins jobs, then it's running in parallel.

ignatenkobrain:
@j-mracek Unfortunately I don't have enough time for it. Main problem there is joining jobs and gathering results. There are 2 ways:

* Somehow ask jenkins to wait for another job and combine py2/py3 results and update gh status within one context (IMO this is the proper way)
* Use 2 different github contexts for py2/py3 (but I think homu (@dnf-bot) doesn't support checking for multiple contexts, but this is possible to check)

ignatenkobrain:
quick check in the code for 2nd case -- homu doesn't support waiting for multiple contexts.

j-mracek:
@ignatenkobrain No idea what are you talking about.

ignatenkobrain:
screenshot from 2016-10-09 14-59-31


This thing is called context

j-mracek:
@ignatenkobrain Still not sure whats the problem. The PR should do for jenkins job: test config-1 is run with dnf-2 and dnf-3. During executing the test is renamed into config-1-dnf-2 and config-1-dnf-3, therefore handler of output will see two different tests therefore results are not overwritten. 
But for local test it runs each test with dnf-2 and dnf-3 (each in separate container one by one). It will requires double time but all required test will be performed. 
This PR should fix basic problem that python2 runs were ignored. It doesn't solve any performance issue. It also allows to run test with [yum, dnf-2, dnf-3]. The step one is that tests have to be performed. Then we can talk about performance.

ignatenkobrain:
@j-mracek problem is that it's already slow and it will not help in the future (as you said adding `yum`), because it will triple time and etc. There are some plans for PackageKit to be added, so it will be 4x times more.

I was thinking for a while about this as for now and I think proper way will be to use 2 different contexts for dnf-2 and dnf-3 and write some patch for homu which will allow specifying > 1 context (but this can be ignored for now). So can you make jenkins to spawn 2 different `*-test` jobs with py2/py3 accordingly?

j-mracek:
@ignatenkobrain Still don't get the point. Jenkins runs docker instances in parallel, therefore there is any difference if I run 2X jobs in parallel (one job that runs all jobs in parallel) in comparison with 2x X in parallel (2 jobs that runs containers in parallel).

ignatenkobrain:
1. With this PR you don't run tests in parallel (at all)
2. It's something what jenkins should do
3. After this PR you will get something like `100/100 tests passed`, but actually there are only 50 tests and it will mess up jenkins JUnit report

jobs/github-testers-macros.yaml
```diff
@@ -265,7 +265,7 @@
           TEST_DIR=`readlink -f test-results`
           chcon -Rt svirt_sandbox_file_t "$TEST_DIR"
           TESTS=(${{TESTS[@]}})
-          parallel -j100% sudo docker run --rm -v "$TEST_DIR":/junit "$CONTAINER" launch-test "{{}}" dnf ::: "${{TESTS[@]}}" || :
+          parallel -j100% sudo docker run --rm -v "$TEST_DIR":/junit "$CONTAINER" launch-test "{{}}" ::: dnf-2 dnf-3 ::: "${{TESTS[@]}}" || :
```
j-mracek: @ignatenkobrain Here is parallel trigger, I only extended it. Therefore jenkins do parallel run!!!

dnf-testing.sh
j-mracek: @ignatenkobrain This is local shell script - it runs test one by one. The behavior was not changed, only number of test is extended. Parallel run could be introduced but not by the PR.

mluscon:
May I give the green light to this PR @jsilhan, or do you want to wait for a better solution?

jsilhan:
This is improvement of current situation and does not have drawbacks so it could be merged. there was a lot of time to come up with better solution. Once we have better design we can replace it.

PR #189, opened 2016-10-11
Separate docker run from network during jenkins run

j-mracek:


j-mracek:
There is problem with test plugin-path-1 because it runs "dnf --installroot=/dockertesting copr list rpmsoftwaremanagement" and "dnf copr list rpmsoftwaremanagement"

ignatenkobrain:
@j-mracek hmm. We should not test COPR that way.

jobs/github-testers-macros.yaml
```diff
@@ -265,8 +265,8 @@
           TEST_DIR=`readlink -f test-results`
           chcon -Rt svirt_sandbox_file_t "$TEST_DIR"
           TESTS=(${{TESTS[@]}})
-          parallel -j100% sudo docker run --rm -v "$TEST_DIR":/junit "$CONTAINER" launch-test "{{}}" dnf ::: "${{TESTS[@]}}" || :
-          JSON=$(sudo docker run --rm "$CONTAINER" behave-2 --dry-run --no-summary -f json /behave/)
+          parallel -j100% sudo docker run --net=isolated --rm -v "$TEST_DIR":/junit "$CONTAINER" launch-test "{{}}" dnf ::: "${{TESTS[@]}}" || :
+          JSON=$(sudo docker run --net=isolated --rm "$CONTAINER" behave-2 --dry-run --no-summary -f json /behave/)
```
ignatenkobrain: actually this command should not access network, but anyway it will not do any harm.

jobs/github-testers-macros.yaml
```diff
@@ -265,8 +265,8 @@
           TEST_DIR=`readlink -f test-results`
           chcon -Rt svirt_sandbox_file_t "$TEST_DIR"
           TESTS=(${{TESTS[@]}})
-          parallel -j100% sudo docker run --rm -v "$TEST_DIR":/junit "$CONTAINER" launch-test "{{}}" dnf ::: "${{TESTS[@]}}" || :
-          JSON=$(sudo docker run --rm "$CONTAINER" behave-2 --dry-run --no-summary -f json /behave/)
+          parallel -j100% sudo docker run --net=isolated --rm -v "$TEST_DIR":/junit "$CONTAINER" launch-test "{{}}" dnf ::: "${{TESTS[@]}}" || :
```
ignatenkobrain: it should be `--net=none`, but I would also to add `--icc-false` to disallow any inter-container communication.

README.md
```diff
@@ -61,6 +61,9 @@ jenkins ALL=(ALL) NOPASSWD: /usr/bin/docker
 EOF
 ```
 
+docker need to be configured by command:
```
ignatenkobrain: nah, that's not true if you use `--net=none`

j-mracek:
Anyway we should disable auto test-run for people without rights until solution will be ready.

jobs/github-testers-macros.yaml
j-mracek: --icc-false option doesn't exist for docker run. The presented solution do both, no network, and disallow any inter-container communication.

j-mracek: @ignatenkobrain @mluscon @jsilhan @MichaelMraka  @mhatina I found the solution to disable networking and any inter-container communication for ours CI, but we have test that use 'dnf copr' command (test plugin-path-1) that requires internet connection, therefore I strongly recommend other disused solution (disable auto test for PRs from untrusted source) until we will be able to fix the test with copr.

README.md
j-mracek: I am closing the PR, because there is a conflict with test plugin-path-1 and until it is not fixed we have to go with other solution.

jobs/github-testers-macros.yaml
MichaelMraka: What about:
- mark tests as networking/no-networking so only containers running networking tests have it enabled?
- limit outgoing network in containers, e.g. by firewall, only to copr?

PR #199, opened 2016-11-24
jobs: stop doing nightlies for F23

ignatenkobrain:
It's outdated and will be EOL in 1 month. Since officially DNF is coming
only in F26 it's fine to stop producing RPMs for such system.


jsilhan:
@dnf-bot r+

PR #207, opened 2016-12-15
Shutup bot

ignatenkobrain:
Sending useless messages about verifying the patch doesn't make any sense at all. People hate this. They don't understand why it's needed when they send PR to RPM or libhif or libsolv.

@pmatilai, now people can finally be happy


Conan-Kudo:
@ignatenkobrain does this mean that everyone will have the ability to merge things by issuing commands to dnf-bot? Because it does our merging instead of humans...

ignatenkobrain:
@Conan-Kudo, no it's just disables this pointless messages "Can anyone of admins verify the patch?"

j-mracek:
@ignatenkobrain But Igor you remove test for copr. This is not good idea, the copr is important part of plugins.

ignatenkobrain:
@j-mracek 1) COPR is not service which we run, so if something happens with COPR -- we have no way to fix it 2) it doesn't really test COPR, it just enables some repo (which btw will disappear soon) 3) if you really want to test plugin, you probably want to mock its API reply.

ignatenkobrain:
@j-mracek, decision was made to make secure environment (one option is to disable automatic building of PRs which is total bullshit, second option is to disable networking). Disabling networking is a proper solution.

j-mracek:
@ignatenkobrain I can only accept PR that will enhance copr test or add new scenarios but not remove it.

Conan-Kudo:
For what it's worth, I side with @ignatenkobrain here. The current commands against COPR are functionally useless. It would probably be better to do some mocking instead. Disabling networking makes for a much more secure environment, too. I'm very surprised you're running CI tests with networking on, as that's a very different environment from how Koji would run the same tests.

j-mracek:
@ignatenkobrain I suggest:
1. change copr test to cover more functionality without network requirement
2. Disable network but keep restriction for automatically run tests for everyone

ignatenkobrain:
> change copr test to cover more functionality without network requirement
it's not related to issue which this PR is fixing.

> Disable network but keep restriction for automatically run tests for everyone
That's totally useless and goes against what this PR is fixing.

MichaelMraka:
@j-mracek well, the most common / used way to test such thing is to alias copr.fedorainfracloud.org to 127.0.0.1 and start a local httpd server with appropriate paths created.

jsilhan:
I am fine with dismissing `copr list` test - it's not core functionality of copr plugin and it could be tested differently.

PR #210, opened 2017-01-03
Image size reduction

j-mracek:


j-mracek:
@ignatenkobrain The proposed change mimics very common user behavior when user upgrade only dnf or dnf first. If dnf is incapable to work  in such conditions we did really something incorrect.

MichaelMraka:
> when you don't update whole image, it means you test GA version + our updated packages.
> Not gonna happen.

The other way round - that's what users can do, install system and upgrade just one (dnf) package, and  we definitely have to support it. If some upgraded packages are necessary they must be properly required thus will be updated in tests.

ignatenkobrain:
> The other way round - that's what users can do, install system and upgrade just one (dnf) package, and we definitely have to support it. If some upgraded packages are necessary they must be properly required thus will be updated in tests.

How often people do reinstall of their system? I would say that 80% of cases that people have fully updated system, 10% that they have half-updated system and 10% that they have fresh system.

And actually I was not pointing on DNF requires. Workability of DNF depends not only on its requires. For instance, take more-or-less breakage of pygpgme. With GA version of gnupg2/pygpgme everything works, with updated version of gnupg2 it (pygpgme) doesn't work anymore (and this is what users see). That's how it was spotted and fixed by me when CI started to fail, so only limited people had this breakage (we got around 6 bugreports). If we would test with GA version, until people have broken system -- we don't notice.

I would say that we should test both, but primary method should stay *updated system*.

PR #212, opened 2017-01-05
Allow to match regexps in cmd outputs

glum23:
only exact multiline outputs can be matched now, this allows to match also regexps

### DIFF
--- a/dnf-docker-test/features/steps/command_steps.py
+++ b/dnf-docker-test/features/steps/command_steps.py
@@ -46,3 +46,10 @@ def step_the_command_stream_should_contain_exactly(ctx, stream):
 def step_the_command_stream_should_be_empty(ctx, stream):
     ctx.text = ""
     step_the_command_stream_should_contain_exactly(ctx, stream)
+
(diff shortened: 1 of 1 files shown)


dnf-docker-test/features/steps/command_steps.py
```diff
@@ -46,3 +46,10 @@ def step_the_command_stream_should_contain_exactly(ctx, stream):
 def step_the_command_stream_should_be_empty(ctx, stream):
     ctx.text = ""
     step_the_command_stream_should_contain_exactly(ctx, stream)
+
+@then("the command {stream:stdout_stderr} should match regexp")
```
glum23: everywhere in the file there is one blank line in a similar context

ignatenkobrain:
merged with squashing and style fix.

glum23:
It would be nice when dnf-bot doesn't require something (2 empty lines) that you then manually remove

ignatenkobrain:
@glum23 it doesn't require, it's just warnings (which I disabled already, btw) ;)

PR #213, opened 2017-01-06
dnf-testing.sh: ability to reserve a running container

kkaarreell:
This commit adds 1 command and 1 option to ease debugging
 -r, --reserve  to open a bash session in a running container after the test is execution
 shell to simply run a bash session in a container, no tests are executed in this case

### DIFF
--- a/dnf-testing.sh
+++ b/dnf-testing.sh
@@ -20,6 +20,12 @@ set_devel()
     devel="$PROG_PATH/dnf-docker-test/features:/behave:Z"
 }
 
+set_reserve()
+{
+    PARAM_RESERVE="-r"
+    PARAM_TTY="-it"
+}
+
 show_help()
 {
     cat << EOF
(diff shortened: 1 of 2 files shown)


dnf-testing.sh
```diff
@@ -153,3 +163,15 @@ run()
     exit $failed
 }
 [ "$action" = "run" ] && run
+
+shell()
+{
+    if [ -z "$devel" ];then
+        printf "\nsudo docker run -i --rm "$IMAGE" bash\n"
```
ignatenkobrain: No need to print useless stuff :)

kkaarreell: Such prints are present also in the original code, that's why I have added it too.

kkaarreell:
I realized that more changes are needed, running docker with -it is not good when reservation is not needed as terminal decoration characters may/would pollute the output. I will rewrite the script and post a new version.

PR #215, opened 2017-01-06
make shell_commands.py follow CommandResult format

kkaarreell:
This commit changes step_i_run_dnf_shell_command() so it uses CommandResult() object to store the command line output. Therefore newly added functions 
  step_the_command_output_shoud_contain()
  step_the_command_output_shoud_not_contain()
are not needed any more as the previously existing step
  "the command stdout should contain exactly"
could be used.

This commit also updates dnf shell tests accordingly.

### DIFF
--- a/dnf-docker-test/features/steps/shell_commands.py
+++ b/dnf-docker-test/features/steps/shell_commands.py
@@ -3,6 +3,7 @@ from behave import when
 from behave import then
 import pexpect
 import sys
+from command_utils import CommandResult
 
 
 @given('I have dnf shell session opened with parameters "{parameters}"')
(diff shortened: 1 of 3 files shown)


dnf-docker-test/features/steps/shell_commands.py
```diff
@@ -17,25 +18,15 @@ def step_i_have_dnf_shell_session_opened_with_parameters(context, parameters):
...
 @when('I run dnf shell command "{command}"')
 def step_i_run_dnf_shell_command(context, command):
-    context.cmd_result = None
+    context.cmd_result = CommandResult()
     context.assertion.assertNotEqual(None, context.pexpect_session, "dnf shell session must be opened first")
     context.pexpect_session.sendline(command)
     if command.strip() == 'quit' or command.strip() == 'exit':
         context.pexpect_session.expect(pexpect.EOF)
-        context.cmd_result = context.pexpect_session.before
+        context.cmd_result.stdout = context.pexpect_session.before[len(command):].strip()
```
ignatenkobrain: not sure if strip() is good idea because it could remove unwanted newlines or spaces.

kkaarreell: I would need to take a loot at it. The problem is that pexpect is adding extra characters around the message (^M). strip() was an easy method how to get rid of them but I probably need to take a look how it works for multi-line output captured by pexpect.

PR #217, opened 2017-01-10
Adjust downgrade-1 to new dnf behavior

j-mracek:
DNF does not require --allowerasing option to downgrade dependency


mhatina:
@dnf-bot r+

PR #220, opened 2017-01-10
Change xfail setting for few tests

j-mracek:


ignatenkobrain:
@j-mracek description is not enough describing why this is done

j-mracek:
@ignatenkobrain Tests that don't fail - xfails removed, falling added.

PR #225, opened 2017-01-16
Fix problem when --devel mod is used

j-mracek:
It renames feature file after test to original name to not rename local project
files

### DIFF
--- a/dnf-docker-test/launch-test
+++ b/dnf-docker-test/launch-test
@@ -3,5 +3,10 @@ set -xeuo pipefail
 
 httpd -k start
 new_name=${1}-${2}
-mv /behave/${1}.feature /behave/${new_name}.feature
+EXIT_STATUS=0
+cp /tests/${1}.feature /behave/${new_name}.feature || EXIT_STATUS=1
+if [ "$EXIT_STATUS" = 1 ]; then
+    >&2 echo "Feature file ${1}.feature not found"
+    exit 1
+fi
 behave-2 -i $new_name -D dnf_cmd=$2 --junit --junit-directory /junit/ /behave/
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ ARG type=local
 
 COPY dnf-docker-test/repo /var/www/html/repo/
 COPY dnf-docker-test/features /behave/
+COPY dnf-docker-test/features /tests/
 COPY rpms /rpms/
 
 RUN echo -e '\necho "deltarpm=0" >> /etc/dnf/dnf.conf' \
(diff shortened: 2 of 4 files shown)


kkaarreell:
The behave doesn't have to _crash_, it would simply exit with non-zero exit code due to a failing test.

j-mracek:
@ignatenkobrain Refactored. Please can you make very fast review. It is very important.

Dockerfile.local
```diff
@@ -3,6 +3,7 @@ ENV LANG C
 
 COPY dnf-docker-test/repo /var/www/html/repo/
 COPY dnf-docker-test/features /behave/
+COPY dnf-docker-test/features /tests/
```
ignatenkobrain: what's the point to copy it 2 times?

dnf-docker-test/launch-test
```diff
@@ -3,5 +3,10 @@ set -xeuo pipefail
 
 httpd -k start
 new_name=${1}-${2}
-mv /behave/${1}.feature /behave/${new_name}.feature
+EXIT_STATUS=0
+cp /tests/${1}.feature /behave/${new_name}.feature || EXIT_STATUS=1
```
ignatenkobrain: if you want to exit here, just don't do `||`, it doesn't make any sense

Dockerfile.local
j-mracek: ```COPY dnf-docker-test/features /behave/```
Here I need to copy everything except *.feature files (dirs and .*py files)
```COPY dnf-docker-test/features /tests/```
Here I need to copy *.feature files

dnf-docker-test/launch-test
j-mracek: I would like to do two things:
1. Print human understable what went wrong("Feature file ${1}.feature not found") 
2. End return with exit code 1.

Dockerfile.local
j-mracek: Anyway, I see another problem in implementation. Steps are not shared.

ignatenkobrain:
don't really like idea, but since we have to make some fix ASAP, merging.

kkaarreell:
What is the purpose of renaming the feature file anyway? I can imagine that the only reason could be to let user easily find which dnf version was executed but given that the launch-test script prints every command this information is already available.

j-mracek:
@kkaarreell The reason was simple. Tests create .xml output where name is taken from name of behave test. When we want to run dnf with python2 or puthon3, the second test overwrites the xml output from previous.  The reason was to enable test runs with multiple prefixes (dnf-2, dnf-3).

kkaarreell:
I see. Thank you.

PR #227, opened 2017-01-17
add support for http and ftp repository types

kkaarreell:
Updates current step 'Given repository "reponame" with packages' to optionally define a repository type.
Currently implemented types are local (default), http and ftp.
local repo remains under /tmp, http repos are located under /var/www/html, ftp repos under /var/ftp/pub.
vsftpd is added to Dockerfile and started later to serve as ftp server.

file_utils.py have been updated with a new function set_dir_content_ownership() that is used to change file ownership of html or ftp content to respective users (or to a user passed as an argument).

If this gets merged, I will also provide fixes for updateinfo.xml and comps.xml creation where the directory ownership has to be changed to root before calling modifyrepo_c or createrepo_c and later restored, otherwise it won't work properly for http|ftp repositories.

### DIFF
--- a/dnf-docker-test/features/steps/repo_steps.py
+++ b/dnf-docker-test/features/steps/repo_steps.py
@@ -83,6 +83,18 @@ def parse_enable_disable(text):
 
 register_type(enable_disable=parse_enable_disable)
 
+@parse.with_pattern(r"local |http |ftp |")
+def parse_repo_type(text):
+    if text == "http ":
+        return "http"
+    elif text == "ftp ":
+        return "ftp"
+    elif text == 'local ' or text == '':
+        return "file"
+    assert False
+
+register_type(repo_type=parse_repo_type)
+
 @when('I remove all repositories')
 def step_i_remove_all_repositories(ctx):
     """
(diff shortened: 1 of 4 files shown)


dnf-docker-test/features/steps/repo_steps.py
```diff
@@ -83,6 +83,18 @@ def parse_enable_disable(text):
 
 register_type(enable_disable=parse_enable_disable)
 
+@parse.with_pattern(r"local |http |ftp |")
```
ignatenkobrain: I prefer to not have spaces here, but have them in `@given`

dnf-docker-test/features/steps/repo_steps.py
```diff
@@ -165,12 +186,15 @@ def given_repository_with_packages(ctx, repository):
...
+    # set proper directory content ownership
+    file_utils.set_dir_content_ownership(ctx, tmpdir)
+
     repofile = REPO_TMPL.format(repository)
     ctx.table = Table(HEADINGS_INI)
     ctx.table.add_row([repository, "name",     repository])
     ctx.table.add_row(["",         "enabled",  "False"])
     ctx.table.add_row(["",         "gpgcheck", "False"])
-    ctx.table.add_row(["",         "baseurl",  "file://{!s}".format(tmpdir)])
+    ctx.table.add_row(["",         "baseurl",  "{!s}://{!s}".format(rtype, repopath)])
```
ignatenkobrain: I guess this will not work with `local`, since it will put `local://` instead of `file://`

kkaarreell: I am doing s/local/file/ later in the code, the point is to let the user to refer to 'local repository' instead of 'file repository' in the step.

dnf-docker-test/features/steps/repo_steps.py
kkaarreell: That's what I started with but I didn't want to loose the previous/default scenario whe the type is not specified at all.

PR #231, opened 2017-01-19
ability to reserve a running container

kkaarreell:
This commit adds 1 command and 2 options to ease debugging
-r, --reserve - to open a bash session in a running container after every test executed
-R, --reserveonfail -  to open a bash session in a running container upon test failure
shell to simply run a bash session in a container, no tests are executed in this case

### DIFF
--- a/dnf-docker-test/launch-test
+++ b/dnf-docker-test/launch-test
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -xeuo pipefail
 
+RESERVE=0
+if [ "$1" == "-r" -o "$1" == "-R" ]; then
+    RESERVE="$1"
+    shift
+fi
+
 httpd -k start
 vsftpd
 new_name=${1}-${2}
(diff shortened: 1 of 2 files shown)


dnf-docker-test/launch-test
```diff
@@ -10,4 +16,11 @@ if [ "$EXIT_STATUS" = 1 ]; then
     >&2 echo "Feature file ${1}.feature not found"
     exit 1
 fi
-behave-2 -i $new_name -D dnf_cmd=$2 --junit --junit-directory /junit/ /behave/
+
+TEST_EXIT=0
```
ignatenkobrain: this is not needed

kkaarreell: I believe it is due to the bash settings at the beginning of the script.

-u      Treat unset variables and parameters other than the spe-
                      cial  parameters "@" and "*" as an error when performing
                      parameter expansion.  If expansion is  attempted  on  an
                      unset  variable  or parameter, the shell prints an error
                      message, and, if not interactive, exits with a  non-zero
                      status.

j-mracek:
@kkaarreell Thanks for contribution

PR #232, opened 2017-01-19
adds repo enable/disable test for dnf shell

kkaarreell:
Previously created tests are missing the scenario explicitly mentioning 'repo disable' so adding one.


ignatenkobrain:
@kkaarreell don't write private links in public places ;)

kkaarreell:
I will create a new test for failing commands, including pkg installation/removal, group installation/removal and repo enabling/disabling as I expect few of them to be failing.

ignatenkobrain:
@kkaarreell do you want to do this in this PR or another?

kkaarreell:
Preferably I will create a new PR, I don't have the tests ready yet anyway.

PR #234, opened 2017-01-20
add support for building x86_64 and i686 rpms

kkaarreell:
Adds ability to build also x86_64, i386 and i686 rpms with noarch being the default.
The architecture is specified using the Arch tag when describing a package.

### DIFF
--- a/dnf-docker-test/features/steps/repo_steps.py
+++ b/dnf-docker-test/features/steps/repo_steps.py
@@ -29,7 +29,9 @@ Release:        {{ release|default("1") }}%{?dist}
 
 License:        {{ license|default("Public Domain") }}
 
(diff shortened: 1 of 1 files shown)


dnf-docker-test/features/steps/repo_steps.py
```diff
@@ -29,7 +29,7 @@
 
 License:        {{ license|default("Public Domain") }}
 
-BuildArch:      noarch
+BuildArch:      {{ buildarch|default("noarch") }}
```
ignatenkobrain: We should not include any BuildArch if it's != noarch

kkaarreell: fixed

PR #236, opened 2017-01-20
Dockerfile: set tsflags=nodocs

ignatenkobrain:
This should not install any docs which should reduce image size.


MichaelMraka:
Very good idea.

PR #238, opened 2017-01-23
tests for matching expected dnf shell command's output

kkaarreell:
These tests won't pass as dnf currently provides either slightly different output (which I find inconsistent) or no output at all.


ignatenkobrain:
can you format feature file to match other's style? thanks

kkaarreell:
I am not really sure which style you are referring to... :-) 
But I have changed the indentation a bit, hopefully it is better now.

ignatenkobrain:
you did exactly what I meant ;)

PR #239, opened 2017-01-26
add step "the command exit code is {values}"

kkaarreell:
Compares the exit code of the previous command with a given value or comma separated list of values or ranges, e.g. '1,3,100-102'.


dnf-docker-test/features/steps/command_steps.py
```diff
@@ -34,6 +34,23 @@ def step_the_command_should_pass(ctx):
...
+    Compares the exit code of the previous command with a given value or comma separated list of values or ranges, e.g. '1,3,100-102'.
+    """
+    codes = []
+    parts = values.split(',')
+    for part in parts:
+        if '-' in part:  # range a-b
+            (lower, upper) = part.split('-', 1)
+            lower = int(lower.strip())
+            upper = int(upper.strip())
+            codes.extend(range(lower, upper+1))
```
ignatenkobrain: I think it will not work when upper is lower than lower and it will just loop forever

kkaarreell: It shouldn't loop. With 'upper' lower than 'lower' the range() returns empty list, which is IMHO reasonable behaviour.

>>> range(1,2)
[1]
>>> range(1,1)
[]
>>> range(1,0)
[]
>>> range(1,-1)
[]

PR #245, opened 2017-02-02
Do not create arch specific subdirectories in repos

janblazek:
This patch changes the test generated repositories to have all rpms in single directory instead of arch specific subdirectories. This would a prerequisite for 'When I run "{command}" in directory of "{repository}" repository'.

### DIFF
--- a/dnf-docker-test/features/steps/repo_steps.py
+++ b/dnf-docker-test/features/steps/repo_steps.py
@@ -224,12 +224,13 @@ def given_repository_with_packages(ctx, rtype, repository):
         ctx.text = template.render(name=name, **settings)
         fname = "{!s}/{!s}.spec".format(tmpdir, name)
         step_a_file_filepath_with(ctx, fname)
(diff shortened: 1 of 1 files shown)


dnf-docker-test/features/steps/repo_steps.py
```diff
@@ -224,12 +224,13 @@ def given_repository_with_packages(ctx, rtype, repository):
         ctx.text = template.render(name=name, **settings)
         fname = "{!s}/{!s}.spec".format(tmpdir, name)
         step_a_file_filepath_with(ctx, fname)
+        buildname = '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}.rpm'
```
ignatenkobrain: actually, Epoch is missing. and there are plans to add couple of tests for epoch... could you fix it?

janblazek: Why should be epoch part of the rpm filename? At least Fedora and RHEL do not have epoch in the rpm filename.

ignatenkobrain: weird...

PR #253, opened 2017-02-13
add steps for GPG key generation and rpm signing

kkaarreell:
Adds support for GPG key creation and rpm signing via following steps:
  Given GPG key "James Bond"
     And GPG key "James Bond" imported in rpm database
     And repository "TestRepo" with packages signed by "James Bond"

Following packages needs to be added to the Docker image:
  gnupg2 - in order to be able to generate and export GPG key
  rng-tools - to ensure we have enough _random_ data for key generating
  rpm-sign - for the actual rpm signing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,12 @@ RUN set -x && \
     # jinja2:     rpmspec template
     # pexpect:    shell tests
     # rpm-build:  building dummy RPMs
-    dnf -y install httpd vsftpd python2-behave python2-six python-enum34 python2-whichcraft python-jinja2 python2-pexpect rpm-build openssl mod_ssl && \
+    # openssl:    generating TLS certificates
+    # mod_ssl:    https-style repos
+    # gnupg2:     GPG keys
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN set -x && \
     # jinja2:     rpmspec template
     # pexpect:    shell tests
     # rpm-build:  building dummy RPMs
-    dnf -y install httpd vsftpd python2-behave python2-six python-enum34 python2-whichcraft python-jinja2 python2-pexpect rpm-build openssl mod_ssl && \
+    dnf -y install httpd vsftpd python2-behave python2-six python-enum34 python2-whichcraft python-jinja2 python2-pexpect rpm-build openssl mod_ssl pinentry gnupg2 rng-tools rpm-sign && \
```
ignatenkobrain: can you add comment above what which tool stands for?

kkaarreell: Done, it turned out that pinentry is not actually necessary for generating unprotected keys

PR #254, opened 2017-02-13
test shell quit

MichaelMraka:
this should workaround failures of shell enable/disable test
which sometime fails with pexpect.EOF exception

### DIFF
--- a/dnf-docker-test/features/shell-5.feature
+++ b/dnf-docker-test/features/shell-5.feature
@@ -22,3 +22,6 @@ Scenario: Enabling and disabling a repository in dnf shell
        And I run dnf shell command "repolist"
       Then the command stdout should match regexp "TestRepoA"
        And the command stdout should not match regexp "TestRepoB"
+
+      When I run dnf shell command "quit"
+      Then the command stdout should match regexp "Leaving Shell"


ignatenkobrain:
That seems hacky and happening only on py2... Karel, have you encountered this issue? Any suggestions?

kkaarreell:
Yes, I have encountered this issue to, though it is quite rare. I think that neither expecting EOF nor extended the test scenario would solve the problem. According to the log 
  http://ci.dnf.baseurl.org/job/dnf-test/648/testReport/junit/shell-5-dnf-2/Enabling%20and%20disabling%20a%20repository%20in%20dnf%20shell/Enabling_and_disabling_a_repository_in_dnf_shell/
the problem appeared before the end of the test and last 3 steps were skipped. Therefore adding step for 'leaving shell' won't help.
Unfortunately, I don't know what is the root cause here.

kkaarreell:
OK, after some testing I think I have figured out the root cause which is probably some race condition happening due to "disablerepo TestRepo\*" followed by "enablerepo TestRepoA".
I don't know what dnf is doing but if I update the test case to do "enablerepo TestRepoB" instead it doesn't fail.
If anyone could confirm my findings I would propose to undo previous attempts to fix this problem and apply described _workaround_.

ignatenkobrain:
@kkaarreell thanks a lot, reverted.

PR #261, opened 2017-02-27
steps: support version comparison

ignatenkobrain:
Fixes: #185

### DIFF
--- a/dnf-docker-test/features/steps/repo_steps.py
+++ b/dnf-docker-test/features/steps/repo_steps.py
@@ -65,7 +65,7 @@ PKG_TMPL = """
 Name:           {{ name }}
 Summary:        {{ summary|default("Empty") }}
 Version:        {{ version|default("1") }}
-Release:        {{ release|default("1") }}%{?dist}
+Release:        {{ release|default("1") }}
 
 License:        {{ license|default("Public Domain") }}
 
--- a/dnf-docker-test/features/steps/rpm_utils.py
+++ b/dnf-docker-test/features/steps/rpm_utils.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import collections
 import enum
 import functools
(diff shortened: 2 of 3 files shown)


dnf-docker-test/features/steps/rpm_utils.py
```diff
@@ -47,17 +48,36 @@ def hdr2nevra(hdr):
...
+    else:
+        name = pkg
+        if ":" in evr:
+            epoch = evr.split(":", 1)[0]
+            evr = evr[len(epoch) + 1:]
+        if "-" in evr:
+            version, release = evr.split("-")
+        else:
+            version = evr
+
```
kkaarreell: The block doing the EVR parsing starting line 63 doesn't belong to this "else" branch bug the "if" branch above it.

kkaarreell: Hm, I can see that github reduced the code block but hopefully you would get the point.

dnf-docker-test/features/steps/repo_steps.py
```diff
@@ -65,7 +65,7 @@
 Name:           {{ name }}
 Summary:        {{ summary|default("Empty") }}
 Version:        {{ version|default("1") }}
-Release:        {{ release|default("1") }}%{?dist}
+Release:        {{ release|default("1") }}
```
kkaarreell: I am afraid that removing the dist tag could break some existing tests (I didn't try it yet though).
Do you think that preserving it and doing the actual comparison as follows is a bad idea?
`(release is None or p.release == release or p.release.split('.')[0] == release):`

ignatenkobrain: we don't need it at all, I included it there because I used to use it in Fedora. Without this distro-dependent data it will be only better.

PR #264, opened 2017-02-28
added list-1 test

glum23:
test for dnf list, all documented list suboptions and yum compatibility

### DIFF
--- a/dnf-docker-test/features/list-1.feature
+++ b/dnf-docker-test/features/list-1.feature
@@ -1,0 +1,199 @@
+Feature: Test for dnf list (including all documented suboptions and yum compatibility)
+ repo base: TestA-1 TestB-1
+ repo ext: XTest-2 TestA-2 TestB-2
+
+  @setup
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
same question here about partial output comparison. For example, when you will try to compare

```

+  Scenario: dnf list TestB TestA (when TestB is not installed)
+       When I successfully run "dnf list TestB TestA"
+       Then the command stdout should match regexp "TestA.*1-1"
+        And the command stdout should match regexp "Installed Packages"
+        And the command stdout should match regexp "TestB.*1-1"
+        And the command stdout should match regexp "Available Packages" 
```

it can be totally different from what you are thinking about...

glum23:
@ignatenkobrain Of course, imho it's an acceptable approximation. For an accurate check there should be better support in the testsuite steps. I really don't like that "match exactly", it's completely unusable for wildcards in pkgnames and it could be easily broken by some dnf additional messages (like "loaded plugins" in yum).

PR #267, opened 2017-03-01
Remove xfail from shell and swap tests

j-mracek:
We already fixed one issue with memory leak that affect shell and swap command,
therefore we can test if another issue appear. But to make it visible we have to
remove xfail mark.

### DIFF
--- a/dnf-docker-test/features/shell-1.feature
+++ b/dnf-docker-test/features/shell-1.feature
@@ -1,4 +1,3 @@
-@xfail
 Feature: Installing updating and removing a package in dnf shell
 
   Scenario: Installing packages
--- a/dnf-docker-test/features/shell-2.feature
+++ b/dnf-docker-test/features/shell-2.feature
@@ -1,4 +1,3 @@
-@xfail
 Feature: Running dnf shell commands in a batch
 
   @setup
--- a/dnf-docker-test/features/shell-3.feature
+++ b/dnf-docker-test/features/shell-3.feature
@@ -1,4 +1,3 @@
-@xfail
 Feature: Switching conflicting packages in dnf shell
 
   @setup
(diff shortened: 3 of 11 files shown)


j-mracek:
@ignatenkobrain Please look at it and merge it. Thanks

ignatenkobrain:
For how long is it merged?

http://ci.dnf.baseurl.org/job/dnf-test/test/?width=800&height=600&failureOnly=true

I've seen that shell stuff is still failing from time to time recently

j-mracek:
Ok, but now it looks like another problem. Probably proposed fix from @mmraka will help

j-mracek:
@ignatenkobrain Can you revert your revert of a proposal fix of @MichaelMraka that was supposed to help?

ignatenkobrain:
@j-mracek which one? point to commit please.

j-mracek:
@ignatenkobrain I guess this two, but you was more involve than me:
https://github.com/rpm-software-management/ci-dnf-stack/commit/a3083bdd086f13b7731deff92c5661b23a501e42
https://github.com/rpm-software-management/ci-dnf-stack/commit/6b5633a8450fd311773c1aca4ec56b7247014c99

And also think about closed PR: https://github.com/rpm-software-management/ci-dnf-stack/pull/254

kkaarreell:
Those patches are wrong and could be just hiding real bugs. 
Looking at the log
  http://ci.dnf.baseurl.org/job/dnf-test/732/testReport/junit/shell-4-dnf-2/Installing%20a%20package%20group%20in%20dnf%20shell/Updating_a_package_group_in_dnf_shell_using_the_upgrade_command/
it can be seen that the timeout is reached after the transaction is executed (run cmd) but the dnf shell prompt never appears again. This is simply wrong.
Unfortunately, I am not able to reproduce these failures locally.

PR #268, opened 2017-03-01
added tests for "dnf --security" and related options

glum23:
security_5 test currently fails due to --sec-severity (afaik it is implemented but probably not present in the latest dnf-nightly build yet)

### DIFF
--- a/dnf-docker-test/features/security_3.feature
+++ b/dnf-docker-test/features/security_3.feature
@@ -68,7 +68,6 @@ Feature: Test for check-update, upgrade, update, upgrade-minimal and update-mini
        Then rpmdb changes are
          | State      | Packages           |
          | updated    | TestA/4            |
-         | unchanged  | TestB              |
 
   Scenario: Cleanup after security update
        When I save rpmdb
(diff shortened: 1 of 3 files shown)


dnf-docker-test/features/security_3.feature
```diff
@@ -0,0 +1,119 @@
...
+         | updated    | TestA/4            |
+         | unchanged  | TestB              |
+
+  Scenario: Cleanup after security update
+       When I save rpmdb
+        And I successfully run "dnf -y history undo last" 
+       Then rpmdb changes are
+         | State      | Packages             |
+         | downgraded | TestA/1              |
+         | unchanged  | TestB                |
```
ignatenkobrain: note that you don't need to write this explicitly, it's default behavior for all non-mentioned packages in other sections

glum23: OK, removed unchanged states.
I've just wanted to stress the expected behaviour once the state is available.

ignatenkobrain:
LGTM, but just one nitpick. @glum23, do you really want to compare partial output instead of full?

PR #273, opened 2017-03-10
Introduced `remove-clean-requirements` test

edynox:
Unit test for testing this feature was removed
in swdb PR and is being replaced by functional one.

Test is using --setopt=clean_requirements_on_remove=True
or False to set if leaf package should be removed with
main package.


ignatenkobrain:
Merged with rewritten test :)

PR #243, opened 2017-02-01
introduce test for installroot without specifying releasever

mhatina:

PR #277, opened 2017-03-20
erase failing test from vars-releasever

mhatina:
Erased case is tested in https://github.com/rpm-software-management/ci-dnf-stack/pull/243


j-mracek:
Thanks

PR #279, opened 2017-03-22
implement steps for cmd output section matching

kkaarreell:


dnf-docker-test/features/steps/command_steps.py
```diff
@@ -79,3 +79,43 @@ def step_the_command_stream_should_match_regexp(ctx, stream, regexp):
...
+              TestB            noarch            1.0.0-1             test-1            5.7 k
+             \"\"\"
+            And the command stdout section "Installed:" should match regexp "TestA\.noarch.*TestB\.noarch"
+    """
+    ctx.assertion.assertIsNotNone(ctx.text, "Multiline text is not provided")
+    text = getattr(ctx.cmd_result, stream)
+    section_content = command_utils.extract_section_content_from_text(section, text)
+    ctx.assertion.assertRegexpMatches(section_content, ctx.text)
+
+@then('the command {stream:stdout_stderr} section "{section}" should match regexp "{regexp}"')
```
kkaarreell: Do you mean why the regexp matching? Why not? I am not saying it has to be used but why not to have it available? It would allow us to deal with some runtime dependent output, like installation time etc.

PR #285, opened 2017-04-18
Revert "introduce test for installroot without specifying releasever"

ignatenkobrain:
Reverts rpm-software-management/ci-dnf-stack#243

It is broken test


j-mracek:
@ignatenkobrain @mhatina Lets revert revert because of revert in dnf :-)

PR #286, opened 2017-04-20
steps to modify INI and repo files

kkaarreell:
adds steps to modify INI files and repo files in particular.

### DIFF
--- a/dnf-docker-test/features/steps/file_steps.py
+++ b/dnf-docker-test/features/steps/file_steps.py
@@ -128,7 +128,6 @@ def step_an_ini_file_filepath_modified_with(ctx, filepath):
             if not conf.has_section(section):
                 conf.add_section(section)
             keys = settings.keys()
-            keys.sort()
             for key in keys:
                 if key.startswith("-"):
                     key = key[1:]
--- a/dnf-docker-test/features/steps/file_steps.py
+++ b/dnf-docker-test/features/steps/file_steps.py
@@ -84,3 +84,55 @@ def step_an_ini_file_filepath_with(ctx, filepath):
             conf[section] = settings
 
     file_utils.create_file_with_contents(filepath, conf)
+
+@given('an INI file "{filepath}" modified with')
+def step_an_ini_file_filepath_modified_with(ctx, filepath):
+    """
+    Similar to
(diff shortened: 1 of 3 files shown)


dnf-docker-test/features/steps/file_steps.py
```diff
@@ -84,3 +84,55 @@ def step_an_ini_file_filepath_with(ctx, filepath):
...
+    updates = table_utils.parse_skv_table(ctx, HEADINGS_INI)
+    sections = updates.keys()
+    sections.sort()  # sort so we have removal first
+    conf = configparser.ConfigParser()
+    conf.read(filepath)
+    for section in sections:
+        settings = updates[section]
+        if section.startswith("-"):
+            section = section[1:]
+            ctx.assertion.assertTrue(conf.remove_section(section), "No such section '%s' in '%s'" % (section, filepath))
```
ignatenkobrain: shouldn't we silently skip it or print warning?

dnf-docker-test/features/steps/file_steps.py
```diff
@@ -84,3 +84,55 @@ def step_an_ini_file_filepath_with(ctx, filepath):
...
+        else:
+            if not conf.has_section(section):
+                conf.add_section(section)
+            keys = settings.keys()
+            keys.sort()
```
ignatenkobrain: why do we want to sort them?

ignatenkobrain:
overall looks good apart from overlaping code

dnf-docker-test/features/steps/file_steps.py
kkaarreell: I think it might be important to know that as it suggests that the test case is not correct.

dnf-docker-test/features/steps/file_steps.py
kkaarreell: Sorting would arrange removal (- prefix) first which let's user e.g. remove/clear whole section first and then add only required keys and values. I think it might be handy.

kkaarreell:
I wanted to avoid overlaps by reusing the code from file_steps.py but the problem is that the repo scenario has only table with 2 columns, not 3 and I didn't know how to simply convert it. Maybe you would know how to do that.

PR #288, opened 2017-04-24
tests for ssl related features

kkaarreell:


Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN set -x && \
     # gnupg2:     GPG keys
     # rng-tools:  to generate enough _random_ data for GPG keys
     # rpm-sign:   rpm signing
-    dnf -y install httpd vsftpd python2-behave python2-six python-enum34 python2-whichcraft python-jinja2 python2-pexpect rpm-build openssl mod_ssl gnupg2 rng-tools rpm-sign && \
+    dnf -y install httpd vsftpd python2-behave python2-six python-enum34 python2-whichcraft python-jinja2 python2-pexpect rpm-build openssl mod_ssl gnupg2 rng-tools rpm-sign psmisc && \
```
ignatenkobrain: why thing package is needed?

kkaarreell: It is needed for the killall command. Unfortunately, there is no pidof nor killall in the current image and therefore restart of httpd service is difficult to achieve. Maybe it could be done via an extra step that would read /var/run/httpd/httpd.pid in python and kill the service (not sure if kill is available though).

kkaarreell: I have uploaded an updated version that doesn't need the new package, the solution was obvious. :-)

PR #291, opened 2017-05-03
step for signing one GPG key with another

kkaarreell:
Necessary for testing gpgcakey functionality, unfortunately not yet implemented in DNF. FYI, into yum it was introduced via RHBZ#662712.

Following feature file would do the required setup

```
     Feature: Repodata signatures

         Scenario: Setup repository with signed metadata
           Given GPG key "JamesBond"
           Given GPG key "M"
           Given GPG key "JamesBond" signed by "M"
             And GPG key "JamesBond" imported in rpm database
             And enabled repository "TestRepo" with packages signed by "JamesBond"
               | Package | Tag | Value |
               | TestA   |     |       |
             And repository "TestRepo" metadata signed by "JamesBond"
             And a repo file of repository "TestRepo" modified with
               | Key           | Value |
               | repo_gpgcheck | True  |
	       | gpgcakey      | file:///root/M.pubkey |

```
When implemented properly, the installation of TestA should trigger the request to import the M key (CA key) and then the other key (JamesBond) should be imported automatically.

### DIFF
--- a/dnf-docker-test/features/steps/gpg_steps.py
+++ b/dnf-docker-test/features/steps/gpg_steps.py
@@ -26,6 +26,27 @@ GPGKEY_FILEPATH_TMPL = "/root/{!s}.{!s}"
 
 JINJA_ENV = jinja2.Environment(undefined=jinja2.StrictUndefined)
 
+@given('GPG key "{signed_key}" signed by "{signing_key}"')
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
would be nice to have that xfail'ed testcase for what you are talking about ;)

kkaarreell:
Yes, I am planning to do that.

PR #292, opened 2017-05-04
do not create subkey by default

kkaarreell:
Subkeys are causing issues when I want to reuse the code on RHEL-7. I propose not using them unless explicitly requested by the user.


ignatenkobrain:
what kind of problem does it cause on RHEL7? just wondering

kkaarreell:
I can't remember exactly but I think that for package signing a pubkey was used for some reason and verification with subkeys was not implemented in rpm from RHEL-7.
See also https://www.redhat.com/archives/rpm-list/2006-November/msg00105.html

ignatenkobrain:
@kkaarreell even with RPM 4.13? (since for DNF we require 4.13+, it is not a problem)

kkaarreell:
With RPM 4.13 it works. The question is what is the benefit of using subkeys.

PR #297, opened 2017-05-29
Add test install-downgrade-1

j-mracek:
The test cover behavior when install package with lover version that installed
(downgrade) but dependency is also required to be downgraded. Dnf should should
handle it without requirement if allowerasing.

### DIFF
--- a/dnf-docker-test/features/install-downgrade-1.feature
+++ b/dnf-docker-test/features/install-downgrade-1.feature
@@ -1,0 +1,27 @@
+Feature: Installing package with lover version
+
+  @setup
+  Scenario: Feature Setup
+      Given repository "available" with packages
+         | Package   | Tag       | Value       |
+         | TestA     | Version   | 1           |
+         |           | Requires  | TestB = 1-1 |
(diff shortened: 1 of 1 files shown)


dnf-docker-test/features/install-downgrade-1.feature
```diff
@@ -0,0 +1,27 @@
+Feature: Installing package with lover version
+
+  @setup
+  Scenario: Feature Setup
+      Given repository "available" with packages
+         | Package   | Tag       | Value       |
+         | TestA     | Version   | 1           |
+         |           | Requires  | TestB = 1-1 |
+         | TestA v2  | Version   | 2           |
```
ignatenkobrain: Please split this to second repository and then when you test upgrade, enable it

j-mracek: @ignatenkobrain Why I have to do it like that? This is valid user-case with single repository.

j-mracek:
@ignatenkobrain Please can you finish review?

j-mracek:
@ignatenkobrain Please can you finish the review.

PR #298, opened 2017-06-01
Change expected output in tree tests

j-mracek:
It reflects that in output of DNF is also present information about downloaded
metadata.

### DIFF
--- a/dnf-docker-test/features/repolist-enabled-disabled.feature
+++ b/dnf-docker-test/features/repolist-enabled-disabled.feature
@@ -14,40 +14,16 @@ Feature: Repolist with enabled/disabled repositories
 
   Scenario: Repolist without arguments
        When I successfully run "dnf repolist"
-       Then the command stdout should match exactly
-            """
-            repo id                               repo name                           status
-            TestB                                 TestB                               0
-            TestC                                 TestC                               3
-
-            """
+       Then the command stdout should match regexp "repo id +repo name +status\s+TestB +TestB +0\s+TestC +TestC +3"
 
   Scenario: Repolist with "enabled"
        When I successfully run "dnf repolist enabled"
-       Then the command stdout should match exactly
-            """
-            repo id                               repo name                           status
-            TestB                                 TestB                               0
-            TestC                                 TestC                               3
-
-            """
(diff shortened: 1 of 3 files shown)


dnf-docker-test/features/repolist-enabled-disabled.feature
```diff
@@ -14,40 +14,16 @@ Feature: Repolist with enabled/disabled repositories
...
        When I successfully run "dnf repolist all"
-       Then the command stdout should match exactly
-            """
-            repo id                             repo name                         status
-            TestA                               TestA                             disabled
-            TestB                               TestB                             enabled: 0
-            TestC                               TestC                             enabled: 3
-
-            """
+       Then the command stdout should match regexp "repo id +repo name +status\s+TestA +TestA +disabled\s+TestB +TestB +enabled: 0\s+TestC +TestC +enabled: 3"
```
ignatenkobrain: I would prefer to not have anyregexp in repolist at all, since 1) it is not visible what output should be 2)complicates everything

j-mracek: @ignatenkobrain Unfortunately we cannot use "Then the command stdout should match exactly" statement therefore there is not much options left how to handle it. From my point of view only one left.

ignatenkobrain: Can you implement `contains` step so it will match if substring specified by user is inside output? That way we still can handle exact output, but we don't get into trouble with repo download output and such.

j-mracek: @ignatenkobrain I am just using already implemented infrastructure. The presented patch is not sensitive to changes in number of white spaces, therefore this is exactly what I need.

jrohel: Cache is created during first test. So information about downloaded metadata are present only in output of first test.
Probably better is prepare cache during setup of test (Scenario Setup).  Then changes of "expected output in tree tests" will not needed. Tests will not be affected by informations about downloading.
PR: https://github.com/rpm-software-management/ci-dnf-stack/pull/301 changes of "expected output in tree tests".

jrohel:
PR was closed because isn't needed now. PR: #301 was used.

PR #300, opened 2017-06-30
Add test for main repoquery functionality

janblazek:
This is a test for dnf repoquery main functionality,
i.e. for options --requires, --provides, --conflicts, --obsoletes,
--whatrequires, --whatprovides, --whatconflicts, --whatobsoletes

### DIFF
--- a/dnf-docker-test/features/repoquery-2.feature
+++ b/dnf-docker-test/features/repoquery-2.feature
@@ -1,0 +1,60 @@
+Feature: Test for main repoquery functionality
+  for options --requires, --provides, --conflicts, --obsoletes,
+  --whatrequires, --whatprovides, --whatconflicts, --whatobsoletes
+
+  @setup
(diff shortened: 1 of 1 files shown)


kkaarreell:
Honzo,
would it be possible to make the check more specific, i.e. instead of 
`the command stdout should match regexp "TestB"`
use
`the command stdout should match exactly`?
The output of repoquery should not be complicated and therefore it seems to me better to match whole output.

ignatenkobrain:
Fully agree with @kkaarreell ;)

janblazek:
I have rewritten the test to use the "match exactly".
@kkaarreell , the "Last metadata expiration check" message is written to stderr and I am comparing just the stdout.

ignatenkobrain:
@janblazek could you rebase please?

janblazek:
@ignatenkobrain done.

PR #306, opened 2017-07-28
Added option to ignore packages when checking rpmdb changes

kkaarreell:
Adds new state 'ignored' which allows tested to list package
name(s) that should be ignored during the rpmdb changes check.
The package name could contain Unix shell-style wildcard to cover 
multiple packages.


dnf-docker-test/features/steps/rpm_steps.py
```diff
@@ -6,6 +6,7 @@
 from behave.model import Table
 import six
 from six.moves import zip
+import re
```
ignatenkobrain: you can use fnmatch for globs ;)

PR #312, opened 2017-08-24
support for @backup and @restore scenarios

kkaarreell:
Scenarios tagged with @backup and @restore can be used to restore specific changes made by the test to ease the execution of multiple tests outside of the Docker environment. @backup and @restore scenarios won't be executed by default when running in Docker.

### DIFF
--- a/dnf-docker-test/features/ssl-1.feature
+++ b/dnf-docker-test/features/ssl-1.feature
@@ -16,6 +16,10 @@ Feature: DNF SSL related features - Package installation
          | State     | Packages |
          | installed | TestA    |
 
+  @backup
+  Scenario: Backup /etc/httpd/conf.d/ssl.conf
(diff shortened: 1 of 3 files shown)


dnf-docker-test/features/ssl-1.feature
```diff
@@ -33,3 +37,8 @@ Feature: DNF SSL related features - Package installation
        When I run "dnf -v -y install TestC"
        Then the command should fail
         And the command stdout should match regexp "Cannot download repomd.xml"
+
+  @restore
+  Scenario: Restore httpd setting
+      Given I successfully run "mv /etc/httpd/conf.d/ssl.conf.testbackup /etc/httpd/conf.d/ssl.conf"
```
ignatenkobrain: probably `-f`?

kkaarreell: It is not necessary, the file would be overwritten by default.

PR #315, opened 2017-09-01
fix marking of unneeded packages

janblazek:
Fix marking of unneeded packages in Dockerfile when "dnf repoquery --unneeded" prints empty list.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,7 @@ RUN set -x && \
         # workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1398272
         rpm -q dnf && \
         # some unknown thing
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -40,7 +40,8 @@ RUN set -x && \
         # workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1398272
         rpm -q dnf && \
         # some unknown thing
-        dnf -y mark install $(dnf repoquery --unneeded -q); \
+        UNNEEDED=$(dnf repoquery --unneeded -q | xargs) && \
```
ignatenkobrain: `dnf -q repoquery --unneeded | xargs --no-run-if-empty dnf mark install`

janblazek: Thanks, this looks better.

PR #320, opened 2017-10-09
added download-1 test

glum23:
passed for 
dnf-2.7.3-1g5de9bd30.fc26.noarch
dnf-plugins-core-2.1.5-1.fc26.noarch

Scenarios for --source should be added when there's support for .src.rpm in test repos.


j-mracek:
The code looks good

j-mracek:
And it works

PR #327, opened 2017-10-12
pull dnf and libdnf updates from updates-testing repo too

kkaarreell:


ignatenkobrain:
Yeah, let's do this workaround for now.

PR #338, opened 2017-10-25
do not lauch @xfail scenarios

kkaarreell:
So that every test that should currently pass would pass.


ignatenkobrain:
It seems to conflict....

ignatenkobrain:
We really want to run xfail tests, but ignore the results... Whats the point of this?

ignatenkobrain:
As discussed on IRC, this is not needed.

PR #341, opened 2017-10-27
accept bash wildcards for test specification

kkaarreell:


ignatenkobrain:
you can use `shopt` if you need to modify behavior. But no pushd/echo/popd.

kkaarreell:
Fair enough. Would you mind providing a working code? I do not understand how shopt would help.

kkaarreell:
I did try different approach using find, avoiding pushd/popd. Despite being longer, it works a bit better.

PR #347, opened 2017-11-15
Added Dockerfile.boltron for running the testsuite on Boltron compose

kkaarreell:


Dockerfile.boltron
```diff
@@ -0,0 +1,73 @@
+FROM jamesantill/boltron-27
```
ignatenkobrain: we should really use original Fedora images here

kkaarreell: This is the image Irina was supposed to used for the testing. I have double checked that with her. We can update it to a different image when needed.

ignatenkobrain: Just saying that this image is not official in any way.

PR #348, opened 2017-11-24
added repo-priorities-1 test

glum23:
Created a test for repository priorities, cmds install, upgrade, downgrade.
The last scenario fails, pkg TestB is not downgraded to a version available in the highest prio repo but to a closest lower version. Is this OK?

### DIFF
--- a/dnf-docker-test/features/repo-priorities-1.feature
+++ b/dnf-docker-test/features/repo-priorities-1.feature
@@ -1,0 +1,116 @@
+Feature: Test for repository priorities
+
+  @setup
+  Scenario: Preparing the test repositories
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
I would also add distro-sync test here..

>  The last scenario fails, pkg TestB is not downgraded to a version available in the highest prio repo but to a closest lower version. Is this OK?

I would say this is a bug in dnf.

j-mracek:
I created a patch that should cange a behavior of dnf for downgrade command - https://github.com/rpm-software-management/dnf/pull/989

j-mracek:
Thanks a lot, tested and it works fine.

PR #353, opened 2018-01-19
Temporarily disable all check tests

m-blaha:
There is problem with rich dependencies in dnf check command,
so this tests now constantly produce confusing errors like:

cmake-rpm-macros-3.10.1-11.fc26.noarch has missing requires of (cmake = 3.10.1-11.fc26 if cmake < 3.10.1-11.fc26)
cmake-rpm-macros-3.10.1-11.fc26.noarch has missing requires of (cmake-data = 3.10.1-11.fc26 if cmake-data < 3.10.1-11.fc26)

### DIFF
--- a/dnf-docker-test/features/check-1.feature
+++ b/dnf-docker-test/features/check-1.feature
@@ -1,3 +1,4 @@
+@xfail
 Feature: Test for dnf check --duplicates command
 
   Scenario: Install package in version 1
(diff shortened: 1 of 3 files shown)


ignatenkobrain:
@m-blaha what's the point of sending PR and then merging it in next few minutes? why not to commit directly?

PR #356, opened 2018-02-08
overlay: freeze librepo

m-blaha:
freeze librepo distgit spec file to the latest working version


ignatenkobrain:
Commit message is actually inaccurate... Because distgit spec **is** working...

PR #364, opened 2018-03-16
Fix failing test with extended support of callback by dnf

j-mracek:
Match exactly is very sensitive if any new line appears.


m-blaha:
@rh-atomic-bot r+

PR #365, opened 2018-03-16
Not build SQLite DB

j-mracek:
This is a performance improvement. We do not run CI with yum therefore we don't
need this metadata format. The generation of repositories should be now about 5x
faster.


ignatenkobrain:
I don't think that 5x is real measurement... But sure.

PR #367, opened 2018-03-21
Remove distgit setting for librepo

j-mracek:
Due to presence of spec in upstream, it should work without it.


j-mracek:
Lets try if this will work

PR #368, opened 2018-03-26
Modify tests for software database

edynox:
Requires: https://github.com/rpm-software-management/libdnf/pull/433

Modify test cases to match SWDB.


j-mracek:
It works

PR #375, opened 2018-04-26
remove python3 incompatible an unnecessary decode() conversions

kkaarreell:
These changes are necessary to make the test suite work with python3. I have verified that it doesn't cause new test failures when run using python2.


kkaarreell:
@m-blaha , could you pls review this PR and merge? Thank you.

PR #380, opened 2018-05-04
Dockerfile for Fedora 28 image and Python 3

kkaarreell:


Dockerfile.f28.py3
```diff
@@ -0,0 +1,76 @@
...
+    # jinja2:       rpmspec template
+    # pexpect:      shell tests
+    # rpm-build:    building dummy RPMs
+    # openssl:      generating TLS certificates
+    # mod_ssl:      https-style repos
+    # gnupg2:       GPG keys
+    # rng-tools:    to generate enough _random_ data for GPG keys
+    # rpm-sign:     rpm signing
+    # createrepo_c: building repos
+    dnf -y install httpd vsftpd python3-behave python3-six python-enum34 python3-whichcraft python3-jinja2 python3-pexpect rpm-build openssl mod_ssl gnupg2 rng-tools rpm-sign createrepo_c && \
```
ignatenkobrain: `python-enum34` is py2

kkaarreell: Thank you. I have removed it completely as it should be provided by python3 directly.

PR #384, opened 2018-05-17
tests for module profile subcommand

janblazek:


kkaarreell:
I guess that for the last scenario it may be better if the ModuleB would have more profiles than just one.. however that would require more changes so let's not block the PR on that.

PR #386, opened 2018-05-23
added --noxfail and --tags options to dnf-testing.sh

kkaarreell:
Handy especially when developing tests and running them locally.


m-blaha:
These new options should be documented in --help command

kkaarreell:
I have updated the --help description.

PR #391, opened 2018-07-31
make releasever parsing RHEL compatible

kkaarreell:
RHEL differs in respective RPM provides from Fedora. The proposed change should work for both.


ignatenkobrain:
If you could include example how it looks on rhel, would be much better.

kkaarreell:
Hi Igor, I  have updated the description with samples.

PR #394, opened 2018-09-18
Fix module test according to new behavior of module solver

j-mracek:


kkaarreell:
Please do not remove comments with bug numbers. These would be necessary for bug verification to identify appropriate scenario. If the scenario is passing remove only @xfail and @bz123245 labels but preserve bug reference in the comment.

PR #396, opened 2018-09-20
Modularity info fixes

j-mracek:


j-mracek:
The module--info-1 test is broken therefore I am pushing it directly.

PR #400, opened 2018-10-23
add podman support

kkaarreell:


m-blaha:
One of the big advangages of podman is an ability to run as non-root. Would it be possible to get rid of all these sudo command in case of podman?

dnf-testing.sh
```diff
@@ -166,18 +179,18 @@ run()
...
                 if [ $? -ne 0 ]; then let ++failed && failed_test_name+=" $feature-$command"; fi
             done
         done
     else
         for feature in "${TESTS[@]}"; do
             feature=${feature%.feature}  # cut-off .feature suffix if present
-            for command in dnf-2 dnf-3; do
-                printf "\nsudo docker run $PARAM_TTY --rm -v "$devel" "$IMAGE" launch-test $PARAM_RESERVE $PARAM_TAGS "$feature" $command\n"
-                sudo docker run $PARAM_TTY --rm -v "$devel" -v "$devel_steps" "$IMAGE" launch-test $PARAM_RESERVE $PARAM_TAGS "$feature" $command >&2 || \
+            for command in dnf-3; do
```
m-blaha: I'm not opposing this, but shouldn't be there an option to re-enable testing also with dnf-2?

kkaarreell: Thanks for the catch, this is definitely not intended! Fixed now.

PR #405, opened 2018-11-07
remove Lazy from scenario name

kkaarreell:


kkaarreell:
confirmed to work on Fedora too

PR #407, opened 2018-11-07
module-install-pkg-4: added scenarios

glum23:
module-install-pkg-4: added further scenarios related to hybrid content support


kkaarreell:
I can confirm that the proposed tests are functional and passing.

PR #414, opened 2018-11-22
subtest for non-active stream content unavailability

kkaarreell:


kkaarreell:
confirmed working both on Fedora29 and RHEL-8 Beta

PR #416, opened 2018-11-26
added module-disable-4

glum23:
added new test module-disable-4.feature due to the bug 1613910


glum23:
passed both on rhel (dnf-4.0.4-1.el8) and fedora (dnf-4.0.9-0.32gcf6adacb.fc29)

PR #418, opened 2018-11-27
updated module-remove-1

glum23:
added reference to a relevant bug + removed workaround


glum23:
passed both for rhel and fedora29 (dnf-4.0.9)

PR #420, opened 2018-11-29
 added module-install-4

glum23:
added a test with scenario related to bz1566078


glum23:
passed both on rhel and fedora29

PR #421, opened 2018-11-29
added a test scenarios for module hotfixes

janblazek:


janblazek:
passed both on rhel and fedora29

PR #424, opened 2018-12-07
added module-install-5

glum23:
added a new test for installing modules with only non-modular packages (bug 1592408)


glum23:
passed both on rhel8 and fedora29

PR #437, opened 2018-12-20
added test for bug 1580455

kkaarreell:


kkaarreell:
> However, there's missing the bug reference (@bz1580455) 
fixed
> and for the scenario "Any stream for..." it would be nice when there are more streams available for the required module (moduleD). I've tried this and it seems that the first stream according to alphabetical order is chosen.
One such stream is intentional. The specification isn't specific about a stream and therefore we should not do any assumptions about the chosen stream if there is more than one available. Moreover, I believe the specific order is currently runtime dependent (i.e. something like an iteration through a set or dictionary keys)

kkaarreell:
Thank you.

PR #440, opened 2019-01-09
module-update-2: adjusted

glum23:
changed module-update-2 according to current modularity behaviour


glum23:
passed both for f29 and rhel8

PR #463, opened 2019-02-05
Best

Kratochvilova:
Hello, I made some tests for the new behavior with the "best" option, as requested by these bugs:
https://bugzilla.redhat.com/show_bug.cgi?id=1670776
https://bugzilla.redhat.com/show_bug.cgi?id=1671683
The tests fail without the following PRs:
https://github.com/rpm-software-management/dnf/pull/1311
https://github.com/rpm-software-management/libdnf/pull/678
I also needed to change few of the current tests to reflect the change.

I am not sure about the tags, though. I referenced both bugs in all the scenarios that require the changed behavior, but if I should do this differently just let me know.

In addition, the test in obsoletes-keep-reason.feature now found following bug: https://bugzilla.redhat.com/show_bug.cgi?id=1672618

### DIFF
--- a/dnf-docker-test/features/config-best.feature
+++ b/dnf-docker-test/features/config-best.feature
@@ -1,0 +1,63 @@
+Feature: Test for config option best and commandline options --best and --nobest
+
+  @setup
+  Scenario: Feature Setup
+      Given repository "test" with packages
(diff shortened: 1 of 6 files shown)


j-mracek:
Thanks, nice work.

PR #476, opened 2019-02-18
added test for bz1626093

kontura:
Testing whether plugin can successfully use DNF api to set http headers.
https://bugzilla.redhat.com/show_bug.cgi?id=1626093

### DIFF
--- a/dnf-docker-test/features/plugin-api-1.feature
+++ b/dnf-docker-test/features/plugin-api-1.feature
@@ -1,0 +1,35 @@
+Feature: DNF test (plugin api)
+
+@bz1626093
+Scenario: Plugin can edit http headers
+  Given http repository "base" with packages
+    | Package | Tag       | Value  |
+    | TestA   |           |        |
+  When I save rpmdb
+  And I enable repository "base"
+  Given a file "/test/plugins/test.py" with 
+  """
+  import dnf
+  import dnf.cli
+  import dnf.exceptions
+  import dnf.rpm.transaction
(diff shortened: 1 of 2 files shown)


dnf-docker-test/features/plugin-api-1.feature
```diff
@@ -0,0 +1,39 @@
...
+            repo.set_http_headers([ 
+                "custom_user_key20190218: custom_user_value", 
+            ])
+  """
+  Given a file "/etc/dnf/dnf.conf" with 
+  """
+  [main]
+  pluginpath=/test/plugins
+  """
+  And I successfully run "dnf clean all"
```
kkaarreell: When seems to be more appropriate than Given (And)

dnf-docker-test/features/plugin-api-1.feature
```diff
@@ -0,0 +1,39 @@
...
+  Then I sniff packets expecting "custom_user_key20190218: custom_user_value"
+  Given a file "/etc/dnf/dnf.conf" with
```
kkaarreell: What is the purpose of this Given? Is this a clean up step (which is not necessary)? Or is the test somehow incomplete?

kontura: It was supposed to be clean up, I originally had it in different feature with other tests. You are right, it is not necessary.

kontura:
Thank you for feedback. I updated the commit.

PR #477, opened 2019-02-19
Add new scenario for module disable command

j-mracek:


kkaarreell:
the test is failing for me on F29 with dnf-4.2.0-0.12g2a097cbd.fc29.noarch. If the fix it not in F29 it should be probably marked as @xfail.

j-mracek:
It requires patch from https://github.com/rpm-software-management/dnf/pull/1326 that was not merged yet.

m-blaha:
Required patch for dnf was merged into upstream.

PR #479, opened 2019-02-19
added test for bz1633151

kontura:


kkaarreell:
I confirm this is passing on RHEL-8.0. Can't test on F29 though. Thank you

kkaarreell:
tested also on F29. Could you please also create rhel-8.0.0 PR?

PR #480, opened 2019-02-19
Add test for checking number of updates, bz1585138

pkratoch:


dnf-docker-test/features/upgrade-5.feature
```diff
@@ -0,0 +1,62 @@
+@bz1585138
+Feature: Print correct number of available updates if update type is given
+
+  @setup
+  Scenario: setup
+    Given repository "base" with packages
+         | Package  | Tag      | Value  |
+         | TestA    |          |        |
+         | TestB    |          |        |
+         | TestC    |          |        |
```
m-blaha: I think you need to add a package without update available to reproduce the bug 1585138. The root cause of the bug was, that dnf reported number of all packages instead of number of packages with updates available. Now you have both these numbers the same.

pkratoch: You are right, I now added packages that don't have any updates, and tested it fails on a system with ``dnf-2.7.5`` (and still passes for the upstream). Thank you!

PR #491, opened 2019-02-21
Minor Dockerfile polishment

dmnks:
Makes it easier to use (I hope).
Please note that `dnf-testing.sh` would have to be updated as well, however it currently doesn't seem to work anyway (it executes docker with "launch-test" which doesn't seem to exist).


lukash:
Sadly this is way too outdated now, I'll close it, if you've maintained your improvements, @dmnks, perhaps open another PR?

dmnks:
Oh, nevermind. This was just a cosmetic thing but you're right, the codebase has changed a lot since then so let's forget about it.

PR #499, opened 2019-02-22
Add test for "Better counting" feature

dmnks:
Covers the F30 change [DNF Better Counting](https://fedoraproject.org/wiki/Changes/DNF_Better_Counting).

Associated PRs: rpm-software-management/libdnf#684 and rpm-software-management/dnf#1324


kkaarreell:
Seems there is also unrelated Dockerfile update.
Btw, I would propose to use generic @step instead of @given, @when, @then etc. so we won't limit the possible usage unnecessarily.

dmnks:
> Seems there is also unrelated Dockerfile update.

Correct; it's a convenience thing for myself. I have actually created a separate pull request for that change, on which I based this branch.  However, please keep in mind this is still a WIP so I'll rebase it however necessary once it's ready for review.

> Btw, I would propose to use generic @step instead of @given, @when, @then etc. so we won't limit the possible usage unnecessarily.

Good point, thanks, will incorporate.

PR #504, opened 2019-02-25
Added test for RhBug:1680684

m-blaha:


dnf-docker-test/features/module-install-7.feature
```diff
@@ -0,0 +1,24 @@
+Feature: Profile installation errors
+
+# package FileConflict-1.0-1.x86_64 has file conflicts with
+# FileConflict-0:2.0.streamB-1.x86_64 from module test-module
+
+@xfail @bz1656782
```
kkaarreell: Is the bug reference incorrect?
Anyway, if the bug won't be fixed in 8.0.0 we don't need a PR for that branch.

m-blaha: Oh, sorry. The test is for similar bug https://bugzilla.redhat.com/show_bug.cgi?id=1680684. I'll fix it in master branch PR and close it here.

PR #505, opened 2019-02-25
Add test for downloading package from multiple repositories, bz1612874

pkratoch:
The test passes with the current version of ``dnf-plugins-core``, and fails with ``dnf-plugins-core-3.0.1-2.fc29.noarch`` . However, it is not 100 % reliable. Since the bug depends on the order of downloads, there might be a run when the test passes even though it shouldn't.

Interestingly, it sometimes fails with "Status code: 416" (just like in the original bug) and sometimes with "checksum doesn't match".

### DIFF
--- a/dnf-docker-test/features/download-2.feature
+++ b/dnf-docker-test/features/download-2.feature
@@ -1,0 +1,35 @@
+Feature: Test for download when there are multiple packages of the same NEVRA
+
+  @setup
+  Scenario: Feature Setup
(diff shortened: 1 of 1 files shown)


kkaarreell:
we will need PR for rhel-8.0.0. branch

kkaarreell:
PR for rhel-8.0.0 https://github.com/rpm-software-management/ci-dnf-stack/pull/509

PR #528, opened 2019-04-17
increase sleep as the test is fragile with the current value

kkaarreell:
I am encountering occasional test failures in CI and the change resolves it.


lukash:
As much as I hate sleeps, I guess if expiration time is 1s, we need 2s sleep to ensure it works.

PR #534, opened 2019-04-26
Tests for builddep command in modular environment

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1677583

### DIFF
--- a/dnf-behave-tests/features/plugins-core/builddep-modularity.feature
+++ b/dnf-behave-tests/features/plugins-core/builddep-modularity.feature
@@ -1,0 +1,92 @@
+Feature: Tests for builddep command on modular system
+
+Background: Enable builddep plugin
+  Given I enable plugin "builddep"
+    And I use the repository "dnf-ci-fedora"
+
+# dnf-ci-fedora-modular repo:
+#   module ninja:master [d] contains ninja-build-0:1.8.2-4.module_1991+4e5efe2f.x86_64
+#   ninja:legacy contains ninja-build-0:1.5.2-1.module_1991+4e5efe2f.x86_64
+#   ninja:development contans ninja-build-1.9.2-1.module_1991+4e5efe2f.x86_64
+# dnf-ci-fedora repo (non-modular):
+#   ninja-build-0:1.8.2-5.fc29.x86_64
+
+Scenario: Builddep installs non-modular build requirements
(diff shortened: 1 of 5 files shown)


dnf-behave-tests/features/plugins-core/builddep-modularity.feature
```diff
@@ -0,0 +1,90 @@
...
+   When I execute dnf with args "builddep {context.dnf.fixturesdir}/repos/dnf-ci-builddep/src/build-requires-ninja-build-1.0-1.src.rpm"
+   Then the exit code is 0
+    And Transaction contains
+        | Action                | Package                                           |
+        | install               | ninja-build-0:1.5.2-1.module_1991+4e5efe2f.x86_64 |
+
+
+# build-requires-ninja-build-0.5-1.src.rpm build requires ninja-build < 1.8
+@1677583
+Scenario: Builddep reports error where required package is available in non-enabled non-default stream
```
pkratoch: I am missing there a word "only" here - only in non-enabled non-default stream. Because if it's also in some enabled stream it doesn't report the error.
Otherwise, LGTM.

dnf-behave-tests/features/plugins-core/builddep-modularity.feature
```diff
@@ -0,0 +1,90 @@
+Feature: Tests for builddep command on modular system
+
+Background: Enable builddep plugin
+  Given I enable plugin "builddep"
+    And I use the repository "dnf-ci-fedora"
+
+# module ninja:master [d] contains ninja-build-0:1.8.2-4.module_1991+4e5efe2f.x86_64
+# ninja:legacy contains ninja-build-0:1.5.2-1.module_1991+4e5efe2f.x86_64
+# ninja:development contans ninja-build-1.9.2-1.module_1991+4e5efe2f.x86_64
```
pkratoch: Maybe also add that these three module streams are from dnf-ci-fedora-modular repo, so that it's immediately clear. But it's not that important.

dnf-behave-tests/features/plugins-core/builddep-modularity.feature
m-blaha: Fixed.

dnf-behave-tests/features/plugins-core/builddep-modularity.feature
m-blaha: Thanks for feedback, fixed.

PR #535, opened 2019-04-29
Add tests for download --source (RhBug:1649627)

pkratoch:
This is a WIP for now, since the fix for bug 1649627 is not developed yet (either that or the test is wrong).

All lower-version package specifications in the examples are failing - dnf returns exit code 1 and the package flac-1.3.3-2.fc29.src.rpm never gets downloaded.

The scenario "Check using repoquery that the source RPM exists" shows, however, that the package exists (this scenario can be removed after the fix is developed).

### DIFF
--- a/dnf-behave-tests/features/plugins-core/download-source.feature
+++ b/dnf-behave-tests/features/plugins-core/download-source.feature
@@ -73,4 +73,27 @@ Examples:
   | wget-1.19.5-5.fc29.x86_64   | wget-1.19.5-5.fc29.src.rpm     |
 
 
+@xfail
+Scenario Outline: Download a source RPM when there are more epochs available
+  Given I use the http repository based on "dnf-ci-fedora-updates-testing"
+   When I execute dnf with args "download --destdir={context.dnf.tempdir} --source <pkgspec>"
+   Then the exit code is 0
+    And stdout contains "<srpm>"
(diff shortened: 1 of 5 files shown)


pkratoch:
The tests are passing with the new PR: https://github.com/rpm-software-management/dnf-plugins-core/pull/331

I removed the  repoquery test.

pkratoch:
I extended the tests with package specs that contain epoch numbers.

The tests now fail for ``wget-0:1.19.5-5.fc29*`` package specs. All these packages should be downloaded from ``dnf-ci-fedora`` repo, but are actually downloaded from ``dnf-ci-fedora-updates-testing`` repo (where is a package with the same NVR but higher epoch).

j-mracek:
> I extended the tests with package specs that contain epoch numbers.
> 
> The tests now fail for `wget-0:1.19.5-5.fc29*` package specs. All these packages should be downloaded from `dnf-ci-fedora` repo, but are actually downloaded from `dnf-ci-fedora-updates-testing` repo (where is a package with the same NVR but higher epoch).

It sounds like a problem in another part of the code.

pkratoch:
I split the tests into two. The epoch is a separate issue and doesn't need to be addressed as a part of bug 1649627.

@j-mracek, thanks for review and for showing me how the package specs with epochs are queried.

PR #537, opened 2019-05-03
Add steps and first batch of tests for createrepo_c

kontura:


kontura:
I have converted most of the [acceptance tests from createrepo_c](https://github.com/rpm-software-management/createrepo_c/tree/master/acceptance_tests/tests). Because the original acceptance tests were comparing output of createrepo and createrepo_c I left out a few tests that didn't make much sense without running the original createrepo.

On top of that this PR contains tests for --keep-all-metadata, repo with subdirs and mergerepo_c with modular metadata.

kontura:
Removed most of the specific checksums as they make the tests fragile, small change in sqlite version or xml format can change checksum of the file and break the tests.

kontura:
I have added tags fedora_ge_30 to tests covering newer functionality that is only available in fedora 30.

lukash:
Can we *please* separate the `createrepo_c` tests if at all possible? I haven't checked how much of the existing code you're reusing, but still, we should look for basically any other solution first.

You're adding a huge amount of code all on the same pile and it is going to become a mess. We should actually try to separate the tests as much as possible, not bunch them together.

dnf-behave-tests/features/steps/common/file.py
```diff
@@ -6,6 +6,41 @@
...
+    from backports import lzma
+
+def regex_find_file(filepath):
+    files = os.listdir(os.path.dirname(filepath))
+    filepath = filepath + '$'
+    pattern = re.compile(os.path.basename(filepath))
+    matched_files = list(filter(pattern.match, files))
+    match_count = len(matched_files)
+    if (match_count > 1):
+        raise AssertionError("Multiple files matching: {0}, found:\n{1}" .format(filepath, '\n'.join(matched_files)))
```
m-blaha: You are printing filepath with '$' in the end.
You probably should not change filepath and do something like
pattern = re.compile(os.path.basename(filepath) + '$')

dnf-behave-tests/features/steps/common/file.py
```diff
@@ -16,6 +51,10 @@ def ensure_directory_exists(dirname):
 def ensure_file_exists(filename):
     assert os.path.exists(filename), "ENSURE: file exists {!r}".format(filename)
 
+def regex_ensure_file_exists(filename):
```
m-blaha: I'd rather do
```
f = regex_find_file(filename)
ensure_file_exists(f)
```
instead of creating this new function. It is only called from one place.

dnf-behave-tests/features/steps/common/file.py
```diff
@@ -35,6 +74,17 @@ def create_file_with_contents(filename, contents, encoding="utf-8"):
         outstream.flush()
     assert os.path.exists(filename), "ENSURE: file exists {!r}".format(filename)
 
+def regex_read_file_contents(filename, encoding="utf-8"):
```
m-blaha: Rather reuse current read_file_contents. You can extend it to be able to cope with compressed files, but I do not see strong reason for creating this new function.

dnf-behave-tests/features/steps/common/file.py
```diff
@@ -44,3 +94,51 @@ def read_file_contents(filename, encoding="utf-8"):
 def copy_tree(source, destination):
     shutil.copytree(source, destination)
     assert os.path.exists(destination), "copy_tree {} -> {} failed".format(source, destination)
+
+def copy_file(source, destination):
+    shutil.copyfile(source, destination)
```
m-blaha: shutil.copyfile would probably raise Error in case of troubles with copying (before your assert checks it on the next line). May be just wrap it in try: block.

dnf-behave-tests/features/steps/common/repodata_representation.py
```diff
@@ -0,0 +1,167 @@
+# -*- coding: utf-8 -*-
+
+class MetadataItem:
```
m-blaha: Please, use new style classes for python2: `class MetadataItem(object)`
Just in case we needed to run the tests on python2.

dnf-behave-tests/features/steps/common/repodata_representation.py
```diff
@@ -0,0 +1,167 @@
...
+    DIFF_ATTR = tuple()
+    NONE_AND_EMPTY_ARE_SAME = tuple()
+
+    def diff(self, other):
+        comparable_attributes = set(self.DIFF_ATTR).intersection(set(other.DIFF_ATTR))
+        for key in comparable_attributes:
+            a = getattr(self, key)
+            b = getattr(other, key)
+
+            if key in self.NONE_AND_EMPTY_ARE_SAME:
```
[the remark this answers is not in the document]
kontura: They should not be missing, getattr will raise if the key is not an attribute. 

It's to compare values from sqlite which will set None if it doesn't have the value while xml works with strings so it ends up as empty string.

dnf-behave-tests/features/steps/cmd.py
```diff
@@ -41,6 +42,33 @@ def when_I_execute_dnf_with_args(context, args):
...
+
+
+@behave.step("I execute modifyrepo_c with args \"{args}\" in directory \"{cwd}\"")
+def when_i_execute_modifyrepo_c_with_args(context, args, cwd):
+    when_i_execute_binary_with_args(context, "modifyrepo_c", args, cwd)
+
+@behave.step("I execute sqliterepo_c with args \"{args}\" in directory \"{cwd}\"")
+def when_i_execute_sqliterepo_c_with_args(context, args, cwd):
+    when_i_execute_binary_with_args(context, "sqliterepo_c", args, cwd)
+
```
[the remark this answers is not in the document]
kontura: I started to use the generic step at first but I didn't like the quotes around the name of the binary in each step. Therefore I created these thin wrappers since there is only four of them I though its fine and it simplified reading of the steps a bit for me (one less arguments to worry about).

kontura:
Thank you for the very helpful feedback @m-blaha! I fixed all the problems that had clear cut solutions. 

@lukash it looks like its gonna end up together, from what I understand there are also infrastructure reasons for this.

lukash:
@kontura well, I'd still like to know the reasons...

dnf-behave-tests/features/steps/common/sqlite_repodata.py
```diff
@@ -0,0 +1,134 @@
...
+        #      it with only one empty string
+        splited_filenames = filenames.split('/')
+        filtered_splited_filenames = []
+        if len(filenames):
+            filtered_splited_filenames.append(splited_filenames[0])
+        for i in range(1, len(filenames.split('/'))):
+            if splited_filenames[i] == '' and splited_filenames[i - 1] == '':
+                continue
+            filtered_splited_filenames.append(splited_filenames[i])
+        # XXX: End
```
m-blaha: This part looks strange. I do not understand what `If there is dir without name '/' it's splited into 2 empty string` means. At first sight stripping trailing '/' before filenames.split('/') should be enough?

kontura: I think that would not work, sqlite can have a row with filenames column like this: `//testa/testb` with dirname just `/` this is caused by rpm which has just "/" in its %files section. 
We want the leading single "/" to result in one empty string that is than used with the dirname `/`. To get the xml equivalent `<file type="dir">/</file>`

I think this is a special case for the '/' because there is no better way how to express it with the dirname, filenames schema.

kontura:
I have fixed up all the feedback + I added separate test data for almost all the tests except for mergerepo_c merging modules.yamls because that would require two additional new modular repositories and I don't think that is justified.

dnf-behave-tests/features/steps/common/sqlite_repodata.py
kontura: I think I managed to simplify the code in the newest version.

kontura:
I fixed a mistake I made while simplifying code from https://github.com/rpm-software-management/ci-dnf-stack/pull/537#discussion_r298978367.

dnf-behave-tests/fixtures/specs/createrepo_c-ci-notinstallable/createrepo_c-0.14.4.spec
```diff
@@ -0,0 +1,145 @@
+%global libmodulemd_version 2.3.0
+
+Summary:        Creates a common metadata repository
+Name:           createrepo_c
```
lukash: I find it confusing you are using the `createrepo_c` package as test data for `createrepo_c` tests. It appears like this package here somehow could have a specific role in the tests, until you dig a bit and find out it can just be arbitrary test data. Also it will be unnecessarily popping up while grepping for e.g. "createrepo_c".

And, it seems this is the full spec file, which contains a lot of (I assume) redundant things, like listing all the dependencies and installs a lot of files, of which you probably only need a few?

Seems you don't need too many packages for the tests, which is good, why not further reduce the footprint and only leave what's actually important in the test data?

dnf-behave-tests/features-createrepo_c/createrepo_c_empty_repository.feature
```diff
@@ -0,0 +1,396 @@
...
+     | other_db     | ${checksum}-other.sqlite.bz2                                                          | sha256        | bz2              |
+  And repodata "/temp-repos/temp-repo/repodata/" are consistent
+  And file "/temp-repos/temp-repo/repodata/repomd.xml" contains lines
+  """
+  <revision>XYZ</revision>
+  """
+
+
+Scenario: Repo from empty directory with --skip-symlinks and symlinked package
+ Then I create symlink from "{context.dnf.repos_location}/dnf-ci-fedora/noarch/setup-2.12.1-1.fc29.noarch.rpm" to "/temp-repos/temp-repo/setup-2.12.1-1.fc29.noarch.rpm"
```
lukash: Applies to all the feature files, I've just picked this one instance.

As I've said before it's kind of a misfeature of behave to allow basically any combination of `{Given,When,Then} something` (although we could alleviate that by defining the steps better perhaps), while at the same time the semantics often have to depend on the author, for:

`Given` - test setup, you shouldn't have to test it was successful
`When` - run the command that is being tested
`Then` - validate the result of the test

`And` - supposedly used as a continuation of one of the three above steps, doesn't really make things easier especially when you move lines around :slightly_frowning_face: 

At least that's how I understand it... Anyway, if you wouldn't mind fixing it throughout :slightly_smiling_face:

dnf-behave-tests/features-createrepo_c/createrepo_c_cache.feature
```diff
@@ -0,0 +1,44 @@
...
+  And repodata "/temp-repo/repodata/" are consistent
+  And repodata in "/temp-repo/repodata/" is
+     | Type                | File                                | Checksum Type | Compression Type |
+     | primary             | ${checksum}-primary.xml.gz          | sha256        | gz               |
+     | filelists           | ${checksum}-filelists.xml.gz        | sha256        | gz               |
+     | other               | ${checksum}-other.xml.gz            | sha256        | gz               |
+     | primary_db          | ${checksum}-primary.sqlite.bz2      | sha256        | bz2              |
+     | filelists_db        | ${checksum}-filelists.sqlite.bz2    | sha256        | bz2              |
+     | other_db            | ${checksum}-other.sqlite.bz2        | sha256        | bz2              |
+ And primary in "/temp-repo/repodata" contains only packages
```
lukash: Indentation is another thing, I'm really not a fan of this style of alignment exactly because this iadvertedly will start happening - this `And` is one space off to the left, as well as the tables above and below it. This also repeats throughout the whole commit.

Please align the tables as well as string literals to the six space indent and fix the `And`s that are off... It probably wouldn't be hard to come up with a sed command to fix the majority of it.

kontura:
Alright I have updated this to use the newly separated step structure and I have done couple of smaller tweaks as well (such as addressing @lukash's last review).

There are also two completely new commits:
- one ensures that dnf and createrepo_c streps both access `repos_location` the same way 
- second adds createrepo_c tests to dnf-testing.sh so that they are used.

Lastly there is a few Sider issues left unresolved, I think 5 of them are due to the fact we are modifying `sys.path` before imports in order to import common steps and the last one is explained [here](https://lintlyci.github.io/Flake8Rules/rules/W503.html).

dnf-behave-tests/features-createrepo_c/createrepo_c_regular_repository.feature
```diff
@@ -0,0 +1,165 @@
...
+      | filelists_db        | ${checksum}-filelists.sqlite.bz2    | sha256        | bz2              |
+      | other_db            | ${checksum}-other.sqlite.bz2        | sha256        | bz2              |
+  And primary in "/temp-repo/repodata" has only packages
+      | Name          | Epoch | Version | Release | Architecture |
+      | package       | 0     | 0.2.1   | 1.fc29  | x86_64       |
+      | package-devel | 0     | 0.2.1   | 1.fc29  | x86_64       |
+      | package-libs  | 0     | 0.2.1   | 1.fc29  | x86_64       |
+
+
+Scenario: create regular consistent repository wihile excluding all packages
```
m-blaha: typo `wihile`

dnf-behave-tests/features-createrepo_c/createrepo_c_regular_repository.feature
```diff
@@ -0,0 +1,165 @@
...
+      | primary_db          | ${checksum}-primary.sqlite.bz2      | sha256        | bz2              |
+      | filelists_db        | ${checksum}-filelists.sqlite.bz2    | sha256        | bz2              |
+      | other_db            | ${checksum}-other.sqlite.bz2        | sha256        | bz2              |
+  And primary in "/temp-repo/repodata" has only packages
+      | Name          | Epoch | Version | Release | Architecture |
+      | package       | 0     | 0.2.1   | 1.fc29  | x86_64       |
+      | package-libs  | 0     | 0.2.1   | 1.fc29  | x86_64       |
+
+
+Scenario: create regular consistent repository wihile excluding packages by wildcards
```
m-blaha: again `wihile`

kontura:
I think I fixed all the review points. I have split the changes to already present steps into separate commits, it could also be done for the new createrepo_c specific steps if required. 

I ended up not unifying the already present function for sha256 hashing with the new hash functions (I put them in `createrepo_c/steps/lib/file.py`) the new one works on files and its not entirely straightforward to modify it so that it is able to work on strings. But I might spent some more time on in possibly in a different commit?

dnf-behave-tests/features/microdnf/install-xml-base.feature
```diff
@@ -0,0 +1,100 @@
+@no_installroot
+@destructive
```
m-blaha: The `@no_installroot` decorator itself is listed among destructive tags, you do not add `@destructive`

dnf-behave-tests/common/file.py
```diff
@@ -47,9 +47,8 @@ def step_impl(context, filepath):
 @behave.given('I create symlink "{dst}" to file "{src}"')
 def step_impl(context, dst, src):
     dst = prepend_installroot(context, dst)
-    src = prepend_installroot(context, src)
     ensure_directory_exists(os.path.dirname(dst))
-    os.symlink(src, dst)
+    os.symlink(src.format(context=context), dst)
```
lukash: I think we should be consistent in how we handle the paths in the file manipulation steps. I'm already pretty confused about what steps do substitution and where and what steps prepend the installroot. Not necessarily in this PR, but for this one function... is there a problem with both prepending the installroot and doing the substitution? I'd just do both on all the paths in all the steps :slightly_smiling_face: (so, e.g. here I'd do it on destination too - perhaps we'd just have to check if it's a relative path and don't prepend installroot for that case)

dnf-behave-tests/common/lib/file.py
```diff
@@ -58,3 +72,28 @@ def copy_file(source, destination):
...
+    else:
+        basename = os.path.basename(src)
+        dst = os.path.join(dst_folder, basename)
+        if src.endswith(".bz2"):
+            destination = dst[:-4]
+        else:
+            destination = dst[:-3]
+
+        open(destination, "wb").write(content)
+        return destination
```
lukash: This function returns a mix of very different values depending on the situation (`False`, path to a file, file contents).

I think this one should return the decompressed file content if it was an archive or `None` if it is not. Then you can have another function that will call this one and store the content in the destination. That one could probably return the decompressed file path. Also, this second function is really just used in the new tests in `createrepo_c/steps/repodata.py`, so you can put it there.

dnf-behave-tests/features/microdnf/install-xml-base.feature
```diff
@@ -0,0 +1,100 @@
...
+  #  And I execute "mergerepo_c --repo file://{context.dnf.repos[microdnf-ci-xml-base].path} --repo file://{context.dnf.repos[dnf-ci-thirdparty].path}" in "{context.dnf.installroot}"
+  #  And I configure a new repository "merged-repo" with
+  #      | key     | value                                        |
+  #      | baseurl | file://{context.dnf.installroot}/merged_repo |
+  # When I execute microdnf with args "install xml_base_package alternator"
+  # Then the exit code is 0
+  # # We don't want to cache local packages
+  #  And file "/var/cache/dnf/merged-repo*/packages/alternator*" does not exist
+  # # We want to cache remote packages
+  #  And file "/var/cache/dnf/merged-repo*/packages/xml_base_package*" exists
```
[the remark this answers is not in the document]
kontura: I am sorry I included the install-xml-base.feature for microdnf by mistake. Removed.

dnf-behave-tests/common/lib/file.py
kontura: I like your suggestion much better than my version, will do, thanks.

dnf-behave-tests/common/file.py
kontura: I didn't realize that the prepend_installroot function already does this, it can prepend installroot and do the substitution as well. So I can undo my change and just prepend all the paths with `/`. 

Since https://github.com/rpm-software-management/ci-dnf-stack/commit/0bfed1f6e418800b8e19e7c9d73fc350b1e584db I can just do `/{context.scenario.repos_location/path/to/createrepo_c/repo` and the prepend installroot will do the substitution and since it will have two `//` it will not prepend the installroot.

Is this what you mean by doing both?

lukash: Yeah, that's what I meant. And yeah, it does indeed do that already, so you can use that :slightly_smiling_face:

kontura: To answer your question I don't think it makes much sense to try to do both at the same time. Prepending installroot while doing substitution on `{context.scenario.repos_location}` does not seem helpful because repos_location is not inside installroot.

dnf-behave-tests/common/lib/file.py
```diff
@@ -44,6 +53,11 @@ def create_file_with_contents(filename, contents, encoding="utf-8"):
 
 def read_file_contents(filename, encoding="utf-8"):
     assert os.path.exists(filename), "ENSURE: file exists {!r}".format(filename)
+
+    output = decompress_file_by_extension(filename)
+    if output:
```
lukash: This should now be `if output is not None:`.

dnf-behave-tests/createrepo_c/steps/lib/file.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def file_as_blockiter(afile, blocksize=65536):
+    with afile:
+        block = afile.read(blocksize)
+        while len(block) > 0:
+            yield block
+            block = afile.read(blocksize)
+
+def decompress_file_by_extension_to_dir(compressed_filepath, dest_dir):
```
lukash: Two blank lines between functions :slightly_smiling_face:

dnf-behave-tests/createrepo_c/steps/lib/file.py
```diff
@@ -0,0 +1,91 @@
...
+    content = decompress_file_by_extension(compressed_filepath)
+
+    if not content:
```
lukash: Also should be `if content is None:`

dnf-behave-tests/createrepo_c/steps/lib/repodata.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+
+
```
lukash: An extra blank line here...

kontura: They should all be resolved. :+1:

lukash:
Good job :slightly_smiling_face: ready to merge from my side.

m-blaha:
Thanks Ales!

PR #539, opened 2019-05-07
Add dummy yum pkg and test if it is protected on RHEL>=8 (RhBug:1639363)

kontura:
Requires dnf-4.2.1 on RHEL8

### DIFF
--- a/dnf-behave-tests/features/remove-protected.feature
+++ b/dnf-behave-tests/features/remove-protected.feature
@@ -1,0 +1,10 @@
+Feature: Do not remove protected RPMs
+
+@use.with_os=rhel__ge__8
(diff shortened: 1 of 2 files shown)


kontura:
Added @xfail flag

dnf-behave-tests/features/remove-protected.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Do not remove protected RPMs
+
+# should fail until dnf-4.2.1 is on RHEL8
+@xfail
```
m-blaha: The scenario is already passing on RHEL, please remove @xfail and let's merge it.

PR #540, opened 2019-05-14
Add test for group install

j-mracek:
It test situation for group install when group is installed but not
available.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,1 +1,1 @@
-Dockerfile.f29
+Dockerfile.f30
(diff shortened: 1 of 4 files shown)


m-blaha:
Please, also scenario
`Scenario: A proper error message is displayed when I try to install a non-existent module using group syntax`
in module-install-errors.feature should be fixed. The error message has changed:
```
-    Warning: Module or Group 'NonExistentModule' does not exist.
+    Warning: Module or Group 'NonExistentModule' is not available.
```

PR #542, opened 2019-05-17
Use modifyrepo_c instead of modifyrepo

pkratoch:
The modifyrepo is not obsoleted and replaced by modifyrepo_c on Fedora <= 30.


pkratoch:
Sorry, it's already fixed.

PR #543, opened 2019-05-17
 Test module update with profiles depending on other-module rpms (RhBug:1647429)

pkratoch:
There is already a test for the bug 1647429, but this adds tests for updating modules with profiles that contain either non-modular rpms or rpms from another module.

### DIFF
--- a/dnf-behave-tests/features/module-update.feature
+++ b/dnf-behave-tests/features/module-update.feature
@@ -170,3 +170,111 @@ When I execute dnf with args "module update nodejs"
   And modules state is following
       | Module    | State     | Stream    | Profiles  |
       | nodejs    | enabled   | 11        |           |
+
+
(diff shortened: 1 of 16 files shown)


dnf-behave-tests/features/module-update.feature
```diff
@@ -170,3 +170,114 @@ When I execute dnf with args "module update nodejs"
   And modules state is following
       | Module    | State     | Stream    | Profiles  |
       | nodejs    | enabled   | 11        |           |
+
+
+@not.with_os=rhel__eq__8
```
m-blaha: All these tests are already passing on RHEL 8. Let's remove this tags and merge it.

pkratoch: I removed the tags. Thanks.

PR #547, opened 2019-05-23
Test that package with dependent package is not reinstalled

m-blaha:
$ dnf module install Module
$ dnf install Package  -- Package depends on other packages installed with Module
$ dnf history undo last-1

This command should fail instead of reinstalling packages from Module

https://github.com/rpm-software-management/dnf/pull/1392


j-mracek:
The test looks good

PR #553, opened 2019-06-06
remove temporary directories of tests in /tmp

lukash:
I'm getting thousands of these directories after running the tests for a while. They're eating my RAM as /tmp is mounted as tmpfs.

Fun fact is the default size in Fedora is half the RAM size. If you run two Fedora containers, does it equal your whole RAM size? And if you run more? :slightly_smiling_face: 

Is it ok to delete the directories? If someone needs them for post-mortem examination, should we add a switch somewhere to prevent the deletion?

As an alternative, I'd maybe preferred if they were created in a temporary directory inside the tests directory. There they are easy to find and delete (git clean) at will.

### DIFF
--- a/dnf-behave-tests/run-tests
+++ b/dnf-behave-tests/run-tests
@@ -11,6 +11,17 @@ ENV="$ENV -Ddnf_command=dnf-3"
 
 TAGS=""
 
+for opt in "$@"; do
+  shift
+  case $opt in
+    -p|--preserve)
+      ENV="$ENV -Dpreserve=yes"
+      continue
+      ;;
+  esac
+  set -- "$@" "$opt"
+done
+
 # run only tier1 tests
 #TAGS="$TAGS --tags @tier1"
 
--- a/dnf-behave-tests/features/environment.py
+++ b/dnf-behave-tests/features/environment.py
@@ -104,6 +104,7 @@ class DNFContext(object):
         self.module_platform_id = userdata.get("module_platform_id", DEFAULT_PLATFORM_ID)
         self.reposdir = userdata.get("reposdir", DEFAULT_REPOSDIR)
         self.repos_location = userdata.get("repos_location", DEFAULT_REPOS_LOCATION)
+        self.preserve_temporary_dirs = True if userdata.get("preserve", "no") in ("yes", "y", "1", "true") else False
         self.fixturesdir = FIXTURES_DIR
         self.disable_plugins = True
         self.disable_repos_option = "--disablerepo='*'"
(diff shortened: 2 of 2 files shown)


m-blaha:
I'm using these repositories a lot when investigating why the test has failed. You can just use this directory as an installroot, run whatever dnf command you need and watch the results.
So I would appreciate a way how to say "hey, keep all the temporary files around". Something like -r / -R option of dnf-testing.sh script. And I do not mind creating the temp directories inside the test directory assuming they are created in a subdirectory of the test directory.

lukash:
Ok, I'll see if we can pass a switch to behave, as it is where they're deleted.

How do you deal with the thousands of files in /tmp actually?

lukash:
I've added the switch, check it out!

m-blaha:
At the moment this is not working for me (the shift part is strange):

```
./run-tests -t~xfail -i features/module-defaults.feature
Loading config defaults from "./behave.ini"
Using defaults:
          color True
  show_snippets True
   show_skipped True
        dry_run False
    show_source True
   show_timings True
 stdout_capture True
 stderr_capture True
    log_capture True
 logging_format %(levelname)s:%(name)s:%(message)s
  logging_level 20
  steps_catalog False
        summary True
          junit False
          stage None
       userdata {'os': 'fedora__29'}
 default_format pretty
   default_tags 
scenario_outline_annotation_schema {name} -- @{row.id} {examples.name}
more_formatters {}
usage: behave [options] [ [DIR|FILE|FILE:LINE] ]+
behave: error: unrecognized arguments: features/module-defaults.feature
```

lukash:
Yeah, I'm not entirely sure how to use getopts for this - pick some arguments and keep others.

If we want to be strict here, we can't mix arguments for the shell script and for the inner behave like this. But I don't want to complicate it too much either... Any idea how to keep it simple and not too fragile?

dnf-behave-tests/run-tests
```diff
@@ -11,6 +11,16 @@ ENV="$ENV -Ddnf_command=dnf-3"
 
 TAGS=""
 
+OPTIND=1
+while getopts ":p" opt; do
+  case $opt in
+    p)
+      ENV="$ENV -Dpreserve=yes"
+      ;;
```
m-blaha: I thing moving shift from line 22 here (without argument) should help.

lukash: It does, but the `-p` has to be the first argument and if it isn't, the error message is somewhat confusing...

If we're fine with this for the sake of simplicity...

m-blaha: True. We can leave getopts out and do the parsing manually:
```
for opt in "$@"; do
  shift
  case $opt in
    -p|--preserve)
      ENV="$ENV -Dpreserve=yes"
      continue
      ;;
  esac
  set -- "$@" "$opt"
done
```

lukash: Clever! Pushed, thanks.

PR #556, opened 2019-06-11
Remove unneeded 'dnf makecache' steps

m-blaha:
They slow the test run down, so avoid them when possible

### DIFF
--- a/dnf-behave-tests/features/module-platform.feature
+++ b/dnf-behave-tests/features/module-platform.feature
@@ -14,7 +14,6 @@ Given I create file "/etc/os-release" with
     PRETTY_NAME="PseudoDistro 6 (dwm-team)"
     """
  And I do not set default module platformid
- And I execute dnf with args "makecache"
 
   
 Scenario: I can't enable module requiring different platform pseudo module
@@ -36,12 +35,13 @@ Scenario: I can't see pseudo-module in module listing
  Then stdout does not contain "pseudoplatform"
 
 
+@wip
 Scenario: I can't list info for the pseudo-module
  When I execute dnf with args "module info pseudoplatform"
  Then the exit code is 1
   And stdout matches line by line
    """
-   ?Last metadata
+   ^dnf-ci-pseudo-platform-modular\s+
    ^Unable to resolve argument pseudoplatform
    """
   And stderr is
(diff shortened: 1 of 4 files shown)


dnf-behave-tests/features/module-platform.feature
```diff
@@ -39,11 +38,7 @@ Scenario: I can't see pseudo-module in module listing
 Scenario: I can't list info for the pseudo-module
  When I execute dnf with args "module info pseudoplatform"
  Then the exit code is 1
-  And stdout matches line by line
-   """
-   ?Last metadata
-   ^Unable to resolve argument pseudoplatform
-   """
+  And stdout contains "Unable to resolve argument pseudoplatform"
```
lukash: Can you please check for possibly the exact output string, or keep the regexp if that's not possible (it's still more precise than a "contains").

dnf-behave-tests/features/module-platform.feature
```diff
@@ -93,13 +88,7 @@ Scenario: I can't update pseudo-module
...
  When I execute dnf with args "module remove pseudoplatform:6.0"
  Then the exit code is 0
-  And stdout matches line by line
-  """
-  ?Last metadata
-  ^Dependencies resolved.
-  ^Nothing to do.
-  ^Complete!
-  """
+  And stdout contains "Nothing to do."
```
lukash: Same here, check for exact match if possible.

m-blaha: Thanks, both are fixed now.

dnf-behave-tests/features/module-platform.feature
```diff
@@ -42,6 +41,7 @@ Scenario: I can't list info for the pseudo-module
   And stdout matches line by line
    """
    ?Last metadata
+   ^dnf-ci-pseudo-platform-modular\s+
```
lukash: A bit unrelated, but the `makecache` command changes the behaviour with regards to modules? How?

Also, is it not possible to simply use `And stdout is ...`? I believe the "?Last metadata" line is not needed. Unless  the `\s+` is actually a variable string, `And stdout is ...` is simpler and more effective?

m-blaha: It is not related to modularity per se. dnf-ci-pseudo-platform-modular is just the name of repository being downloaded. Using "std out is" step is not possible because this step do not accept regexp and the output of the command is not exactly the same for each run. It looks like this:

dnf-ci-pseudo-platform-modular                  1.6 MB/s | 1.7 kB     00:00    
Unable to resolve argument pseudoplatform

I removed the Last metadata part though.

lukash: Ok, merging.

I'm wondering why it sometimes downloads and sometimes doesn't though.

PR #560, opened 2019-06-12
Change bugzilla reference and fix scenario

m-blaha:
The originally referenced bug is not tested by this scenario.
Also the test was changed according the new bug fix.


lukash:
I've tried to rebase your branch and run the test, it now fails on:
```
    And stderr contains "Error: Transaction check error:"
      Assertion Failed: Stderr doesn't contain: Error: Transaction check error:
...
      Captured stderr:
      Error: Transaction test error:
...
```

(so the actual output now says _test_ instead of _check_, which is weird and I think it should really say _check_...)

Could you also change this to test for the exact output if possible? (I'm really annoyed by tests that do not check for the exact output, especially now since my colored diff is in place, but can't be used here and I need to keep squinting at the output to figure out what's wrong...)

PR #561, opened 2019-06-13
Test for repoquery --querytags option

m-blaha:


dnf-behave-tests/features/repoquery.feature
```diff
@@ -153,3 +153,12 @@ Given I use the repository "dnf-ci-thirdparty"
...
  """
+
+
+Scenario: dnf repoquery --querytags is working
+ When I execute dnf with args "repoquery --querytags"
+ Then the exit code is 0
+ And stdout contains lines
+ """
+ Available query-tags: use --queryformat ".. %{tag} .."
+ """
```
pkratoch: How about testing the exact match in this case?
The querytags are always printed in the same order and the lines are not wrapped with smaller terminal width, so it shouldn't be a problem.

"""
Available query-tags: use --queryformat ".. %{tag} .."

name, arch, epoch, version, release, reponame (repoid), evr,
debug_name, source_name, source_debug_name,
installtime, buildtime, size, downloadsize, installsize,
provides, requires, obsoletes, conflicts, sourcerpm,
description, summary, license, url

"""

m-blaha: No problem, done.

PR #565, opened 2019-06-20
Add tests for module install that must not change a stream

j-mracek:
Related to @bz1706215

### DIFF
--- a/dnf-behave-tests/features/module-enable-errors.feature
+++ b/dnf-behave-tests/features/module-enable-errors.feature
@@ -22,6 +22,40 @@ Scenario: Fail to enable a different stream of an already enabled module
     And stderr contains "The operation would result in switching of module 'nodejs' stream '8' to stream '10'"
     And stderr contains "Error: It is not possible to switch enabled streams of a module."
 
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/module-enable-errors.feature
```diff
@@ -22,6 +22,40 @@ Scenario: Fail to enable a different stream of an already enabled module
...
+   When I execute dnf with args "module install nodejs:10"
+   Then the exit code is 1
+    And modules state is following
+        | Module    | State     | Stream    | Profiles  |
+        | nodejs    | enabled   | 8         |           |
+    And stderr contains "The operation would result in switching of module 'nodejs' stream '8' to stream '10'"
+    And stderr contains "Error: It is not possible to switch enabled streams of a module."
+
+@bz1706215
+Scenario: Fail to install a different stream of an already enabled module
```
m-blaha: Please distinguish this scenario from the previous one by changing the name to something like
`Fail to install a different stream of an already enabled module using @module:stream syntax`

PR #567, opened 2019-06-25
Test for bz1722493

m-blaha:
Tests correct parsing of a value followed in config file by line
containing only whitespaces.
E.g.

gpgcheck=1
  <some spaces are here>

results in gpgcheck=1 not in gpgcheck='1\n'

### DIFF
--- a/dnf-behave-tests/features/config.feature
+++ b/dnf-behave-tests/features/config.feature
@@ -178,3 +178,26 @@ Scenario: Test usage of not existing config file
    When I execute dnf with args "list"
    Then the exit code is 1
     And stderr contains "Config file.*does not exist"
+
+
+@not.with_os=rhel__eq__8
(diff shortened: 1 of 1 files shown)


pkratoch:
At first I thought you are trying to check for white spaces in the ``config-manager --dump`` output, but you are actually testing that the default value from the [main] section is not used as a fallback. That's good, however, the last line in the commit message ("results in gpgcheck=1 not in gpgcheck='1\n'") is quite confusing, because the final result is the default value, not the '1\n'. Can you please improve the message, so that it's clear right away?

Otherwise, it looks good.

m-blaha:
Thanks Pavla, I tried to clarify the commit message to better describe the problem.

pkratoch:
Great, thanks! The tests are passing locally, merging.

PR #568, opened 2019-06-27
Adjust tests for module install without a default profile

j-mracek:
The new behavior of dnf that install commands requires to install a
profile must be also reflected in tests. In case that there is no
default provide, dnf provides a hint about available profiles and raises
an error.

The patch also deletes two scenario because they are nor completely
irrelevant and adjusting them to new behavior makes them redundant.

### DIFF
--- a/dnf-behave-tests/features/module-enable-errors.feature
+++ b/dnf-behave-tests/features/module-enable-errors.feature
@@ -31,7 +31,7 @@ Scenario: Fail to install a different stream of an already enabled module
     And modules state is following
         | Module    | State     | Stream    | Profiles  |
         | nodejs    | enabled   | 8         |           |
-   When I execute dnf with args "module install nodejs:10"
+   When I execute dnf with args "module install nodejs:10/minimal --skip-broken"
    Then the exit code is 1
     And modules state is following
         | Module    | State     | Stream    | Profiles  |
@@ -49,7 +49,7 @@ Scenario: Fail to install a different stream of an already enabled module using 
     And modules state is following
         | Module    | State     | Stream    | Profiles  |
         | nodejs    | enabled   | 8         |           |
-   When I execute dnf with args "install @nodejs:10"
+   When I execute dnf with args "install @nodejs:10/minimal --skip-broken"
    Then the exit code is 1
     And modules state is following
         | Module    | State     | Stream    | Profiles  |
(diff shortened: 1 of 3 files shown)


pkratoch:
@j-mracek, the tests are good, however, I would like to add some more scenarios.

Firstly (this is not entirely relevant to the bug), we don't have any scenario testing that dnf fails if there are multiple module specs and only one of them fails.

Secondly, what is the decision on a situation when there are no packages in the requested/default profile? Should it get installed or exit with error? In either case, I'd like to have tests for that (there is already one in module-install-default.feature).

And lastly, can a module profile be installed as a dependency? Can it be without specifying a profile (i.e. default profile should be picked)? I don't think it's possible, but I am asking to be sure.

I can write the tests, so that you are not testing your own patch :)

j-mracek:
> @j-mracek, the tests are good, however, I would like to add some more scenarios.
> 
> Firstly (this is not entirely relevant to the bug), we don't have any scenario testing that dnf fails if there are multiple module specs and only one of them fails.

Mere tests - better future :)

> 
> Secondly, what is the decision on a situation when there are no packages in the requested/default profile? Should it get installed or exit with error? In either case, I'd like to have tests for that (there is already one in module-install-default.feature).

When the profile has no packages, dnf returns 0 and mark profile as installed

> 
> And lastly, can a module profile be installed as a dependency? Can it be without specifying a profile (i.e. default profile should be picked)? I don't think it's possible, but I am asking to be sure.

No profile can be installed as a dependency. Only packages from profile can as a dependency of rpms.

> 
> I can write the tests, so that you are not testing your own patch :)

Thank you Pavla

pkratoch:
Thanks for your response. I will make a PR later on (it's related but not necessary for the bug), let's first merge yours.

Also, I have some suggestions after all.

Can you check the exact match in your test? Like this:
```
    And stderr is
        """
        No default profiles for module DnfCiModuleNoDefaults:stable. Available profiles: default
        Error: Problems in request:
        missing groups or modules: DnfCiModuleNoDefaults:stable
        """
```
Because you are checking the same lines anyway and this is better to ensure there are no additional errors. Thanks.

Also, some commit message corrections:
- "The new behavior of dnf that install commands requires to install a profile must be also reflected in tests." -> "The new behavior of dnf, that install command requires to install a profile, must be also reflected in tests."
- "default provide" -> "default profile"
- "nor completely irrelevant" -> "now completely irrelevant"

j-mracek:
@pkratoch Commit message was adjusted according to 2 of 3 suggestions. The last one was fixed little bit differently. Thank you for corrections.

pkratoch:
Great, thanks. The tests are passing locally with the patch, merging.

PR #570, opened 2019-07-04
script to enable test suite execution on non-x86_64 architectures

kkaarreell:
the script modifies the test suite files so it can be run on different architectures

### DIFF
--- a/dnf-behave-tests/arch-switch
+++ b/dnf-behave-tests/arch-switch
@@ -1,0 +1,30 @@
+#!/bin/bash
+#
+# the script modifies the test suite files so it can be run on different architectures
+# Usage:
+#  $ ./arch-switch
+#  $ ./dnf-testing.sh build
+#  $ ./dnf-testing.sh run
+
+set -xe
+
+PRIMARYARCH=`arch`
+
+if [ "$PRIMARYARCH" == "x86_64" ]; then
+  echo "No architecture changes needed for x86_64"
+  exit 0
+fi
+
+# these secondary archs are mostly useless as packages built won't be installable anyway
+# we need to skip test scenarios based on secondary architectures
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/arch-switch
```diff
@@ -0,0 +1,30 @@
...
+# we need to skip test scenarios based on secondary architectures
+[ "$PRIMARYARCH" == "ppc64le" ] && SECONDARYARCH=ppc
+[ "$PRIMARYARCH" == "s390x" ] && SECONDARYARCH=s390
+[ "$PRIMARYARCH" == "aarch64" ] && SECONDARYARCH=aarch
+
+echo "Using $PRIMARYARCH, $SECONDARYARCH"
+
+# substitution in testing files
+sed -i "s/x86_64/$PRIMARYARCH/g" $( grep -l -R x86_64 fixtures dnf-behave-tests/features )
+sed -i "s/i686/$SECONDARYARCH/g" $( grep -l -R i686 fixtures dnf-behave-tests/features )
```
lukash: Is this is just a temporary hack? Obviously the issue is hard to fix, but we should still solve it properly and sed really isn't it...

kkaarreell: Yes, it is a hack that allows us NOW to run the test suite on other architectures. From these reasons the script is only attached to the project, not called directly.

lukash: Okay. It would have been a good idea to mention that in the commit message, otherwise people might think you mean this to stay there... :slightly_smiling_face:

kkaarreell: Well, I think that there is nothing in the test suite that one could say it will last there forever. 
I asked Marek whether I should add the script to the repo or keep it downstream. He preferred to have it upstream and hence the PR. And if objections against the approach used in the script would force someone to implement it nicely, the better.

PR #575, opened 2019-07-12
Added test for bz1686602, reposync --urls

m-blaha:


jamescassell:
Would be good to have a test also in combination w/ the `--download-metadata` option.

PR #577, opened 2019-07-15
Repoquery command tests

lukash:
This big PR adds (reasonably) comprehensive tests for the repoquery command. It's meant to cover all the functionality (a few niche bits may be missing).

The reason for this is I had to make some big changes to the behavior of the command and wanted to make sure I'm not breaking anything. The last commit adds tests for that.

Note this shares some commits with my previous PR here: https://github.com/rpm-software-management/ci-dnf-stack/pull/566

That should go in first.

Not all the tests are passing. They're marked with @xfail and there is a TODO for each with a comment on what I think. I'd like to discuss how those should be resolved (i.e. desired behavior or bug...).

The tests completely replace the previous repoquery tests (apart from repoquery-cache.feature which perhaps still remains to move) and each feature file has its own separate repository, which makes it independent of the other tests.

### DIFF
--- a/dnf-behave-tests/features/repoquery/rich-deps.feature
+++ b/dnf-behave-tests/features/repoquery/rich-deps.feature
@@ -32,6 +32,27 @@ Scenario: repoquery --whatrequires for "(a1-prov1 if b1)"
       """
 
 
+@bz1534123
+@bz1698034
+Scenario: repoquery --whatrequires NAME for "(b1-prov2 >= 1.0 with b1-prov2 < 2.0)"
+ When I execute dnf with args "repoquery --whatrequires b1"
+ Then the exit code is 0
+  And stdout is
+      """
+      c1-0:1.0-1.x86_64
+      """
+
+@bz1534123
+@bz1698034
+Scenario: repoquery --whatrequires PROVIDE_NAME = VERSION for "(b1-prov2 >= 1.0 with b1-prov2 < 2.0)"
+ When I execute dnf with args "repoquery --whatrequires 'b1-prov2 = 1.0'"
+ Then the exit code is 0
+  And stdout is
+      """
+      c1-0:1.0-1.x86_64
+      """
+
+
 # a1-1.0: Conflicts: ((b1 and x1) or c1)
 Scenario: repoquery --whatconflicts for "((b1 and x1) or c1)"
  When I execute dnf with args "repoquery --whatconflicts b1"
@@ -70,6 +91,26 @@ Given I successfully execute dnf with args "install x1"
   And stdout is
       """
       a1-0:1.0-1.x86_64
+      """
+
+@bz1534123
+@bz1698034
+Scenario: repoquery --whatconflicts for "(d1-prov1 >= 1.0 with d1-prov1 < 2.0)"
+ When I execute dnf with args "repoquery --whatconflicts d1-1.0"
+ Then the exit code is 0
+  And stdout is
+      """
+      c1-0:1.0-1.x86_64
+      """
+
+@bz1534123
+@bz1698034
+Scenario: repoquery --whatconflicts PROVIDE_NAME = VERSION for "(d1-prov1 >= 1.0 with d1-prov0 < 2.0)"
+ When I execute dnf with args "repoquery --whatconflicts 'd1-prov1 = 1.0'"
+ Then the exit code is 0
+  And stdout is
+      """
+      c1-0:1.0-1.x86_64
       """
 
 
@@ -114,6 +155,26 @@ Scenario: repoquery --whatsuggests for "((b1 with b1-prov2 > 1.7) or (c1 <= 1.0 
   And stdout is
       """
       a1-0:1.0-1.x86_64
+      """
+
+@bz1534123
+@bz1698034
+Scenario: repoquery --whatsuggests for "(d1-prov1 >= 1.0 with d1-prov1 < 2.0)" - only d1-1.0 should match
+ When I execute dnf with args "repoquery --whatsuggests d1"
+ Then the exit code is 0
+  And stdout is
+      """
+      b1-0:1.0-1.x86_64
+      """
+
+@bz1534123
+@bz1698034
+Scenario: repoquery --whatsuggests with provide for "(d1-prov1 >= 1.0 with d1-prov1 < 2.0)" - only b1-1.0 should match
+        When I execute dnf with args "repoquery --whatsuggests 'd1-prov1 = 1.0'"
+ Then the exit code is 0
+  And stdout is
+      """
+      b1-0:1.0-1.x86_64
       """
 
 
--- a/dnf-behave-tests/fixtures/specs/repoquery-rich-deps/b1-1.0-1.spec
+++ b/dnf-behave-tests/fixtures/specs/repoquery-rich-deps/b1-1.0-1.spec
@@ -9,6 +9,7 @@ Provides:       b1-prov1
 Provides:       b1-prov2 = 1.0
 
 Enhances:       (a1 unless x1 else c1)
+Suggests:       (d1-prov1 >= 1.0 with d1-prov1 < 2.0)
 
 Summary:        Rich deps package.
 
--- a/dnf-behave-tests/fixtures/specs/repoquery-rich-deps/c1-1.0-1.spec
+++ b/dnf-behave-tests/fixtures/specs/repoquery-rich-deps/c1-1.0-1.spec
@@ -5,7 +5,12 @@ Release:        1
 License:        Public Domain
 URL:            None
 
+Provides:       c1-prov1 = 1.0
+
 Requires:       (a1-prov1 if b1)
+Requires:       (b1-prov2 >= 1.0 with b1-prov2 < 2.0)
+
+Conflicts:      (d1-prov1 >= 1.0 with d1-prov1 < 2.0)
 
 Recommends:     (b1 and a1)
 
--- a/dnf-behave-tests/fixtures/specs/repoquery-rich-deps/c1-2.0-1.spec
+++ b/dnf-behave-tests/fixtures/specs/repoquery-rich-deps/c1-2.0-1.spec
@@ -8,6 +8,9 @@ URL:            None
 Provides:       c1-prov1 = 2.0
 
 Requires:       (a1-prov1 if b1)
+Requires:       (b1-prov2 >= 2.0 with b1-prov2 < 3.0)
+
+Conflicts:      (d1-prov1 >= 2.0 with d1-prov1 < 3.0)
 
 Recommends:     (b1 and a1)
 
--- a/dnf-behave-tests/fixtures/specs/repoquery-rich-deps/d1-1.0-1.spec
+++ b/dnf-behave-tests/fixtures/specs/repoquery-rich-deps/d1-1.0-1.spec
@@ -1,0 +1,17 @@
+Name:           d1
+Version:        1.0
+Release:        1
+
+License:        Public Domain
+URL:            None
+
+Provides:       d1-prov1 = 1.0
+
+Summary:        Rich deps package.
+
+%description
+Dummy.
+
+%files
+
+%changelog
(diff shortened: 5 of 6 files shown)


dnf-behave-tests/features/repoquery/globs.feature
```diff
@@ -0,0 +1,180 @@
...
+
+Background:
+ Given I use the repository "repoquery-globs"
+
+
+# <name> globs
+# TODO it's weird this doesn't actually work (lists nothing)
+@xfail
+Scenario: repoquery *
+ When I execute dnf with args "repoquery *"
```
dmach: replace * with '*' to avoid shell expansion

dnf-behave-tests/features/repoquery/globs.feature
```diff
@@ -0,0 +1,180 @@
...
+  And stdout is
+      """
+      desktop-utils-1:1.23.9-1.src
+      desktop-utils-1:1.23.9-1.x86_64
+      """
+
+# TODO this seems to incorrectly match many more packages
+@xfail
+Scenario: repoquery *top[-a-f]*-1.[!1-4]*
+ When I execute dnf with args "repoquery *top[-a-f]*-1.[!1-4]*"
```
dmach: Subject considers '-' in the [!1-4] range as NEVRA separator, which probably causes the problem.
```
>>> import dnf
>>> subj = dnf.subject.Subject("*top[-a-f]*-1.[!1-4]*")
>>> for i in subj.get_nevra_possibilities():
...     print(i.name, i.epoch, i.version, i.release, i.arch)
... 
*top[-a-f]*-1.[!1-4]* None None None None
*top[-a-f]* None 1.[!1 4]* None
*top[-a-f]*-1.[!1 None 4]* None None
# ^^^ last 2 lines are most likely wrong
```

lukash: Yes, funny thing is the correct variant is not actually there at all. I still haven't looked up exactly where the matching goes wrong here, but in general, I think we have a problem.

The forms are matched as regular expressions and it is not (AFAIK) possible to handle dashes nested inside `[]` as a special case in regexes. I think we have two options in general:

1. use something more powerful for the regexes, like [boost xpressive](https://github.com/lukash/librepo/commit/8d4dd9f8618fa64b37d17c0e65e9180294fa9d71), which seems to have some facility for nested brackets and more. But I'm not sure how viable it is to use boost. And it doesn't look to be best for the job anyway, so bringing in boost just for that isn't great either.

2. Write a parser that will parse the NEVRAs. It could have potential benefits of better matching than the somewhat crude regexes, but wouldn't probably mesh well with the current "forms" interface. Also, it's somewhat nontrivial...

lukash: Actually, I think I have an idea how to work around it. I'll give it a go later.

lukash: So actually this test in particular was not affected by the faulty regex parsing. The regex forms didn't match anything and there's a last resort branch that tries to match the argument against the full NEVRA of the packages as-is. The pattern was actually matched correctly, the last bit was matching against the architecture which I didn't think of. I've fixed the test and pushed it as a separate commit temporarily (so that it is clear what I changed), I'll squash it later.

dnf-behave-tests/features/repoquery/main.feature
```diff
@@ -0,0 +1,467 @@
...
+      top-a-1:1.0-1.src
+      top-a-1:1.0-1.x86_64
+      top-a-1:2.0-1.src
+      top-a-1:2.0-1.x86_64
+      """
+
+# TODO almost the same as above (note the -), but this one fails
+@xfail
+Scenario: repoquery NAME-VERSION NAME-EPOCH:VERSION_GLOB2-RELEASE
+ When I execute dnf with args "repoquery bottom-a1-1.0 top-a-1:[1-2].0-1"
```
lukash: However, this test is affected by the imperfect regex parsing. Here the dash in the regex causes the forms to also not match anything, and the last-resort full NEVRA match doesn't work either, because the glob argument is missing architecture and therefore is not full NEVRA.

Let's not let this particular issue hold back the PR though. I still want to investigate it a bit more and then I will create a bug for it so that we can work on it separately.

dnf-behave-tests/features/repoquery/deps.feature
```diff
@@ -0,0 +1,565 @@
...
+Scenario: repoquery --requires --resolve --recursive NAME-VERSION
+ When I execute dnf with args "repoquery --requires --resolve --recursive top1-2.0"
+ Then the exit code is 0
+  And stdout is
+      """
+      bottom1-1:2.0-1.x86_64
+      bottom2-1:1.0-1.x86_64
+      middle1-1:2.0-1.x86_64
+      middle2-1:2.0-1.x86_64
+      """
```
dmach: middle1-1:2.0-1.x86_64 which is in the transaction requires /a/bottom4-file
the file is provided bottom4-1:1.0-1.x86_64
I believe it should be added to the output.

bottom3-1:2.0-1.x86_64 is probably pulled it because it obsoletes 'bottom2 = 1:2.0-1'
Probably also expected as the obsolete is in place.

dnf-behave-tests/features/repoquery/main.feature
dmach: Agreed. This will require more work, let's not block the PR on this.

dnf-behave-tests/features/repoquery/rich-deps.feature
```diff
@@ -0,0 +1,213 @@
...
+      """
+
+
+# a1-1.0: Supplements: ((b1 < 2.0 with b1-prov2 > 1.7) or (c1 > 1.0 without c1-prov1 > 0.5))
+# TODO neither of the above two expressions (the "with" and "without" ones) holds true,
+# but the two commands below still list a1 as if it matched
+# note: even if simplified to e.g. (b1 with nonexisting_provide) it still evaluates to true.
+@xfail
+Scenario: repoquery --whatsupplements for "((b1 < 2.0 with b1-prov2 > 1.7) or (c1 > 1.0 without c1-prov1 > 0.5))"
+ When I execute dnf with args "repoquery --whatsupplements b1"
```
dmach: `repoquery --whatsupplements b1`
returns
`a1-0:1.0-1.x86_64`

Isn't it correct?

`repoquery --supplements a1-0:1.0-1.x86_64`
returns
`((b1 < 2.0 with b1-prov2 > 1.7) or (c1 > 1.0 without c1-prov1 > 0.5))`

The first condition is `true` because `--whatsupplements b1` was without a version.

For example `repoquery --whatsupplements 'b1 > 10'` returns nothing

dnf-behave-tests/features/repoquery/globs.feature
lukash: Duh.

dnf-behave-tests/features/repoquery/deps.feature
lukash: > middle1-1:2.0-1.x86_64 which is in the transaction requires /a/bottom4-file
> the file is provided bottom4-1:1.0-1.x86_64
> I believe it should be added to the output.

Yes, that was added later...

> bottom3-1:2.0-1.x86_64 is probably pulled it because it obsoletes 'bottom2 = 1:2.0-1'
> Probably also expected as the obsolete is in place.

Ok, (pulling in obsoletes might be a thing to write a test case for) but I don't see anything depending on `bottom2 = 1:2.0-1`?

dnf-behave-tests/features/repoquery/rich-deps.feature
lukash: This may be a peculiarity of how the rich deps are resolved, but `b1 < 2.0` leaves only a single package, wich is `b1-1.0` and that one has a provide `b1-prov2 = 1.0`, which doesn't hold true for the second part of the `with` statement. So there is no `b1` package that fulfils the requirement.

But thinking about it now, this is probably how it works, since I'm not providing enough information in the query, so dnf lists what possibly supplements b1.

lukash: Btw. this patch is related: https://github.com/openSUSE/libsolv/commit/85f82095c379ca46adecb6a137ce2bd47eb04c47

But AFAICS it won't handle this case because the dependencies aren't on the same provide name...

lukash:
I've pushed fixes for the resolved failures. What remains is:
2 TODOs about --whatconflicts and --whatobsoletes not supporting globs - this is basically about a decision whether we want it to stay that way or flip to glob support (easy to do)

The above discussed TODO with `bottom3-1:2.0-1.x86_64` incorrectly listed in a `--recursive --resolve` query, which seems to be a bug and I'll try to investigate it a bit more later.

A TODO in the last commit adding tests for the version range fix, which doesn't work for querying provides (it's in `rich-deps.feature`), which should be fixed by https://github.com/openSUSE/libsolv/commit/85f82095c379ca46adecb6a137ce2bd47eb04c47 in the next release of `libsolv`.

lukash:
The `bottom3-1:2.0-1.x86_64` was included due to not specifying epoch for `middle2` dependency in `top1-1:2.0-1.spec`:
```
Requires: middle2 > 1:1.5
```

I've fixed that and also verified that git version of `libsolv` does fix the version range check in provides.

@dmach after resolving the glob support in `--whatobsoletes` and `--whatconflicts` we can merge everything except the last commit, which is specific to the rich-deps fix and should be merged together with the other PRs. So lets sort out the globs and do it this way.

j-mracek:
Only two scenarios are failing without patches for dnf/libdnf, therefore they are incorrectly marked or do not test it properly.

Please could you also add test that without patches will return multiple packages (incorrectly more results like in bug report) but with patches only correct result (a single entry).

lukash:
> Only two scenarios are failing without patches for dnf/libdnf, therefore they are incorrectly marked or do not test it properly.

Right... out of the 6 new scenarios, the three that test for a reldep (e.g. `--whatrequires 'b1-prov2 = 1.0'`) were actually fixed by a change in libsolv after I discovered using the new libsolv function doesn't work for these cases. Not sure if they should or shouldn't be tagged with the bug numbers, but I'm inclined to say yes.

> Please could you also add test that without patches will return multiple packages (incorrectly more results like in bug report) but with patches only correct result (a single entry).

I _think_ this is what the `repoquery --whatrequires NAME for "(b1-prov2 >= 1.0 with b1-prov2 < 2.0)"` test was doing initially, but the details are unfortunately gone from my head. But from looking at the test data, it should be matching both `c1-1.0` and `c1-2.0` without the fix. Now it seems to be working correctly without the fix too... Maybe the change in libsolv I've mentioned in the first paragraph actually applies to this too? I'm still thinking about how to verify this...

lukash:
So on old dnf (Fedora 30 without updates), the following test fails with listing two packages, which is what the original bug was reported for. I assume the fix for the reldep version incidentally fixed this in addition to my changes (I'll just also add a note that the fix in libsolv as well as the new libsolv function that is utilized in my PRs are very single-purpose in handling just this specific (and rather common) case and don't solve rich dependencies in general).

```gherkin
  @bz1534123 @bz1698034
  Scenario: repoquery --whatrequires NAME for "(b1-prov2 >= 1.0 with b1-prov2 < 2.0)"  # features/repoquery/rich-deps.feature:37
    Given I use repository "repoquery-rich-deps"                                       # features/steps/repo.py:118 0.000s
    When I execute dnf with args "repoquery --whatrequires b1"                         # features/steps/cmd.py:106 0.180s
    Then the exit code is 0                                                            # common/output.py:58 0.000s
    And stdout is                                                                      # common/output.py:70 0.000s
      """
      c1-0:1.0-1.x86_64
      """
      Assertion Failed: Stdout is not: c1-0:1.0-1.x86_64
      Captured stdout:
      expected           |  found
      c1-0:1.0-1.x86_64  |  c1-0:1.0-1.x86_64
                         |  c1-0:2.0-1.x86_64
```

PR #578, opened 2019-07-16
Add ftp server and complete install-path.feature with http and ftp paths

kontura:
This adds dependency on **pyftpdlib** as Python does not provide any FTP server in the standard library. However it seems to integrate well into the style set up by http server.


lukash:
LGTM. I'd merge this but I'm not sure the new dependency won't break something somewhere. I think you need to add it at least to the docker files? (I couldn't test them atm.) Assigning to @m-blaha just to check the dependency is fine.

m-blaha:
You need to add installing python3-pyftpdlib into Fedora Dockerfiles.
Also please update RHEL8 Dockerfile (currently on internal gitlab) and let now QA to update their RHEL8 Dockerfile after this PR gets merged. There is no pyftpdlib package on RHEL so you will need to use pip (it is already used for behave itself). Hopefully the same would be usable for Centos7.

kontura:
I have added the dependency to fedora dockerfiles here and created a [PR for the rhel one](https://gitlab.cee.redhat.com/rpm-software-management/ci-dnf-stack/merge_requests/2). I will also notify @janblazek about this after it gets merged. Thank you.

PR #580, opened 2019-07-30
Add test for bz1709453

j-mracek:


j-mracek:
The test is for https://github.com/rpm-software-management/dnf/pull/1446

PR #582, opened 2019-08-01
Test update installonly package from lower priority repo

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1733582

### DIFF
--- a/dnf-behave-tests/features/repo-priorities.feature
+++ b/dnf-behave-tests/features/repo-priorities.feature
@@ -60,6 +60,19 @@ Scenario: Upgrade an RPM from the highest-priority repository
         | upgrade       | flac-0:1.3.3-2.fc29.x86_64                |
 
 
+@xfail
+@bz1733582
+Scenario: Do not upgrade installonly package from lower-priority repository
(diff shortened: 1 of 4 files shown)


lukash:
Wow :slightly_smiling_face: do we really need three 7500 lines long files to test this? AFAIK the only thing you need to test installonly package is `Provides: installonlypkg(kernel)` (or something similar). The rest is just noise...

m-blaha:
@lukash: you are right. I've just copied overgrown kernel package from some other repository. I'll strip them down to necessary minimum.

lukash:
@m-blaha strictly speaking, you can strip them even more, as you don't need the `-core` subpackage or the other provides, all you really need is a single package with the one `Provides: installonlypkg(kernel)`.

But if you don't want to tune this further, let me know and I'll merge it as it is.

m-blaha:
Yes, but I did want to keep it close to the real kernel packages.

lukash:
I wouldn't think there's a value in that, it could even hide an issue if the implementation was tied to the kernel package structure more than it should.

Anyway...

PR #581, opened 2019-07-31
Add tests for keeping userinstalled setting when upgrading obsoleted RPM

pkratoch:

PR #585, opened 2019-08-16
Inherit reason from obsoleted packages (bz1672618)

dmach:


dmach:
Related to: rpm-software-management/libdnf#772
Related to: rpm-software-management/dnf/pull/1459

pkratoch:
This bug had different behavior with --best and with --nobest. Your patch seems to work for both, however, there should also be both test scenarios. Can you please add --best to the current one (so that default doesn't matter) and add the same scenario with --nobest? I can then close PR #581.

BTW, the first commit is similar to this one: https://github.com/rpm-software-management/ci-dnf-stack/pull/583/commits/8b31e4b1e82efa57948fe8c8a81c691c9bd4f2bf

pkratoch:
@dmach this is here for quite some time. Do you have any plans to move this further? Alternatively, I can make the changes I proposed in another PR and close this one if you are too busy.

PR #589, opened 2019-08-20
Add tests for xml:base and destdir in dnf (download)

kontura:


dnf-behave-tests/features/install-xml-base.feature
```diff
@@ -0,0 +1,147 @@
...
+  name=testrepo
+  baseurl=file://{context.dnf.installroot}/temp-repo
+  enabled=1
+  gpgcheck=0
+  """
+  And I do not set reposdir
+  And I use the repository "testrepo"
+  And I execute bash with args "createrepo_c --baseurl file://{context.dnf.installroot}/temp-repo/xml_base ." in directory "{context.dnf.installroot}/temp-repo"
+ Then the exit code is 0
+ Then I copy directory "{context.dnf.installroot}/temp-repo/dnf-ci-fedora" to "/temp-repo/xml_base/dnf-ci-fedora"
```
lukash: This is quite confusing... One, I can't easily tell, but all these steps look to be the test setup, so they should be `Given`, right?

Also, these tests have a really long setup, I haven't figured out all the things that are going on here, but I'm wondering if some better step definition would help with this...

kontura: Yes I agree, I will try to abstract several of the steps into new step with hopefully more descriptive name.

kontura:
I have added two new steps, I think the scenarios are clearer now.

kontura:
Rebased onto current master

dnf-behave-tests/features/steps/repo.py
```diff
@@ -200,3 +213,51 @@ def step_countme(context, quantifier):
...
+
+@behave.step("I make packages from \"{repo}\" repository accessible via {rtype:repo_type}")
+def make_repo_accessible(context, repo, rtype):
+    assert (hasattr(context, 'httpd') or hasattr(context, 'ftpd')), \
+        'Httpd or Ftpd fixture not set. Use @fixture.httpd or @fixture.ftpd tag.'
+    if rtype == "http":
+        host, port = context.httpd.new_http_server(context.dnf.repos_location)
+    elif rtype == "ftp":
+        host, port = context.ftpd.new_ftp_server(context.dnf.repos_location)
+    context.dnf._set(rtype + "-" + repo + "-port", port)
```
lukash: I think this here is a different approach to storing the ports than what ended up in _"I use the X repository based on Y"_:
```python
context.dnf.ports[rtype + "-" + repo] = port
```

dnf-behave-tests/features/steps/repo.py
```diff
@@ -200,3 +213,51 @@ def step_countme(context, quantifier):
...
+                  name={name}
+                  baseurl=file://{loc}
+                  enabled=1
+                  gpgcheck=0
+                  \"\"\"
+              And I do not set reposdir
+              And I use the repository "{name}"
+            """.format(
+                loc=prepend_installroot(context, target),
+                name=os.path.basename(target)))
```
lukash: So TBH I really don't like this, this is getting a bit extreme :slightly_smiling_face: 

One thing would be to call functions instead of the steps like this (potentially moving the step code out to functions to facilitate this). But I'm thinking we should kind of consolidate the repo creation into some more concise steps, which includes moving the initial repo creation from `fixtures/specs/build.sh` into the steps as well.

It can then be part of the _I use the X repository_, which would be extended to specify the type, e.g. _I use http repository "dnf-ci-fedora"_, which would call createrepo_c to create the repository in a unique location (probably symlinking the package directories? haven't figured this one out exactly yet) and create the `.repo` config file.

I've got a small PoC which seems to be working for the simple cases, so I think this should work well... :slightly_smiling_face:

lukash: @m-blaha agreed. For the performance of generating repos for each step, I'm planning to cache if the repo was generated for one run of the tests, so during each tests run each repo (of each type) would be generated only once.

dnf-behave-tests/features/install-xml-base.feature
```diff
@@ -0,0 +1,85 @@
...
+ 
+@not.with_os=rhel__eq__8
+@fixture.httpd
+Scenario: Install from local repodata that have packages with xml:base pointing to a remote as well as local packages
+Given I make packages from "dnf-ci-fedora" repository accessible via http
+  And I copy repository "dnf-ci-fedora" for modification
+  And I copy repository "dnf-ci-thirdparty" for modification
+  And I execute "createrepo_c --baseurl http://localhost:{context.dnf.ports[dnf-ci-fedora]} ." in "/{context.dnf.repos[dnf-ci-fedora].path}"
+  And I execute "mergerepo_c --repo file://{context.dnf.repos[dnf-ci-fedora].path} --repo file://{context.dnf.repos[dnf-ci-thirdparty].path}" in "{context.dnf.installroot}"
+  And I use repository "dnf-ci-fedora" with configuration
```
lukash: This is sort-of abusing the "I use repository ..." step by using a different repo and injecting your baseurl... Seeing this I've realized there are other places where this sort of thing is required and whipped up https://github.com/rpm-software-management/ci-dnf-stack/pull/652, which adds a step to create a custom repo config.

dnf-behave-tests/features/steps/repo.py
```diff
@@ -298,3 +298,14 @@ def step_countme(context, quantifier):
         | path                     |
         | */metalink.xml?countme=1 |
     """.format(quantifier))
+
+
+@behave.step("I make packages from \"{repo}\" repository accessible via {rtype:repo_type}")
```
lukash: Can you please phrase it like:
> I make packages from repository \"{repo}\" accessible ...

i.e. flip the "repo" and repository words, that way sounds better in english and is consistent with the other steps.

dnf-behave-tests/features/plugins-core/download-binary-xml-base.feature
```diff
@@ -0,0 +1,69 @@
+Feature: dnf download command on packages with xml:base
```
lukash: Same comments I had for `install-xml-base.feature` apply for this file.

kontura:
I addressed all of your points. Also I would like to point out that I changed the location from which we start http/ftp servers, details in the first commit.

dnf-behave-tests/features/steps/file.py
```diff
@@ -66,13 +66,21 @@ def step_delete_directory(context, dirpath):
...
 def file_exists(context, filepath):
     full_path = prepend_installroot(context, filepath)
-    ensure_file_exists(full_path)
+    result = glob.glob(full_path)
+    if len(result) > 1:
+        raise AssertionError("File path %s matches multiple files: \n%s" % (filepath, '\n'.join(result)))
+    elif len(result) < 1:
+        raise AssertionError("File path %s doesn't match any file." % (filepath))
+    else:
+        ensure_file_exists(result[0])
```
lukash: Is this call really needed if the glob matched the path? (If `ensure_file_exists()` called `os.path.isfile()` it would be an argument, but it just calls `os.path.exists()`)

kontura: I was unsure about it as well. But you are right since it just checks whether it exists the result will be the same (it will still succeed even with directories). I have remove the call.

PR #596, opened 2019-08-23
added test scenario for bz1700529

kkaarreell:


dnf-behave-tests/fixtures/repos.d/dnf-ci-fifthparty-modular.repo
```diff
@@ -0,0 +1,5 @@
+[dnf-ci-fifthparty-modular]
```
lukash: > fifthparty

This is how structuring test cases and their data gets out of hand :slightly_frowning_face:

PR #599, opened 2019-08-26
Add tests for the --setopt option used for repo configs

pkratoch:


dnf-behave-tests/fixtures/specs/setopt-1/wget-1.19.5-5.fc29.spec
```diff
@@ -0,0 +1,24 @@
...
+Provides: bundled(gnulib) 
+
+%description
+GNU Wget is a file retrieval utility which can use either the HTTP or
+FTP protocols. Wget features include the ability to work in the
+background while you are logged out, recursive retrieval of
+directories, file name wildcard matching, remote file timestamp
+storage and comparison, use of Rest with FTP servers and Range with
+HTTP servers to retrieve files over slow or unstable connections,
+support for Proxy servers, and configurability.
```
lukash: I know storage is cheap nowadays.. but why not ditch the description? :slightly_smiling_face: 

Also possibly (if you want) simplify the versions, since they are not important, to something one doesn't need to squint at?

pkratoch: Ok, ditched and simplified. It was a copy-paste from dnf-ci-fedora repo.

PR #604, opened 2019-08-28
dnf nightly build: reenable createrepo_c and add drpm

kontura:
Because createrepo_c depends on drpm.

Maybe we should wait with this until we release drpm 0.4 so that if there are any problems we can just disable the drpm nightly build.


kontura:
drpm 0.4 is available in f30

PR #605, opened 2019-08-28
Add tests for command aliases availability

pkratoch:


dnf-behave-tests/features/command-aliases.feature
```diff
@@ -0,0 +1,104 @@
...
+
+
+Scenario: "nonexistent" is not an alias for any dnf command
+   When I execute dnf with args "nonexistent -h"
+   Then the exit code is 1
+    And stderr contains "No such command: nonexistent"
+
+
+Scenario Outline: "<alias>" is an alias for "<command>"
+   When I execute dnf with args "<alias> -h"
```
lukash: The scenario doesn't even test the `<command>` part of the examples... Looking at the output it seems every help text starts with `usage: dnf <command> ` (followed by the list of options). Perhaps we could check for that?

dnf-behave-tests/features/command-aliases.feature
```diff
@@ -0,0 +1,104 @@
...
+        | updateinfo          | summary-updateinfo           |
+        | upgrade             | upgrade                      |
+        | upgrade             | update                       |
+        | upgrade             | upgrade-to                   |
+        | upgrade             | update-to                    |
+        | upgrade             | localupdate                  |
+        | upgrade             | up                           |
+        | upgrade-minimal     | upgrade-minimal              |
+        | upgrade-minimal     | update-minimal               |
+        | upgrade-minimal     | up-min                       |
```
lukash: I'm wondering about the purpose and usefulness of this test though... We will also have to maintain this list, mainly not forget to add new commands here. Which is surely bound to happen...

pkratoch: It's one of tests for [bug 1690289](https://bugzilla.redhat.com/show_bug.cgi?id=1690289):
https://github.com/rpm-software-management/dnf/commit/da52d0ee8ec53fc67477a448ef9da73e32ce2cf2
  MED - addition of multiple command aliases, we should create test to prevent future regressions

I am also not sure how likely it's to happen that some alias disappears by accident and the tests catch that.

dnf-behave-tests/features/command-aliases.feature
pkratoch: Hey, that's a good idea! I've been thinking that checking the functionality is too much work, so I would only check that it's an existing alias, but I haven't thought of checking the help for the main command. I've added it, thanks.

dnf-behave-tests/features/command-aliases.feature
```diff
@@ -0,0 +1,105 @@
...
+Scenario: "nonexistent" is not an alias for any dnf command
+   When I execute dnf with args "nonexistent -h"
+   Then the exit code is 1
+    And stderr contains "No such command: nonexistent"
+
+
+Scenario Outline: "<alias>" is an alias for "<command>"
+   When I execute dnf with args "<alias> -h"
+   Then the exit code is 0
+    And stdout contains "usage: {context.dnf.dnf_command} <command>"
```
lukash: Hmm, the `{context.dnf.dnf_command}` doesn't work if you have a wrapper script around dnf, which e.g. I do :slightly_frowning_face:. So that isn't great...

dnf-behave-tests/features/command-aliases.feature
lukash: Yeah, sure, I suppose this does the job.

dnf-behave-tests/features/command-aliases.feature
pkratoch: Well, this is unpleasant. It's not so much of a problem for this PR as it is for https://github.com/rpm-software-management/ci-dnf-stack/pull/612 .

I don't see any way of how to get the command printed by help other than running the help. However, I could remove the test 'I execute dnf with args' and leave only the specific 'I execute "dnf" with args' and other variants. I think it would be sufficient. In this PR, I could replace it with ".*", because it's not important part of this test. What do you think?

lukash: Yeah, for #612 the 'I execute "dnf" with args' tests should be sufficient I think.

I'm not fond of the `.*` but it's probably the best option.

I could live with the test failing and I could also work around the issue by e.g. setting up a symlink to dnf that would have the same name as the wrapper script :slightly_smiling_face: but it's a bit cumbersome.

Not sure if others use such wrapper script, but you (AFAIK) need it to run the ci stack against git and I find it very useful, it's the only way I test majority of the time nowv (because it's so fast).

pkratoch: Ok, I removed the test that uses 'I execute dnf with args' from the https://github.com/rpm-software-management/ci-dnf-stack/pull/612 and replaced the command with ".*" here.

lukash:
Cheers!

PR #608, opened 2019-08-29
Update of modular errors generated by libsolv

j-mracek:
In some cases libsolv could return same errors in different order
therefore testing message with `Problem: ` for the first message could
result in false negative results.


j-mracek:
Closed in favor of https://github.com/rpm-software-management/ci-dnf-stack/pull/614

PR #610, opened 2019-08-29
Repoquery tests

lukash:
These are the new / consolidated tests for the `repoquery` command, originally from: https://github.com/rpm-software-management/ci-dnf-stack/pull/577

But without the tests for the rich dependencies fix, so these are mergeable without any dependencies. The first six commits are as they were in the original PR, I've added the last two commits here as the last bit of consolidation of the original `repoquery` tests.

There is only one outstanding issue, marked with two TODOs in `dnf-behave-tests/features/repoquery/deps.feature` and that is `--whatconflicts` and `--whatobsoletes` not supporting globs. Should be an easy fix, we just need to agree on the behavior. Then these can be merged.

Also requires https://github.com/rpm-software-management/dnf/pull/1469 for a small fix of the `--queryformat` option.


lukash:
I've removed the tests for `--queryformat` which depend on https://github.com/rpm-software-management/dnf/pull/1469, so this is now independent.

j-mracek:
The review of such a change is not easy because there is nearly no possibility to know what is new and what was moved. It requires to have a separate commit for movement and then commit for changes.

Also I don't think that the movement of tests is a good idea because it destroys history and there is only limited benefit based on personal preferences.

lukash:
Yeah, I should have done it better. Sorry about that. I'll see if I can figure out how to split the commits, but I'm not sure it will be feasible now...

Also true about the history. If done right (which this wasn't), git can track movement of files so you can follow the history.

If you want, I can move the files from `repoquery/*.feature` back to `repoquery-*.feature`, the structure is mostly similar, so a lot of the tests wont move. I did move some tests from file to file and I think that is an improvement that is worth it. The goal is to have test cases reasonably grouped into feature files that make up logical units with their own separate test data. Separating the tests to make them independent and maintaining reasonably sized test data is a balance that is sometimes not easy to maintain.

dnf-behave-tests/features/repoquery/deps.feature
```diff
@@ -0,0 +1,559 @@
...
+
+Background:
+ Given I use the repository "repoquery-deps"
+
+
+# --requires
+Scenario: repoquery --requires NAME
+ When I execute dnf with args "repoquery --requires middle1"
+ Then the exit code is 0
+  And stdout is
```
j-mracek: Please could you add a description into steps. That it handles Last metadata check or downloading repositories? But it is not relevant to repoquery because it is redirected to stderror.

dnf-behave-tests/features/repoquery/deps.feature
```diff
@@ -0,0 +1,559 @@
...
+ When I execute dnf with args "repoquery --conflicts bottom2-1.0"
+ Then the exit code is 0
+  And stdout is
+      """
+      bottom1
+      bottom3 = 1:1.0-1
+      """
+
+Scenario: repoquery --conflicts NAMEGLOB-VERSIONGLOB
+ When I execute dnf with args "repoquery --conflicts bottom*-?.0"
```
j-mracek: What about to replace `--conflicts bottom*-?.0` by `--conflicts 'bottom*-?.0'`? bash could replace `*` by something else.

dnf-behave-tests/features/repoquery/deps.feature
```diff
@@ -0,0 +1,559 @@
...
+ When I execute dnf with args "repoquery --obsoletes bottom3-1.0"
+ Then the exit code is 0
+  And stdout is
+      """
+      bottom1 < 7
+      bottom2
+      """
+
+Scenario: repoquery --obsoletes NAMEGLOB-VERSION
+ When I execute dnf with args "repoquery --obsoletes bottom[3]-1.0"
```
j-mracek: If I replace argument `bottom[3]-1.0` `by bottom*-1.0`, will be there any difference in the output?

dnf-behave-tests/features/repoquery/deps.feature
```diff
@@ -0,0 +1,559 @@
...
+ When I execute dnf with args "repoquery --whatrequires bottom1"
+ Then the exit code is 0
+  And stdout is
+      """
+      middle1-1:1.0-1.x86_64
+      middle1-1:2.0-1.x86_64
+      middle2-1:2.0-1.x86_64
+      """
+
+Scenario: repoquery --whatrequires NAME-VERSION
```
j-mracek: What about to add test for `--whatrequires NEVRA`?

dnf-behave-tests/features/repoquery/globs.feature
```diff
@@ -0,0 +1,178 @@
+Feature: Glob tests for expanding all the various glob patterns.
+
+Background:
+ Given I use the repository "repoquery-globs"
+
+
+# <name> globs
+Scenario: repoquery '*'
```
j-mracek: Please could you add something like - reports all available packages

j-mracek: `repoquery '*'` == `repoquery` == `repoquery --available` == `repoquery --all`

dnf-behave-tests/features/repoquery/globs.feature
```diff
@@ -0,0 +1,178 @@
...
+      toped-1:1.0-1.x86_64
+      topgit-1:1.0-1.src
+      topgit-1:1.0-1.x86_64
+      topgit-1:1.17.6-1.src
+      topgit-1:1.17.6-1.x86_64
+      toppler-1:1.0-1.src
+      toppler-1:1.0-1.x86_64
+      """
+
+Scenario: repoquery top?d
```
j-mracek: What about to add packages `topad` and `topeed` into repo to make scenario more complex?

dnf-behave-tests/features/repoquery/globs.feature
```diff
@@ -0,0 +1,178 @@
...
+ When I execute dnf with args "repoquery top?d"
+ Then the exit code is 0
+  And stdout is
+      """
+      toped-1:1.0-1.src
+      toped-1:1.0-1.x86_64
+      """
+
+Scenario: repoquery top?{d,it}
+ When I execute dnf with args "repoquery top?{{d,it}}"
```
j-mracek: Why there are two {{?

dnf-behave-tests/features/repoquery/main.feature
```diff
@@ -0,0 +1,521 @@
...
+      mid-a2-1:1.0-1.x86_64
+      top-a-1:1.0-1.src
+      top-a-1:1.0-1.x86_64
+      top-a-1:2.0-1.src
+      top-a-1:2.0-1.x86_64
+      top-a-2:2.0-2.src
+      top-a-2:2.0-2.x86_64
+      """
+
+Scenario: dnf repoquery --all NAME
```
j-mracek: I believe that this behavior is illogical. But lets keep it.

dnf-behave-tests/features/repoquery/main.feature
```diff
@@ -0,0 +1,521 @@
...
+      """
+
+Scenario: repoquery --arch ARCH (nonexisting arch)
+ When I execute dnf with args "repoquery --arch yesarch"
+ Then the exit code is 0
+  And stdout is empty
+
+
+# --cacheonly
+Scenario: repoquery -C (without any cache)
```
j-mracek: What about to add command `dnf clean all`?

dnf-behave-tests/features/repoquery/main.feature
```diff
@@ -0,0 +1,521 @@
...
+Scenario: repoquery -C (with cache, but disabled repository)
+Given I successfully execute dnf with args "makecache"
+Given I disable the repository "repoquery-main"
+ When I execute dnf with args "repoquery --available -C"
+ Then the exit code is 0
+  And stdout is empty
+
+
+# --deplist
+Scenario: repoquery --deplist NAME
```
j-mracek: There is a deplist feature file. What about to move it there. Also deplist is a command as an alias repoquery --deplist. Is there any test for it?

j-mracek: Also valid for scenarios bellow.

dnf-behave-tests/features/repoquery/main.feature
```diff
@@ -0,0 +1,521 @@
...
+
+Scenario: repoquery --installed NAME (no such packages)
+ When I execute dnf with args "repoquery --installed bottom-a1"
+ Then the exit code is 0
+  And stdout is empty
+
+
+# --location
+@bz1639827
+Scenario: repoquery --location NAME
```
j-mracek: With http repo the test will have an additional value (using baseurl or even better with metalink).

dnf-behave-tests/features/repoquery/rich-deps.feature
```diff
@@ -0,0 +1,187 @@
...
+ Then the exit code is 0
+  And stdout is
+      """
+      c1-0:1.0-1.x86_64
+      c1-0:2.0-1.x86_64
+      """
+
+
+# a1-1.0: Conflicts: ((b1 and x1) or c1)
+Scenario: repoquery --whatconflicts for "((b1 and x1) or c1)"
```
j-mracek: Please specify type of arguments - provide or package name?

dnf-behave-tests/features/repoquery/weak-deps.feature
```diff
@@ -0,0 +1,402 @@
+Feature: Tests for repoquery weak deps related functionality:
+ --recommends, --supplements, --suggests, --enhances,
+ --whatrecommends, --whatsupplements, --whatsuggests, --whatenhances
+ --whatdepends
+
+# all the packages a1, a2, b1, b2, c1, c2, d1, d2 and top follow
```
j-mracek: What about to rename package a1 to pkgA1? They could provide provideA1. Or change it to some real name or describe in the name the purpose of the package?

dnf-behave-tests/features/repoquery/deps.feature
lukash: Unrelated to this PR, I'll send another one with that. I've got a branch with some more improvements to it, I'll post it after I give it a bit more testing.

dnf-behave-tests/features/repoquery/deps.feature
lukash: Depends on whether more packages would match, but we do have test with `*` with e.g. the `--conflicts` switch, so having this one be different is better I think.

dnf-behave-tests/features/repoquery/main.feature
lukash: There's a rest for the alias in `features/repoquery/main.feature` and no other tests in other files, I think it's fine.

dnf-behave-tests/features/repoquery/weak-deps.feature
lukash: @m-blaha, @pkratoch what do you think of the names of packages in this file (and possibly the other ones) and do we have any conventions I should follow?

I simply have a bunch of packages here for which I want to test some non-trivial rich dependencies, so I'm not sure anything more fancy makes sense.

IMO adding a prefix like `pkgA1` doesn't add value and just makes lines longer. The provides are always named like `PKGNAME-provX`, so they are differentiated.

dnf-behave-tests/features/repoquery/rich-deps.feature
lukash: Every package provides its name and I assume these commands work with provides, so they are provides, in this case the implicit ones. I don't think there's anything unclear here, but if you do think so, let me know.

dnf-behave-tests/features/repoquery/globs.feature
lukash: I know what you mean, but after thinking about it, I don't think it's worth it to add the packages. The goal of the tests is not to test `?` properly matches multiple characters, just that it works. Many more glob tests match multiple packages, so I think it's fine like this. But let me know if you think otherwise.

dnf-behave-tests/features/repoquery/globs.feature
lukash: It's because `format()` is called on the argument in the steps. The way to escape `{}` for `format()` is to double them.

dnf-behave-tests/features/repoquery/main.feature
lukash: The cache is clean at the start of the scenario. I could add it to "make sure", but it is of no effect at this state and if someone ever changes the tests so that there is a cache at the start, some of the tests will actually start failing. So I don't think such change could go unnoticed and invalidate the tests.

Therefore it is perhaps even better to have tests that rely on the cache being clean at the start, since if anyone messed with this globally, there is a possibility they could be invalidating random other tests, which is not a good thing...

lukash:
(sorry about the intermediate noise)

All comments I didn't reply to were addressed in the commits.

I've also put back the queryformat tests commit, as the PR it depended on is merged now.

@j-mracek there is still the issue with glob support for `--whatconflicts` and `--whatobsoletes`, see tests marked with a `TODO` and an `@xfail` in `deps.feature`.

dnf-behave-tests/features/repoquery/weak-deps.feature
pkratoch: I don't think that "pkg" prefix adds any value, but differentiating between ``a``, ``b``, ``c`` and ``d`` would be nice if possible, because otherwise the reader must remember what's the difference between them. I'm thinking of something like ``recommender``, ``supplementer``, ``suggester`` and ``enhancer``, because the packages in the ``repoquery-weak-deps`` repo are this simple (but it's just a suggestion, the schema was helpful enough once I read the specs).

Btw, you are mentioning rich dependencies in your comment, but these are normal ones (rich are the boolean ones, right?). Is this a mistake? Or are you using packages from the ``repoquery-weak-deps`` repo in some other tests, too?

pkratoch: As I said, with the schema, the tests were readable enough for me.

> I'm thinking of something like `recommender`, `supplementer`, `suggester` and `enhancer`.

I am not super happy about it either, in other tests it works better (in `dnf-ci-obsoletes`, there is `PackageA` and `PackageA-Obsoleter` and it's clear what are the roles), but here both sides of the relations are important and only one is described by the name, so it's not that great.

dnf-behave-tests/features/repoquery/deps.feature
lukash: The PR is here: https://github.com/rpm-software-management/ci-dnf-stack/pull/628

PR #612, opened 2019-08-29
 Add test to check dnf command alias in the help output

pkratoch:
Some packages needed to be added to the dockerfiles in order to have
the yum/yum4 command available.

DNF base command aliases on various systems:
 - Fedora <= 30: "yum" from package "dnf-yum"
 - Fedora >= 31: "yum" from package "yum"
 - RHEL 8: "yum" from package "yum"
 - RHEL 7 (EPEL): "yum4" from package "nextgen-yum4"

Additionally, on all of these systems, there is package "dnf" providing
command "dnf" and "dnf-2" or "dnf-3" based on python version ("dnf-2" in
RHEL7, "dnf-3" otherwise)

### DIFF
--- a/Dockerfile.f31
+++ b/Dockerfile.f31
@@ -1,0 +1,110 @@
+# Build
+# -----
+# $ podman build --build-arg TYPE=local -t dnf-bot/ci-dnf-stack:f31 -f Dockerfile.f31
+#
+#
+# Run
+# ---
+# $ podman run -it dnf-bot/ci-dnf-stack:f31 behave -Ddnf_executable=dnf --junit --junit-directory=/opt/behave/junit/ [--wip --no-skipped]
+#
+#
+# Build types
+# -----------
+# distro
+#       install distro packages
+# copr
+#       install distro packages
+#       then upgrade to copr packages
+# local
+#       install distro packages
+#       then upgrade to copr packages
+#       then install packages from local rpms/ folder
+#       install also additional tools for debugging in the container
+
+
+FROM fedora:31
+ENV LANG C
+ARG TYPE=local
+ARG OSVERSION=fedora__31
+
+
+
+# disable deltas and weak deps
+RUN set -x && \
+    echo -e "deltarpm=0" >> /etc/dnf/dnf.conf && \
+    echo -e "install_weak_deps=0" >> /etc/dnf/dnf.conf
+
+
+# if TYPE == copr or local, enable nightly copr
+RUN set -x && \
+    if [ "$TYPE" == "copr" -o "$TYPE" == "local" ]; then \
+        dnf -y install dnf-plugins-core; \
+        dnf -y copr enable rpmsoftwaremanagement/dnf-nightly; \
+    fi
+
+
+# upgrade all packages to the latest available versions
+RUN set -x && \
+    dnf -y --refresh upgrade
+
+
+# install the test environment and additional packages
+RUN set -x && \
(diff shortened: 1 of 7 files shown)


pkratoch:
The tag centos__eq__7 doesn't work for some reason despite commit https://github.com/rpm-software-management/ci-dnf-stack/commit/d30622545cc307789e5aa8b8a62d5253150c8b27#diff-fb5dc39ddb7e42f372ef4f2513af02e8

And btw, nextgen-yum4 is currently broken in the dnf-nightly repo (missing commit https://github.com/rpm-software-management/dnf/commit/660eec1969fc0679d7febfb3a2cb26f9cf7678ec ), but it can be built using the TYPE=distro

pkratoch:
> The tag centos__eq__7 doesn't work for some reason despite commit [d306225#diff-fb5dc39ddb7e42f372ef4f2513af02e8](https://github.com/rpm-software-management/ci-dnf-stack/commit/d30622545cc307789e5aa8b8a62d5253150c8b27#diff-fb5dc39ddb7e42f372ef4f2513af02e8)

The problem was (and still is) that tags are evaluated with logical OR, meaning that when there are e.g. tags ``@not.with_os=rhel__eq__7`` and ``@not.with_os=centos__eq__7``, the scenario is not skipped because one of the tags evaluates to true.

The tags used in this test should be working now. I replaced ``@not.with_os`` with complementary ``@use.with_os``.

Dockerfile.f31
```diff
@@ -0,0 +1,111 @@
...
+os=$OSVERSION" > /opt/behave/behave.ini
+
+# build test repos from sources
+RUN set -x && \
+    cd /opt/behave/fixtures/specs/ && \
+    ./build.sh --force-rebuild
+
+
+VOLUME ["/opt/behave/junit"]
+WORKDIR /opt/behave
```
lukash: Could the new Dockerfile go into a separate commit?

dnf-behave-tests/features/help.feature
```diff
@@ -26,3 +26,51 @@ Scenario: Command help
...
+    And stdout contains "usage: <command>"
+
+Examples:
+        | args      | command  |
+        | help      | dnf      |
+        | --help    | dnf      |
+        | help      | dnf-3    |
+        | --help    | dnf-3    |
+        | help      | yum      |
+        | --help    | yum      |
```
lukash: Maybe flip the two columns for a bit better readability?

Dockerfile.f31
pkratoch: Done.

dnf-behave-tests/features/help.feature
pkratoch: Done.

dnf-behave-tests/features/steps/cmd.py
```diff
@@ -7,7 +7,12 @@
 import re
 import sys
 from datetime import datetime
-from itertools import zip_longest
+
+PY3 = sys.version_info.major >= 3
+if PY3:
+    from itertools import zip_longest
+else:
+    from itertools import izip_longest as zip_longest
```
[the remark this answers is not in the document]
pkratoch: I don't see much difference between the two and the same is used in ``steps.fixtures.httpd``, so I'd rather stay consistent.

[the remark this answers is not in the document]
lukash: Okay.

PR #616, opened 2019-08-30
Add tests for pkg running scriptlets (BhBug:1724779)

kontura:


dnf-behave-tests/features/scriptlets.feature
```diff
@@ -92,3 +92,106 @@ Scenario: Remove a pkg with a failing %postun scriptlet
...
+       | Action        | Package                             |
+       | install       | Package-triggerin-ok-0:1.0-1.x86_64 |
+  When I execute dnf with args "install Package-install-file"
+  Then the exit code is 0
+   And Transaction is following
+       | Action        | Package                             |
+       | install       | Package-install-file-0:1.0-1.x86_64 |
+   And stdout contains "triggerin scriptlet \(Package-triggerin-ok\) for Package-install-file install/update successfully done"
+   And stdout does not contain "Running scriptlet\s*:\s*Package-install-file"
+   And stdout contains "Running scriptlet\s*:\s*Package-triggerin-ok"
```
lukash: Unless there's something nondeterministic in the output, could you use `stdout is`/`stderr is` and an exact match?

dnf-behave-tests/features/scriptlets.feature
```diff
@@ -92,3 +92,106 @@ Scenario: Remove a pkg with a failing %postun scriptlet
    And Transaction is following
        | Action        | Package                           |
        | remove        | Package-postun-fail-0:1.0-1.x86_64 |
+
+
+Scenario: Correct output for triggered successful scriptlet of package not present in transaction
+  When I execute dnf with args "install Package-triggerin-ok"
```
lukash: I think this is not part of the real test, it's the setup, so you can use:
```
Given I successfully execute dnf with args "install Package-triggerin-ok"
```

And then you don't need to check the exit status...

dnf-behave-tests/features/scriptlets.feature
kontura: I probably could but are you sure it's worth it? It would be a lot of unnecessary text in the scenario since I would have to include the whole transaction table and so on. I like that this approach makes it clear what we are checking.

lukash: Oh, right, there's the transaction as well. Yeah, leave it like this, you'd have to repeat the transaction content and that is counter-productive...

I've got some ideas how we could solve this, but they'll require some time to make reality... :slightly_smiling_face:

kontura:
Setting this as blocked at least until https://github.com/rpm-software-management/dnf/pull/1471 gets merged.

I have updated this PR , there are a few tests marked as failing because of RPM, I described the situation more in comments or even more detailed description is in the https://bugzilla.redhat.com/show_bug.cgi?id=1724779.

kontura:
Additionally tweak some comments.

dnf-behave-tests/features/scriptlets.feature
```diff
@@ -92,3 +92,113 @@ Scenario: Remove a pkg with a failing %postun scriptlet
...
+   And Transaction is following
+       | Action        | Package                             |
+       | install       | Package-install-file-0:1.0-1.x86_64 |
+   And stdout contains "triggerin scriptlet \(Package-triggerin-ok\) for Package-install-file install/update successfully done"
+   And stdout does not contain "Running scriptlet\s*:\s*Package-install-file"
+   And stdout contains "Running scriptlet\s*:\s*Package-triggerin-ok"
+   # We really should be matching the whole NEVRA of pkg but current RPM Python API provides 
+   # only the name (we can fix this by moving logic to libdnf, DNF5).
+   # This same situation is in all of the tests below.
+   # And stdout contains "Running scriptlet\s*:\s*Package-triggerin-ok-1.0-1.x86_64"
```
lukash: I'm not sure how much it is a personal preference, but I'm used to comments going before the line to which they belong. Maybe just a nitpick, but seems to me it would actually be a bit clearer if they went before the last line of the scenario, which is what they are referencing I think (goes for all the comments at ends of scenarios in the file).

dnf-behave-tests/fixtures/specs/dnf-ci-scriptlets/Package-filetriggerin-fail.spec
```diff
@@ -0,0 +1,19 @@
+Name:           Package-filetriggerin-fail
```
lukash: Should the repository be named just `scriptlets` to go with the convention we have?

dnf-behave-tests/features/scriptlets.feature
kontura: Thank you. I think the comment line is indeed usually above. I moved them.

PR #618, opened 2019-09-04
added scenario for listing non-existent module

glum23:
added scenario for listing non-existent module, related to bug 1647382

### DIFF
--- a/dnf-behave-tests/features/module-list.feature
+++ b/dnf-behave-tests/features/module-list.feature
@@ -91,6 +91,11 @@ Scenario: I can list disabled modules (when there are no disabled modules)
     And stderr is empty
     And module list is empty
 
+@bz1647382
+Scenario: Get error message when list of non-existent module is requested
+   When I execute dnf with args "module list non-existing-module"
+   Then the exit code is 1
+    And stderr contains "Error: No matching Modules to list"
 
 Scenario: I can list disabled modules
    When I execute dnf with args "module disable postgresql"


dnf-behave-tests/features/module-list.feature
```diff
@@ -91,6 +91,11 @@ Scenario: I can list disabled modules (when there are no disabled modules)
     And stderr is empty
     And module list is empty
 
+@bz1647382
+Scenario: Get error message when list of non-existent module is requested
+   When I execute dnf with args "module list non-existing-module"
+   Then the exit code is 1
+    And stderr contains "Error: No matching Modules to list"
```
lukash: Is it possible to match for an exact error with `And stderr is ...`?

glum23: it is copied from related scenario (module info) so that scenarios are consistent

glum23: ... and I'm not a fan of exact matching, messages can be updated, extended, ...

lukash: > ... and I'm not a fan of exact matching, messages can be updated, extended, ...

But that is one of the actual points of exact matching. If the output changes, you need to update the test. That way you can also rely on the test catching any inadvertent change you make. I would like to stress how important this is. I want to rely on the tests catching any side-effects of a change I am making, but with a "contains" check, I can't do it. For example, the same error could be logged twice, the string could be prefixed or suffixed with something, or there could be two thousand lines of gibberish outputted here and the tests will pass...

If the output isn't nondeterministic (or perhaps too long), I don't think there's an argument for not doing an exact match.

And also, I've recently enhanced the `std{out,err} is ...` check to print a full colored diff in case of a mismatch, so that you can much more easily see what is different.

PR #621, opened 2019-09-05
Plugin has access to installed pkgs in transaction hook RhBug:1650446

kontura:


dnf-behave-tests/features/plugin-api.feature
```diff
@@ -0,0 +1,35 @@
+Feature: Plugin API
+
+
+Background:
+Given I do not disable plugins
+Given I do not set config file
+  And I create and substitute file "/etc/dnf/dnf.conf" with
```
lukash: A nitpick (or, a rant, really, why isn't there just one syntax to do it):
```
Given a
Given b
Given c
```
or
```
Given a
  And b
  And c
```

dnf-behave-tests/features/plugin-api.feature
```diff
@@ -0,0 +1,35 @@
...
+        downloads = self.base.transaction.install_set
+        for pkg in downloads:
+            print("Plugin has access to installed pkg: " +
+                   pkg.name + "+" + pkg.version + "+" + pkg.release + "+" + pkg.arch)
+
+  """
+  And I use the repository "dnf-ci-fedora"
+ When I execute dnf with args "install setup filesystem"
+ Then stdout contains "Plugin has access to installed pkg: setup\+2.12.1\+1.fc29\+noarch"
+ Then stdout contains "Plugin has access to installed pkg: filesystem\+3.9\+2.fc29\+x86_64"
```
lukash: Sorry about being so insistent... but exact match should be no problem here?

kontura: :-) I think its actually the same scenario, the plugin transaction hook triggers right after the transaction and therefore it prints the messages in the middle of the standard dnf install output.

dnf-behave-tests/features/plugin-api.feature
kontura: updated, thanks

dnf-behave-tests/features/plugin-api.feature
lukash: Oh, gosh darn it :slightly_smiling_face:

PR #625, opened 2019-09-11
Test that upgrade does not take build times into account

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1728252


lukash:
Nice test :slightly_smiling_face:

PR #633, opened 2019-09-19
Download tests are run in temp directory

m-blaha:
Download tests used to download files into directory from where the behave was run.
This PR adds step to change working directory and use it in download tests. Also step for running dnf in repository directory was replaced by setting working directory.

### DIFF
--- a/dnf-behave-tests/features/environment.py
+++ b/dnf-behave-tests/features/environment.py
@@ -115,6 +115,7 @@ class DNFContext(object):
         self.disable_plugins = True
         self.disable_repos_option = "--disablerepo='*'"
         self.assumeyes_option = "-y"
+        self.working_dir = None
 
         # temporarily use DNF0 for substituting fixturesdir in repo files
         # the future could be in named environment variable like DNF_VAR_FIXTURES_DIR
--- a/dnf-behave-tests/features/install-path.feature
+++ b/dnf-behave-tests/features/install-path.feature
@@ -2,7 +2,8 @@ Feature: Tests for installing RPM from paths
 
 
 Scenario Outline: I can install an RPM from path, where path is <path type>
-   When I execute dnf with args "install <path>" from repo "dnf-ci-fedora"
+  Given I set working directory to "{context.dnf.fixturesdir}/repos/dnf-ci-fedora"
+   When I execute dnf with args "install <path>"
    Then the exit code is 0
     And Transaction is following
         | Action        | Package                                   |
@@ -17,7 +18,7 @@ Examples:
 
 
 Scenario: I can install an RPM from path, when specifying the RPM multiple times
-   When I execute dnf with args "install noarch/setup-2.12.1-1.fc29.noarch.rpm noarch/setup-2.12.1-1.fc29.noarch.rpm" from repo "dnf-ci-fedora"
+   When I execute dnf with args "install {context.dnf.fixturesdir}/repos/dnf-ci-fedora/noarch/setup-2.12.1-1.fc29.noarch.rpm {context.dnf.fixturesdir}/repos/dnf-ci-fedora/noarch/setup-2.12.1-1.fc29.noarch.rpm"
    Then the exit code is 0
     And Transaction is following
         | Action        | Package                                   |
@@ -26,7 +27,8 @@ Scenario: I can install an RPM from path, when specifying the RPM multiple times
 
 @xfail
 Scenario: I can install an RPM from path, when specifying the RPM multiple times using different paths
-   When I execute dnf with args "install noarch/setup-2.12.1-1.fc29.noarch.rpm {context.dnf.fixturesdir}/repos/dnf-ci-fedora/noarch/setup-2.12.1-1.fc29.noarch.rpm" from repo "dnf-ci-fedora"
+  Given I set working directory to "{context.dnf.fixturesdir}/repos/dnf-ci-fedora"
+   When I execute dnf with args "install noarch/setup-2.12.1-1.fc29.noarch.rpm {context.dnf.fixturesdir}/repos/dnf-ci-fedora/noarch/setup-2.12.1-1.fc29.noarch.rpm"
    Then the exit code is 0
     And Transaction is following
         | Action        | Package                                   |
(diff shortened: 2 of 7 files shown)


dnf-behave-tests/features/steps/cmd.py
```diff
@@ -40,10 +40,22 @@ def handle_reposync(expected, found):
     return expected, found
 
 
+def run_with_args(context, cmd, **run_args):
```
lukash: `run_with_args` actually isn't a good name, what are the args? The args are arguments to the `run` function, not the arguments of the command itself (which is what this formulation is otherwise used for in the CI stack context).

`run_in_cwd`, `run_in_work_dir` or perhaps just `run_in_context` as there is the faketime thing as well as the directory?

dnf-behave-tests/features/environment.py
```diff
@@ -106,6 +106,7 @@ def __init__(self, userdata, force_installroot=False):
         self.disable_plugins = True
         self.disable_repos_option = "--disablerepo='*'"
         self.assumeyes_option = "-y"
+        self.cwd = None
```
lukash: Not sure `cwd` is actually a convention to call the working directory, I'd prefer to call it e.g. `work_dir`, I myself wouldn't know what `cwd` is.

m-blaha:
Sorry for that conflict. I was just annoyed with all these *.rpm files created in tests directory after running them...
Anyway, I can wait with this change until your PR gets merged.

m-blaha:
I tried doing os.chdir() in before_scenario() but found out that "Scenario Outline" started to traceback badly inside the behave.
I have renamed cwd and run_with_args according your suggestion. Original names were really unclear.

dnf-behave-tests/features/upgrade-from-path.feature
```diff
@@ -49,7 +51,8 @@ Scenario: Upgrade an RPM from path on disk containing wildcards
 
 
 Scenario: Upgrade an RPM from path on disk, when specifying the RPM multiple times
-   When I execute dnf with args "upgrade x86_64/wget-1.19.6-5.fc29.x86_64.rpm x86_64/wget-1.19.6-5.fc29.x86_64.rpm" from repo "dnf-ci-fedora-updates"
+  Given I set working directory to "{context.dnf.fixturesdir}/repos/dnf-ci-fedora-updates"
+   When I execute dnf with args "upgrade x86_64/wget-1.19.6-5.fc29.x86_64.rpm x86_64/wget-1.19.6-5.fc29.x86_64.rpm"
```
lukash: One more thing, in these cases here the full path could be passed to the `upgrade` command, right? There is no need to somewhat arbitrarily split it and set the working directory? (except the lines end up being shorter this way, perhaps)

I think I'd prefer the full path...

dnf-behave-tests/features/upgrade-from-path.feature
```diff
@@ -28,7 +28,8 @@ Scenario: Upgrade an RPM from absolute path on disk
 
 
 Scenario: Upgrade an RPM from relative path on disk
-   When I execute dnf with args "upgrade x86_64/glibc-2.28-26.fc29.x86_64.rpm x86_64/glibc-common-2.28-26.fc29.x86_64.rpm x86_64/glibc-all-langpacks-2.28-26.fc29.x86_64.rpm" from repo "dnf-ci-fedora-updates"
+  Given I set working directory to "{context.dnf.fixturesdir}/repos/dnf-ci-fedora-updates"
+   When I execute dnf with args "upgrade x86_64/glibc-2.28-26.fc29.x86_64.rpm x86_64/glibc-common-2.28-26.fc29.x86_64.rpm x86_64/glibc-all-langpacks-2.28-26.fc29.x86_64.rpm"
```
lukash: ... and another one here.

dnf-behave-tests/features/upgrade-from-path.feature
```diff
@@ -49,7 +50,7 @@ Scenario: Upgrade an RPM from path on disk containing wildcards
 
 
 Scenario: Upgrade an RPM from path on disk, when specifying the RPM multiple times
-   When I execute dnf with args "upgrade x86_64/wget-1.19.6-5.fc29.x86_64.rpm x86_64/wget-1.19.6-5.fc29.x86_64.rpm" from repo "dnf-ci-fedora-updates"
+        When I execute dnf with args "upgrade {context.dnf.fixturesdir}/repos/dnf-ci-fedora-updates/x86_64/wget-1.19.6-5.fc29.x86_64.rpm {context.dnf.fixturesdir}/repos/dnf-ci-fedora-updates/x86_64/wget-1.19.6-5.fc29.x86_64.rpm"
```
lukash: A superfluous indent here.

dnf-behave-tests/features/upgrade-from-path.feature
m-blaha: Here is the upgrade from the relative path tested, so the working dir change is required.

dnf-behave-tests/features/install-path.feature
```diff
@@ -26,7 +27,8 @@ Scenario: I can install an RPM from path, when specifying the RPM multiple times
 
 @xfail
 Scenario: I can install an RPM from path, when specifying the RPM multiple times using different paths
-   When I execute dnf with args "install noarch/setup-2.12.1-1.fc29.noarch.rpm {context.dnf.fixturesdir}/repos/dnf-ci-fedora/noarch/setup-2.12.1-1.fc29.noarch.rpm" from repo "dnf-ci-fedora"
+  Given I set working directory to "{context.dnf.fixturesdir}/repos/dnf-ci-fedora"
+   When I execute dnf with args "install noarch/setup-2.12.1-1.fc29.noarch.rpm {context.dnf.fixturesdir}/repos/dnf-ci-fedora/noarch/setup-2.12.1-1.fc29.noarch.rpm"
```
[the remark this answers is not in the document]
m-blaha: Here the same file using two different paths is used, so one of the paths is relative and the other absolute. So the working dir change is required here.

dnf-behave-tests/features/upgrade-from-path.feature
m-blaha: Sorry for this, fixed.

PR #634, opened 2019-09-20
Fix crashing get_rpmdb_rpms() on Fedora 31.

dmach:


dmach:
Fixed in 96b3fbc25f495badd200a28734bc16d6063b3326 already -> closing

PR #635, opened 2019-09-20
Fix gpg tests for Fedora 31

m-blaha:
Rpm has changed format of gpg package summary between 4.14 (F30) and
4.15 (F31) so the command "rpm -q gpg-pubkey --qf '%{summary}\n'"
provides different outputs on different Fedoras:
4.14: gpg(dnf-ci-gpg)
4.15: dnf-ci-gpg public key

This patch uses query format that mimics the format of summary from rpm
4.14.

Relevant rpm patch: https://github.com/rpm-software-management/rpm/commit/c7f284004addb0b85c5c9c1a02da08c1ad88ac91


dmach:
I've tested this on f31 and f30 and it works fine.

PR #636, opened 2019-09-23
Test 'dnf remove --duplicates' (RhBug:1674296)

dmach:


j-mracek:
@dmach Tests look good. The perfect approach would be to install packages with files with different content, but it requires real crash during transaction. Please feel free to merge both pull-request.

PR #638, opened 2019-09-27
Fix CI stack after default best=False

m-blaha:


dnf-behave-tests/features/config-best.feature
```diff
@@ -49,9 +49,15 @@ Scenario: When installing with option --nobest, install a package from repo with
 
 
 @bz1670776 @bz1671683
-Scenario: When installing with best=1 (default), fail on broken packages, and advise to use --nobest
+Scenario: When installing with best=1 set in dnf.conf, fail on broken packages, and advise to use --nobest
   Given I use the repository "dnf-ci-fedora"
-  Given I use the repository "dnf-ci-fedora-updates"
+    And I use the repository "dnf-ci-fedora-updates"
+    And I do not set config file
+    And I create file "/etc/dnf/dnf.conf" with
```
j-mracek: What about to replace the conf by:
```
gpgcheck=1
installonly_limit=3
clean_requirements_on_remove=True
best=True
```
It is identical to our conf that we ship.

dnf-behave-tests/features/install-installed.feature
```diff
@@ -24,8 +24,14 @@ Scenario: Install installed RPM when upgrade is available
 
 
 @bz1670776 @bz1671683
-Scenario: Install installed RPM when upgrade is available with best=True (default)
+Scenario: Install installed RPM when upgrade is available with best=True (in dnf.conf)
   Given I use the repository "dnf-ci-fedora"
+    And I do not set config file
+    And I create file "/etc/dnf/dnf.conf" with
+    """
+    [main]
```
j-mracek: What about to use full dnf.conf like we ship
```
gpgcheck=1
installonly_limit=3
clean_requirements_on_remove=True
best=True
```

dnf-behave-tests/features/install-installed.feature
```diff
@@ -46,6 +52,31 @@ Scenario: Install installed RPM when upgrade is available with best=True (defaul
...
 
 
+@bz1670776 @bz1671683
+Scenario: Install installed RPM when upgrade is available with best=False (in dnf.conf)
+  Given I use the repository "dnf-ci-fedora"
+    And I do not set config file
+    And I create file "/etc/dnf/dnf.conf" with
+    """
+    [main]
+    best=True
```
j-mracek: best=False !
or alternatively:
```
gpgcheck=1
installonly_limit=3
clean_requirements_on_remove=True
best=False
```

dnf-behave-tests/features/steps/common/rpm.py
```diff
@@ -100,12 +100,20 @@ def diff_rpm_lists(list_one, list_two):
...
 
     list_one = sorted(list_one)
     list_two = sorted(list_two)
 
+    # PRESENT PACKAGES AND UNCHANGED INSTALLONLY PACKAGES
+    for pkg in list_two:
+        result["present"].append(pkg)
+        if (pkg in list_one) and pkg.is_installonly():
+            result["unchanged"].append(pkg)
+    for pkg in result["unchanged"]:
```
j-mracek: Reinstall == unchanged?

dnf-behave-tests/features/install-installed.feature
m-blaha: No, this tests distribution with other defaults (like RHEL).

dnf-behave-tests/features/steps/common/rpm.py
m-blaha: I'm not sure how to distinguish these two states. When comparing two versions of rpm database, both look the same. We probably should not bother with unchanged/reinstall on rpmdb level and consider them one state. This definitely deserves more work.

dnf-behave-tests/features/config-best.feature
m-blaha: I can do that, but other options are irrelevant for the test.

lukash: I'd suggest that we (eventually, not necessarily in this PR) create the config file from the steps in the default location (`/etc/dnf/dnf.conf`). Then we can add a step like this:
```
Given I set in config file
      | key  | value |
      | best | True  |
```

Which would use the defaults and add this to it.

If we go that route, it seems preferable to add the full config to these scenarios now, as @j-mracek suggests, because then whoever implements what I just described will not have to think if the other settings in the default config have any impact on the scenarios.

m-blaha: Yes, I like it. This has advantages of both - only relevant settings in the feature and still full config used for execution. The only problem is what the defaults are? The Fedora ones? We are about to start delivering different dnf.conf files for different distributions.

lukash: > The only problem is what the defaults are? The Fedora ones? We are about to start delivering different dnf.conf files for different distributions.

Right now the defaults are the contents of `dnf-behave-tests/features/dnf.conf`, so in context of the suggested change we'd go with that and effectively make no changes to the configuration... :slightly_smiling_face: 

As to what put in the defaults, not sure, we might need to discuss that? But with generating the config in runtime we have much more control and can more easily test multiple configuration combinations. The current defaults are options that don't change per distribution...

PR #640, opened 2019-09-27
Better counting v2

dmnks:
Depends on:
1) librepo [PR#171](https://github.com/rpm-software-management/librepo/pull/171)
2) libdnf [PR#807](https://github.com/rpm-software-management/libdnf/pull/807)

### DIFF
--- a/dnf-behave-tests/features/steps/repo.py
+++ b/dnf-behave-tests/features/steps/repo.py
@@ -293,6 +293,11 @@ def step_impl(context, client_cert, client_key):
                                                     client_key)
 
 
+@behave.step("I forget any HTTP requests captured so far")
+def step_clear_http_logs(context):
+    context.httpd.clear_log()
+
+
 @behave.step("I am running a system identified as the \"{system}\"")
 def given_system(context, system):
     data = dict(zip(('NAME', 'VERSION_ID', 'VARIANT_ID'), system.split(' ')))
@@ -313,7 +318,6 @@ def step_check_http_log(context, quantifier, command):
     log = context.httpd.get_log()
     assert log is not None, 'Logging should be enabled on the HTTP server'
     log = [rec for rec in log if rec.command == command]
-    context.httpd.clear_log()
     assert log, 'Some HTTP requests should have been received'
 
     # A log dump, printed on failures for convenience
--- a/dnf-behave-tests/features/steps/repo.py
+++ b/dnf-behave-tests/features/steps/repo.py
@@ -273,6 +273,11 @@ def step_set_up_metalink_for_repository(context, repo):
     create_repo_conf(context, repo)
 
 
+@behave.step("the server starts responding with HTTP status code {code}")
+def step_server_down(context, code):
+    context.httpd.configure('status', int(code))
+
+
 @behave.step("I start capturing outbound HTTP requests")
 def step_start_http_capture(context):
     context.httpd.configure('logging', True)
--- a/dnf-behave-tests/features/steps/repo.py
+++ b/dnf-behave-tests/features/steps/repo.py
@@ -309,13 +309,6 @@ def given_no_osrelease(context):
     context.osrelease.delete()
 
 
-@behave.given("I am using {package} of the version X.Y.Z")
-def given_package_version(context, package):
-    rpms = [rpm for rpm in get_rpmdb_rpms() if rpm.name == package]
-    assert len(rpms) == 1, 'There should be exactly one %s installed' % package
-    context.versions = {package: rpms[0].version}
-
-
 @behave.step("{quantifier} HTTP {command} request should match")
 @behave.step("{quantifier} HTTP {command} requests should match")
 def step_check_http_log(context, quantifier, command):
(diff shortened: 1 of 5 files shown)


dnf-behave-tests/features/steps/repo.py
```diff
@@ -107,6 +107,13 @@ def step_impl(context, rtype, repo):
     context.execute_steps('Given I use the repository "{}"'.format(repo_id))
 
 
+@behave.step("the server starts responding with HTTP status code {code}")
+def step_server_down(context, code):
+    path = context.dnf.repos_location
+    response = context.httpd.send_request(path, "STATUS %s" % code)
```
lukash: Hey, this is a funky way to do this :slightly_smiling_face: is it not possible to simply set the configuration to the server directly?

If not, could you use HTTP arguments (the ones after a `?`) instead of putting it in the path?

dmnks: I'd be glad to learn how to do it (setting the configuration directly) :)  The server is a subprocess waiting on client connections and parsing their requests. I don't a simpler way to achieve this, to be honest :)

I do agree it's a bit funky, but to me it really seems to be the easiest solution.

One alternative would be to configure the server to respond with errors before actually spawning it, but then you would need to somehow pass arguments via the `httpd` fixture decorator and I'm not sure how to do that. I settled with this "dynamic" way since it might also be useful to "break" the server in a step dynamically.

dmnks: Also, what do you mean by using HTTP arguments? What's the server considers a "path" here is what goes after `STATUS`.

dmnks: OK, now that I talked with Lukas on IRC, it seems that it really is possible to do this cleanly; I'll just pass the configuration changes via shared memory and adjust the request handler to read from that memory :) Looks easy, I'll give it a try.

dmnks:
Rebased.

dnf-behave-tests/features/steps/repo.py
```diff
@@ -217,6 +217,11 @@ def step_set_up_metalink_for_repository(context, repo):
     create_repo_conf(context, repo)
 
 
+@behave.step("I start capturing outbound HTTP requests")
+def step_start_http_capture(context):
+    context.httpd.configure('logging', True)
```
lukash: The patch looks good, just FTR, is it a performance concern that we don't just log by default and save ourselves the trouble to enable it explicitly?

dnf-behave-tests/features/steps/repo.py
```diff
@@ -298,9 +298,12 @@ def step_check_http_log(context, quantifier, command):
...
         assert all(matches), 'Every request should match the table: ' + dump
-    elif quantifier == 'exactly one':
-        assert len([m for m in matches if m]) == 1, \
-            'Exactly one request should match the table: ' + dump
+    elif quantifier.startswith('exactly '):
+        num = quantifier.split(' ')[1]
+        if num == 'one':
+            num = 1
+        assert len([m for m in matches if m]) == int(num), \
+            'Exactly %s requests should match the table: %s' % (quantifier, dump)
```
lukash: In general I think it is preferable to state what is the error in the message, not what it should have looked like?

dnf-behave-tests/features/steps/repo.py
```diff
@@ -261,6 +261,7 @@ def given_package_version(context, package):
 
 
 @behave.step("{quantifier} HTTP {command} request should match")
+@behave.step("{quantifier} HTTP {command} requests should match")
```
lukash: Is this intentional? Does it still work?

dnf-behave-tests/features/steps/repo.py
dmnks: You're actually right, we could probably just enable it straight away. My concern back then was that the in-memory log could get unnecessarily big if there happens to be a scenario that produces a lot of requests, but now I realize it just doesn't really matter since the logs are always cleared together with the httpd instance.  So yeah, I'll probably just drop this step and enable logging by default.

dnf-behave-tests/features/steps/repo.py
dmnks: Good point, I'll fix that.  For some reason I assumed that "documenting" what the `assert` is supposed to test is better, but it really turns out to be too confusing.  I'll fix that.

dnf-behave-tests/features/steps/repo.py
dmnks: Yes, this is intentional (see the step `Then exactly 48 HTTP GET requests should match:`)

lukash: Okay, but what about the tests that use the singular form? Is it something behave handles transparently? I suppose it has to be so...

dmnks: It indeed does. TBH, I haven't found any mention of this in the behave docs, but it works. I agree it's not the most elegant way, but I haven't found a better one.

dmnks: Oh, I just might have found a better one:
https://stackoverflow.com/a/53614851

Will take a look.

lukash: FWIW I'm fine with the way it is... it might confuse someone and may be hard to grep (when someone say greps for "request should match" to look up the step definition, they won't find it), but then the form with regexes will as well.

As long as it works...

dmnks: OK, let's keep it as it is then.

dnf-behave-tests/features/steps/repo.py
dmnks: I changed my mind - I think we should keep this step so that it's clear when exactly the logging starts. The reason is, there might be other DNF operations that reach out to the repository as part of the scenario setup which we don't want to pollute our log with.

dnf-behave-tests/features/steps/repo.py
dmnks: Fixed in commit 6e4b108

dmnks:
Rebased, please re-review.

dmnks:
Note: We should probably wait for the corresponding libdnf PR to be merged (ETA tomorrow) first before merging this one.

PR #644, opened 2019-10-03
Debug output enhancement and some cleanups

lukash:
Some debugging enhancements that print the last called command and its outputs in a consistent way and with colors, temp directories handling enhancements and some cleanups.

The last commit also shortens the name of the `dnf-ci-fedora-modular-updates` repository across all tests. The name is too long and triggers a bug with dnf output, which had to be workarounded in one scenario. This change is in order to be able to work with the repositories in a uniform way, without a workaround.

### DIFF
--- a/dnf-behave-tests/features/environment.py
+++ b/dnf-behave-tests/features/environment.py
@@ -8,7 +8,7 @@ import os
 import shutil
 import tempfile
 
-from behave import fixture, use_fixture
+from behave import fixture, use_fixture, model
 from behave.tag_matcher import ActiveTagMatcher
 from behave.formatter.ansi_escapes import escapes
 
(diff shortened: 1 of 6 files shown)


pkratoch:
Looks good, just one note: In the commit "steps: Print preserved temporary directories", the commit message contains "we print the directories only if we would indeed be deleting them", but it's the other way around.

pkratoch:
@rh-atomic-bot try

lukash:
> Looks good, just one note: In the commit "steps: Print preserved temporary directories", the commit message contains "we print the directories only if we would indeed be deleting them", but it's the other way around.

Well, the whole paragraph is there to note we are moving around the conditions for deletion because of the messages. And this sentence says "would be deleting them", but we aren't deleting them, and those are exactly the conditions under which we want to print the notice... I think it's logically correct, but may be hard to understand :slightly_smiling_face:

pkratoch:
> Well, the whole paragraph is there to note we are moving around the conditions for deletion because of the messages. And this sentence says "would be deleting them", but we aren't deleting them, and those are exactly the conditions under which we want to print the notice... I think it's logically correct, but may be hard to understand slightly_smiling_face

Ok, it's a bit confusing, but I get what you meant and it makes sense.

The rh-atomic-bot is not in ci-dnf-stack, but I ran the tests locally on Fedora 30 and they are passing, so I'll merge this.

PR #645, opened 2019-10-04
Add test whether disabled makecache --timer expires all repos

kontura:
https://github.com/rpm-software-management/dnf/pull/1493


dnf-behave-tests/features/makecache.feature
```diff
@@ -0,0 +1,13 @@
+Feature: makecache command 
+
+
+@xfail
+@fixture.httpd
+@1745170
```
lukash: Missing a `bz` in the tag... Might wanna reference the bug in the commit message as well?

Also, why mark it `@xfail`, if there's a fix that will be merged along with this? For these, I add the `@not.with_os=rhel__eq__8` tag because the fix won't be on RHEL for some time.

dnf-behave-tests/features/makecache.feature
```diff
@@ -0,0 +1,13 @@
...
+Scenario: disabled makecache --timer does not invalidate cached metadata
+Given I use the http repository based on "dnf-ci-fedora"
+  And I successfully execute dnf with args "makecache"
+ When I execute dnf with args "makecache --timer --setopt=metadata_timer_sync=0"
+  And I execute dnf with args "install setup"
+ # This line matches downloading of metadata (generally first line in transaction output)
+ Then stdout does not contain "http-dnf-ci-fedora\s+.+B\/s \| .+B"
```
lukash: I understand you want to check for the sync line here, as that is what you care about and the rest of the output is irrelevant...

So just a note perhaps (Feel free to leave it as it is), I personally find the regexes hard to read and therefore annoying. While this is a simple case, we do have much longer ones. Sometimes they don't match and you may need to spend a long time trying to figure out what exactly it does and what causes the problem. So the point is, I guess, to try to avoid them and only use them if necessary.

dnf-behave-tests/features/makecache.feature
kontura: oh, yes thank you, I will update that.

About the @xfail I use it because I thought if I just put rhel tag and then the bug fix is released first in for example F30 it will be failing in F31, F32.. until its released there.
But I am not sure about the workflow here. Or are the tests run always on upstream?

dnf-behave-tests/features/makecache.feature
kontura: Yes its probably my fault I have been using the regexes a lot to make my life easier, maybe at the cost of readability. I will try to do better.

m-blaha: Especially in this negative form - stdout does not contain "regex" I consider it fragile. If dnf change its output format, the test will continue to pass regardless dnf downloads the metadata or not. The better approach will probably be "positive" testing stdout is "the output".

lukash: That's actually a very good point. I agree on the `stdout is ...` front :slightly_smiling_face:

kontura:
I have updated the test, I tried using the step "stdout is" but the speeds in the transaction table are unstable so I decided that for such a simple test it may better to just use stdout contains in the positive form (no regexes this time). :-)

dnf-behave-tests/features/makecache.feature
```diff
@@ -0,0 +1,12 @@
...
+
+@not.with_os=rhel__eq__8
+@fixture.httpd
+@bz1745170
+Scenario: disabled makecache --timer does not invalidate cached metadata
+Given I use the http repository based on "dnf-ci-fedora"
+  And I successfully execute dnf with args "makecache"
+ When I execute dnf with args "makecache --timer --setopt=metadata_timer_sync=0"
+  And I execute dnf with args "install setup -v"
+ Then stdout contains "repo: using cache for: http-dnf-ci-fedora"
```
lukash: @kontura can you please rebase and update this, I think the "http-" prefix is now gone from the name of the repo.

PR #649, opened 2019-10-10
Link RhBug:1647345 with already existing test

kontura:
https://bugzilla.redhat.com/show_bug.cgi?id=1647345


lukash:
Not sure how valuable the tag is in this case (the bug doesn't contain any additional information), but perhaps for the linkage...

PR #650, opened 2019-10-10
Fix reposync test not to rely on download order

m-blaha:


dnf-behave-tests/features/plugins-core/reposync.feature
```diff
@@ -148,12 +148,9 @@ Scenario: Reposync preserves remote timestamps of packages
...
    Then the exit code is 0
-    And stdout matches line by line
-    """
-    reposync .*
-    \(1/2\): wget-1\.0-1\.fc29\.x86_64\.rpm .*
-    \(2/2\): wget-1\.0-1\.fc29\.src\.rpm .*
-    """
+    And stdout contains "\([12]/2\): wget-1\.0-1\.fc29\.x86_64\.rpm .*"
+    And stdout contains "\([12]/2\): wget-1\.0-1\.fc29\.src\.rpm .*"
+    And stdout does not contain "\(3/2\): .*"
```
lukash: So if a third (or more) package was there, it wouldn't be (3/2), right, it would be (3/3) or more?

m-blaha: Sorry, my bad.

m-blaha: I removed this line, it is useless. As far as we are testing 1/2 and 2/2 downloads, it is clear that there is no third package.

PR #655, opened 2019-10-18
Preserve temporary directories when run in container with -{r,R}

lukash:
An alternative to this would be to always preserve directories when run
in a container, but this keeps the output cleaner for normal runs, since
behave now prints the information about preserved directories for each
scenario.

### DIFF
--- a/dnf-behave-tests/launch-test
+++ b/dnf-behave-tests/launch-test
@@ -32,6 +32,14 @@ if [ "$DNFCOMMAND" ]; then
     DEFINES="$DEFINES -D dnf_command=$DNFCOMMAND"
 fi
 
+if [ "$RESERVE" == "-r" ]; then
+    # preserve all temporary directories
+    DEFINES="$DEFINES -D preserve=1"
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/launch-test
```diff
@@ -32,6 +32,14 @@ if [ "$DNFCOMMAND" ]; then
     DEFINES="$DEFINES -D dnf_command=$DNFCOMMAND"
 fi
 
+if [ "$RESERVE" == "-r" ]; then
```
j-mracek: It looks fine, but why to not preserve all directory when tests run in container?

lukash: I've described that in the commit message :wink: 

To extend on that, try to run the tests with the `preserve` flag, you'll see it prints additional text for each scenario telling you about the directories it didn't delete. This is because if you don't run in a container and run the test suite multiple times, your /tmp gets filled up pretty bad.

But in regular containerized runs it's better to not add noise to the output.

PR #660, opened 2019-10-22
Initialize shared httpserver structures only once

m-blaha:


dmnks:
@m-blaha We should probably also remove the `None` checks in the individual getters since they're no longer needed.

Other than that, LGTM.  I can also confirm this PR fixes the following scenarios that would previously fail:
- install-xml-base.feature 
- plugins-core/download-binary-xml-base.feature
- plugins-core/download-binary.feature
- plugins-core/download-debuginfo.feature
- plugins-core/download-source.feature 
- plugins-core/reposync.feature

m-blaha:
Thanks, good point. I removed them.

PR #662, opened 2019-10-24
Make the reposync test more stable

m-blaha:
The test was failing randomly due to random order of "Last metadata
expiration check" and testrepo downloading lines.


jrohel:
LGTM

PR #663, opened 2019-10-24
Tests for check --dependencies and repoquery --requires-pre with %pre and %post

kontura:
Covers:
https://github.com/rpm-software-management/dnf/pull/1512
https://github.com/rpm-software-management/libdnf/pull/827

I created separate packages for the check --dependencies and repoquery tests.

### DIFF
--- a/dnf-behave-tests/features/check-dependencies.feature
+++ b/dnf-behave-tests/features/check-dependencies.feature
@@ -1,12 +1,9 @@
 Feature: Check when there is a package with missing dependency
 
 
-Background: Force installation of an RPM without dependencies
+Scenario Outline: Check <option>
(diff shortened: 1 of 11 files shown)


dnf-behave-tests/features/check-dependencies.feature
```diff
@@ -34,3 +33,31 @@ Examples:
         | --duplicates       |
         | --obsoleted        |
         | --provides         |
+
+
+@bz1543449
+Scenario: Removed scriptlet pre and post dependencies are not reported as missing
+  Given I use repository "dnf-ci-check-dependencies"
```
lukash: AFAIK the names of repositories that belong to single feature files so far don't have the "dnf-ci-" prefix, so let's keep it consistent? (and name the repository just `check-dependencies`?)

On a side note, I'm still worried it's becoming a bit of a mess, that's why I wanted to agree on some rules a while ago...

kontura: I see, I have updated it to `check-dependencies`.

I agree, it confused me at least.

PR #664, opened 2019-10-24
Tests for builddep correctly handling excludes with --best

lucab85:
Tests for builddep correctly handling excludes with --best
    
https://bugzilla.redhat.com/show_bug.cgi?id=1758459

### DIFF
--- a/dnf-behave-tests/features/builddep.feature
+++ b/dnf-behave-tests/features/builddep.feature
@@ -80,3 +80,60 @@ Scenario: Builddep on SPEC with non-available Source0
 
    Error: Some packages could not be found.
    """
+
+@bz1758459
+Scenario: I exclude the highest verion of a package and call dnf builddep with --best
+  Given I use repository "dnf-ci-fedora-updates"
+    And I enable plugin "builddep"
+  Given I create file "dummy-pkg.spec" with
+   """
+   Name: dummy-pkg
+   Summary: dummy-pkg summary
+   Version: 1.0
+   Release: 1
+   License: GPL
+   BuildRequires: flac
+   %description
+   This is a dummy-pkg description
+   %build
+   %files
+   %changelog
+   """
+   When I execute dnf with args "builddep {context.dnf.installroot}/dummy-pkg.spec --exclude flac-1.3.3-3.fc29 --best"
+   Then the exit code is 0
+    And Transaction is following
+        | Action                | Package                    |
+        | install               | flac-0:1.3.3-2.fc29.x86_64 |
+
+@bz1758459
+Scenario: I call dnf builddep with --best on a spec file with a modular dependency (tests handling modular excludes)
+  Given I use repository "dnf-ci-fedora-modular"
+    And I use repository "dnf-ci-fedora"
+    And I enable plugin "builddep"
+  Given I create file "dummy-pkg.spec" with
+   """
+   Name: dummy-pkg
+   Summary: dummy-pkg summary
+   Version: 1.0
+   Release: 1
+   License: GPL
+   BuildRequires: nodejs
+   %description
+   This is a dummy-pkg description
+   %build
+   %files
+   %changelog
+   """
+   When I execute dnf with args "builddep {context.dnf.installroot}/dummy-pkg.spec --best"
+   Then the exit code is 0
+    And Transaction is following
+        | Action                 | Package                                       |
+        | install                | setup-0:2.12.1-1.fc29.noarch                  |
+        | install                | filesystem-0:3.9-2.fc29.x86_64                |
+        | install                | basesystem-0:11-6.fc29.noarch                 |
+        | install                | glibc-all-langpacks-0:2.28-9.fc29.x86_64      |
+        | install                | glibc-common-0:2.28-9.fc29.x86_64             |
+        | install                | glibc-0:2.28-9.fc29.x86_64                    |
+        | install                | npm-1:8.11.4-1.module_2030+42747d40.x86_64    |
+        | install                | nodejs-1:8.11.4-1.module_2030+42747d40.x86_64 |
+        | module-stream-enable   | nodejs:8                                      |


dnf-behave-tests/features/builddep.feature
```diff
@@ -80,3 +80,62 @@ Scenario: Builddep on SPEC with non-available Source0
...
+   %build
+   %files
+   %changelog
+   """
+   When I execute dnf with args "builddep {context.dnf.installroot}/dummy-pkg.spec --exclude nodejs-1:5.12.1-1.fc29 --best"
+   Then the exit code is 0
+    And Transaction contains
+        | Action                | Package                       |
+        | install               | nodejs-1:5.11.1-1.fc29.x86_64 |
+    And stderr does not contain "Traceback"
```
lukash: I don't think we should be checking for the "Traceback" word like this :slightly_smiling_face:. The exit code is 0, so there quite certainly wasn't a traceback.

If the command doesn't normally print anything to `stderr`, you can (and probably should) check it's empty.

dnf-behave-tests/features/builddep.feature
```diff
@@ -80,3 +80,62 @@ Scenario: Builddep on SPEC with non-available Source0
...
+        | install                | setup-0:2.12.1-1.fc29.noarch                  |
+        | install                | filesystem-0:3.9-2.fc29.x86_64                |
+        | install                | basesystem-0:11-6.fc29.noarch                 |
+        | install                | glibc-all-langpacks-0:2.28-9.fc29.x86_64      |
+        | install                | glibc-common-0:2.28-9.fc29.x86_64             |
+        | install                | glibc-0:2.28-9.fc29.x86_64                    |
+        | install                | npm-1:8.11.4-1.module_2030+42747d40.x86_64    |
+        | install                | nodejs-1:8.11.4-1.module_2030+42747d40.x86_64 |
+    And stderr does not contain "Traceback"
+    And stderr does not contain "Modular dependency problem"
```
lukash: You also probably shouldn't check for this, because in the specific situation of there being a modular dependency problem, the exit status probably also would be non-zero or at least the output would be different from what you expect? (not entirely sure about this but I think it is so)

Now the command could go haywire in such a way it would actually end with zero return code and the same output and still print the "Modular dependency problem", but then we're getting into the realm of it printing arbitrary errors which this check won't catch (because it's a negative check and that means it will pass on anything but the exact error string). So if you go this path, you end up with the conclusion that you need to check for exact error outputs (as opposed to a "contains" check).

So basically I think this check is redundant. Again, if you can, you should check for stderr being empty. Ultimately, we should be able to check for the exact content of stderr, whatever it is.

dnf-behave-tests/features/builddep.feature
```diff
@@ -80,3 +80,59 @@ Scenario: Builddep on SPEC with non-available Source0
...
+   Release: 1
+   License: GPL
+   BuildRequires: nodejs
+   %description
+   This is a dummy-pkg description
+   %build
+   %files
+   %changelog
+   """
+   When I execute dnf with args "builddep {context.dnf.installroot}/dummy-pkg.spec --exclude nodejs-1:11.0.0-1.module_2311+8d497411.fc29 --best"
```
j-mracek: Please could you provide test without --exclude?

lukash: What @j-mracek means is (we've just discussed this), just drop the `--exclude` switch from this test, there are already modularity excludes present and that's what is being tested.

dnf-behave-tests/features/builddep.feature
```diff
@@ -80,3 +80,62 @@ Scenario: Builddep on SPEC with non-available Source0
 
    Error: Some packages could not be found.
    """
+
+@bz1758459
+Scenario: I can exclude package version and install the best candidate
```
lukash: Please use a bit more concrete description for the scenario, like:
> I exclude the highest verion of a package and call dnf builddep with --best

dnf-behave-tests/features/builddep.feature
```diff
@@ -80,3 +80,62 @@ Scenario: Builddep on SPEC with non-available Source0
...
+   This is a dummy-pkg description
+   %build
+   %files
+   %changelog
+   """
+   When I execute dnf with args "builddep {context.dnf.installroot}/dummy-pkg.spec --exclude nodejs-1:5.12.1-1.fc29 --best"
+   Then the exit code is 0
+    And Transaction contains
+        | Action                | Package                       |
+        | install               | nodejs-1:5.11.1-1.fc29.x86_64 |
```
lukash: Please use the exact match version of this step (instead of `Transaction contains`):
```gherkin
   And Transaction is following
   ...
```

(You'll need to list all the packages in the transaction)

We're planning on deprecating the partial transaction check step, as there is no reason for it (except for being a bit less verbose), checking the exact output is always better in tests.

dnf-behave-tests/features/builddep.feature
```diff
@@ -80,3 +80,62 @@ Scenario: Builddep on SPEC with non-available Source0
...
+    And stderr does not contain "Traceback"
+
+@bz1758459
+Scenario: I can exclude package version and install the best candidate with modules
```
lukash: Similar to the previous step description:
> I call dnf builddep with --best on a spec file with a modular dependency (tests handling modular excludes)

dnf-behave-tests/features/builddep.feature
```diff
@@ -80,3 +80,62 @@ Scenario: Builddep on SPEC with non-available Source0
...
+   BuildRequires: nodejs
+   %description
+   This is a dummy-pkg description
+   %build
+   %files
+   %changelog
+   """
+   When I execute dnf with args "builddep {context.dnf.installroot}/dummy-pkg.spec --exclude nodejs-1:11.0.0-1.module_2311+8d497411.fc29 --best"
+   Then the exit code is 0
+    And Transaction contains
```
lukash: Again, please use `Transaction is following`.

dnf-behave-tests/features/builddep.feature
```diff
@@ -80,3 +80,62 @@ Scenario: Builddep on SPEC with non-available Source0
...
+   Release: 1
+   License: GPL
+   BuildRequires: nodejs
+   %description
+   This is a dummy-pkg description
+   %build
+   %files
+   %changelog
+   """
+   When I execute dnf with args "builddep {context.dnf.installroot}/dummy-pkg.spec --exclude nodejs-1:5.12.1-1.fc29 --best"
```
lukash: Unless I'm mistaken this test isn't related to modularity and so you don't need the `nodejs` package. To avoid adding another version of the `nodejs` package to the `dnf-ci-fedora` repository, could you use a different package that already has multiple versions in the testing repositories? (there are multiple such packages in `dnf-ci-fedora` that have another version in `dnf-ci-fedora-updates`)

dnf-behave-tests/features/builddep.feature
```diff
@@ -80,3 +80,63 @@ Scenario: Builddep on SPEC with non-available Source0
...
+    And Transaction is following
+        | Action                | Package                    |
+        | install               | flac-0:1.3.3-2.fc29.x86_64 |
+
+@bz1758459
+Scenario: I call dnf builddep with --best on a spec file with a modular dependency (tests handling modular excludes)
+  Given I use repository "dnf-ci-fedora-modular"
+    And I use repository "dnf-ci-fedora"
+    And I enable plugin "builddep"
+    And I execute dnf with args "module enable nodejs:8"
```
lukash: Why add the module enable? I'm not sure it has any effect on the result, I would guess not, but... The version without the module enable is what @j-mracek reviewed so I'd stick with that. It's shorter too.

lucab85: Without this code I got the error:
`Assertion Failed: [dnf] Following packages weren't captured in the table for action 'module-stream-enable': nodejs:8`

lukash: Ok, the enabling of the module stream happened automatically in the transaction and you can add it to the transaction like this:
```gherkin
    And Transaction is following
      | Action               | Package                                       |
      ...
      | module-stream-enable | nodejs:8                                      |
```

The way you solved it just took the enable out of the last command output. It's probably functionally equivalent so I'll leave it up to you whether to keep it as is, my way is probably a bit more succinct. The `And modules state is following` at the end is redundant in this test I think...

lukash:
One more thing, you're missing an empty line between the header of the commit message and the description in the commit of this PR. So it's like this:
```
Tests for builddep command with exclude argument
https://bugzilla.redhat.com/show_bug.cgi?id=1758459
```
While the convention (or a rule, even though it doesn't seem to be enforced?) is:
```
Tests for builddep command with exclude argument

https://bugzilla.redhat.com/show_bug.cgi?id=1758459
```

While touching on this, from what I understood a better description would be along the lines of:
> Tests for builddep correctly handling excludes with --best

PR #665, opened 2019-10-25
Add tests for dnf group list

j-mracek:


dnf-behave-tests/features/group.feature
```diff
@@ -185,3 +185,69 @@ Scenario: Group remove does not remove user installed packages
         | group-remove  | DNF-CI-Testgroup                          |
         | unchanged     | filesystem-0:3.9-2.fc29.x86_64            |
         | unchanged     | setup-0:2.12.1-1.fc29.noarch              |
+
+@bz1706382
+Scenario: Group list
+ Given I use repository "dnf-ci-thirdparty"
+   When I execute dnf with args "group list"
+     Then the exit code is 0
+      And stdout is
```
lukash: Indentation off the hook! :slightly_smiling_face: please align. Otherwise LGTM.

j-mracek: @lukash Please could you be more specific? It looks like that everyone tries different schema for intents.

lukash: @j-mracek from what I've seen in our .feature files we (almost?) always align at the end of the step keyword, so that the step text is always aligned as well:
```gherkin
 Given I use repository "dnf-ci-thirdparty"
  When I execute dnf with args "group list"
  Then the exit code is 0
   And stdout is
```
(and then the amount of spaces before the step varies a bit and tables/block text indentation is all over the place)

Scenarios in this file are also that way...

(And by the way I don't like this alignment personally, because I find it hard to maintain manually, but I guess in the particular case of behave steps it is more readable...)

j-mracek:
@lukash Thanks, done

PR #666, opened 2019-10-29
Remove @not.with_os=rhel__eq__8 tags for first rhel-8.2.0 rebase

kontura:


kontura:
Updating tags for rhel-8.2.0, with this patch gating such pass.

PR #669, opened 2019-10-31
Testing script improvements

lukash:
A couple of improvements to `dnf-testing.sh`.

### DIFF
--- a/dnf-testing.sh
+++ b/dnf-testing.sh
@@ -43,6 +43,7 @@ Options:
   -h, --help               Show this help
   -c, --container  IMAGE   Specified Image ID or name if do not want to run the last built image
   -d, --devel              Share local feature/ with docker
+  -f, --file               Path to Dockerfile to use
   -p, --podman             Force using podman instead of docker
   -r, --reserve            Keep bash shell session open after every single test executed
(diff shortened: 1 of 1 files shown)


dnf-testing.sh
```diff
@@ -142,7 +145,7 @@ list()
 build()
 {
     local output=($($DOCKER_BIN build --build-arg TYPE="$type" \
-                    ${BUILD_CACHE} --force-rm -t "$IMAGE" "$PROG_PATH" | \
+                    ${BUILD_CACHE} --tls-verify=false --force-rm -t "$IMAGE" -f "$DOCKER_FILE" "$PROG_PATH" | \
```
m-blaha:
```
$ docker --tls-verify=false build
unknown flag: --tls-verify
```
the flag is `--tlsverify` for **docker** and as it is global option it must be used before the command (build here).
For **podman** on the other hand, the flag is `--tls-verify` and is an option of the build command and must be used after the command.

lukash: Oops, sorry, I've left this in there by accident. Too many issues at once. Not sure what we want to do about this one, I've removed it for now.

m-blaha:
LGTM

PR #673, opened 2019-11-06
repoquery: Add a test for disabling modular filtering

lukash:


dnf-behave-tests/features/repoquery/modules.feature
```diff
@@ -0,0 +1,22 @@
+Feature: repoquery tests for handling modularity.
+
+Background:
+ Given I use repository "repoquery-modules"
+
+
+Scenario: Test --disable-modular-excludes
```
pkratoch: Can you please fix the option name here? Should be ``--disable-modular-filtering``.

pkratoch:
Can you also add one non-modular nodejs package? It should also be visible with ``--disable-modular-filtering`` only.

Also, how about adding one scenario for enabled stream, too?

lukash:
@pkratoch I've fixed the scenario description, added a non-modular package and a test for an enabled module stream.

pkratoch:
Thanks for the changes. Looks good to me now; I will merge it once the dnf PR is ready.

PR #675, opened 2019-11-08
Tests for post-transaction-actions plugin

m-blaha:


jrohel:
Seems OK. Local tests (of this test) with tested plugin passed.

PR #676, opened 2019-11-08
Adjust output according to DNF new code

j-mracek:
The message "No match for argument:" was changed to "All matches were
excluded by modular filtering for argument:" to be more informative

### DIFF
--- a/dnf-behave-tests/features/config.feature
+++ b/dnf-behave-tests/features/config.feature
@@ -38,7 +38,7 @@ Scenario: Test with dnf.conf in installroot (dnf.conf is taken from installroot)
     """
    When I execute dnf with args "install filesystem"
    Then the exit code is 1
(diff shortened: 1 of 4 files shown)


dnf-behave-tests/features/config.feature
```diff
@@ -38,7 +38,7 @@ Scenario: Test with dnf.conf in installroot (dnf.conf is taken from installroot)
     """
    When I execute dnf with args "install filesystem"
    Then the exit code is 1
-    And stdout contains "No match for argument: filesystem"
+    And stdout contains "All matches were excluded by regular filtering for argument: filesystem"
```
lukash: Would you please consider changing these to an exact match using `stdout/stderr is`? While not entirely necessary, it is an improvement and in this particular case I had a doubt there isn't other text in the output so I went to test it manually. If there was an exact match, it would actually save me time...

PR #677, opened 2019-11-08
Stop parsing of docker/podman output

m-blaha:
The output differs in different versions of docker and podman and this
caused problems with parsing of the output.
The only user of the parsed image id was CI jenkins job. But this job
can happily use unique tag generated by uuidgen.

### DIFF
--- a/dnf-testing.sh
+++ b/dnf-testing.sh
@@ -148,24 +148,10 @@ list()
 
 build()
 {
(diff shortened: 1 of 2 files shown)


m-blaha:
Please let me know before merging of this PR. The jobs on jenkins need to be rebuilt.

janblazek:
@m-blaha, thank you for fixing this! The changes look good to me. I would like to cherry-pick it to 8.1.0 branch too.

PR #682, opened 2019-11-12
Tests for dnf remote configuration files

m-blaha:


dnf-behave-tests/features/config.feature
```diff
@@ -172,3 +172,58 @@ Scenario: Lines that contain only whitespaces do not spoil previous config optio
...
+@bz1721091
+@fixture.httpd
+Scenario: Dnf can use config file from remote location
+  Given I copy repository "miscellaneous" for modification
+    And I create file "/{context.dnf.repos[miscellaneous].path}/remote.conf" with
+    """
+    [repo-from-remote-config]
+    baseurl=http://some.url/
+    """
+    And I use repository "miscellaneous" as http
```
lukash: Hm, so while this does work, you are abusing the `miscellaneous` repo to set up the HTTP server just to have the config on it. I think we'd be better off adding a simple `I set up a http server for directory <dir>` and use that, because the whole `miscellaneous` repo is redundant in the setup and makes it quite a bit more complex than it needs to be...

m-blaha: Well, you are completely right. I tried to sneak this quick and dirty hack in ;). I'll rework the tests the clean way.

dnf-behave-tests/features/config.feature
```diff
@@ -172,3 +172,53 @@ Scenario: Lines that contain only whitespaces do not spoil previous config optio
...
+     Curl error \(1\): Unsupported protocol for xxxx://localhost:[\d]+/does-not-exist\.conf \[Protocol "xxxx" not supported or disabled in libcurl\]
+   """
+   # host unknown
+   When I execute dnf with args "-c http://the_host:{context.dnf.ports[/remotedir]}/does-not-exist.conf repolist repo-from-remote-config"
+   Then the exit code is 1
+   And stderr matches line by line
+   """
+   Config error: Configuration file URL "http://the_host:[\d]+/does-not-exist\.conf" could not be downloaded:
+     Curl error \(6\): Couldn't resolve host name for http://the_host:[\d]+/does-not-exist\.conf \[Could not resolve host: the_host\]
+   """
```
lukash: Could you add one more test for a local config file not found error? Seems we don't have that yet, it's a bit related. Thanks!

m-blaha: There is already one in config.feature Scenario: Test usage of not existing config file

PR #685, opened 2019-11-13
Unify colors used with behave

m-blaha:
Using the same ANSI escape sequences as behave. Behave strips these
escapes when the output goes to junit file (which is the case for
Jenkins).

### DIFF
--- a/dnf-behave-tests/features/steps/common/string.py
+++ b/dnf-behave-tests/features/steps/common/string.py
@@ -5,6 +5,7 @@ from __future__ import print_function
 
 import re
 
+from behave.formatter.ansi_escapes import escapes
 import sys
 PY3 = sys.version_info.major >= 3
 if PY3:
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/steps/common/string.py
```diff
@@ -31,11 +32,10 @@ def print_lines_diff(expected, found, num_lines_equal=0):
...
     print("{:{left_width}}  |  {}".format("expected", "found", left_width=left_width))
 
-    green, red, reset = "\033[1;32m", "\033[1;31m", "\033[0;0m"
-
     for line in zip_longest(expected, found, fillvalue=""):
-        col = green if num_lines_equal > 0 or line[0] == line[1] else red
-        print("{}{:{left_width}}  |  {}{}".format(col, line[0], line[1], reset, left_width=left_width))
+        col = escapes['passed'] if num_lines_equal > 0 or line[0] == line[1] else escapes['failed']
+        print("{}{:{left_width}}  |  {}{}".format(
+            col, line[0], line[1], escapes['reset'], left_width=left_width))
```
lukash: Ah, so does this print no escape codes if colors are turned off? If so, great, it was an easier fix than I anticipated...

m-blaha: Not sure about colors being turned off. But junit formatter strips some escape codes (those that behave defines). Unfortunately escape sequences used in our tests were a bit different than those in behave => no stripping => jenkins plugin failed while parsing the file.

lukash: I see. Ok, if this helps...

Anyway, could you use `'passed_arg'` instead of `'passed'` and `'failed_arg'` instead of `'failed'` for the color indexes?

Those are the foreground versions of the colors, at least in my setup they are brighter and I prefer it personally...

My rant would be I'm not sure why the background colors aren't used just for background, seems to me you either need to set very bright colors that makes my eyes bleed or you have readability issues with certain combinations if background colors are used for text... :slightly_frowning_face:

m-blaha: Sure, I did not notice that behave have also those variants.

PR #686, opened 2019-11-13
Kernel obsoletes tests

j-mracek:


dnf-behave-tests/features/install-obsoletes.feature
```diff
@@ -100,3 +100,39 @@ Scenario: Upgrading obsoleted package by its obsoleter keeps userinstalled=false
         """
         Error: No packages to list
         """
+
+Scenario: Install obsoleting package and inherit the best reason - user (with --nobest)
```
pkratoch: There is ``--nobest`` in the description but not in the test.
Also, does it make sense to test both variants (``--best`` and ``--nobest``)? If so, can you add the other variant?

j-mracek: I am sorry, (with --nobest) is incorrectly there. I have to remove it

PR #691, opened 2019-11-15
Tests for microdnf repolist command

jrohel:
Prepares support for microdnf tests.
Adds tests for microdnf repolist command.

Depends on implementation of repolist command  https://github.com/rpm-software-management/microdnf/pull/54

### DIFF
--- a/dnf-behave-tests/features/environment.py
+++ b/dnf-behave-tests/features/environment.py
@@ -88,7 +88,7 @@ active_tag_matcher = VersionedActiveTagMatcher(active_tag_value_provider)
 
 
 class DNFContext(object):
-    def __init__(self, userdata, force_installroot=False):
+    def __init__(self, userdata, force_installroot=False, no_installroot=False):
         self._scenario_data = {}
 
         self.repos = {}
@@ -97,7 +97,11 @@ class DNFContext(object):
         self.tempdir = tempfile.mkdtemp(prefix="dnf_ci_tempdir_")
         # some tests need to be run inside the installroot, it can be forced
         # per scenario by using @force_installroot decorator
-        if "installroot" in userdata and not force_installroot:
+        if no_installroot:
+            self.installroot = "/"
+            # never delete root
+            self.delete_installroot = False
+        elif "installroot" in userdata and not force_installroot:
             self.installroot = userdata["installroot"]
             # never delete user defined installroot - this allows running tests on /
             self.delete_installroot = False
@@ -271,7 +275,8 @@ def before_scenario(context, scenario):
         scenario.skip(reason="DESTRUCTIVE")
 
     context.dnf = DNFContext(context.config.userdata,
-                             force_installroot='force_installroot' in scenario.tags)
+                             force_installroot='force_installroot' in scenario.tags,
+                             no_installroot='no_installroot' in scenario.effective_tags)
 
 
 def after_scenario(context, scenario):
--- a/dnf-behave-tests/features/microdnf/repolist-disabled.feature
+++ b/dnf-behave-tests/features/microdnf/repolist-disabled.feature
@@ -1,0 +1,53 @@
+@no_installroot
+@destructive
+Feature: Repolist when all repositories are disabled
+
+
+Background:
+  Given I delete file "/etc/dnf/dnf.conf"
+    And I delete file "/etc/yum.repos.d/*.repo" with globs
+    And I use repository "dnf-ci-fedora" with configuration
+        |key      | value |
+        | enabled | 0     |
+
+
+Scenario: Repolist without arguments
+   When I execute microdnf with args "repolist"
+   Then the exit code is 0
+    And stdout is empty
+
+
+Scenario: Repolist with "--enabled"
+   When I execute microdnf with args "repolist --enabled"
+   Then the exit code is 0
+    And stdout is empty
+
+
+Scenario: Repolist with "--disabled"
+   When I execute microdnf with args "repolist --disabled"
+   Then the exit code is 0
+    And stdout matches line by line
+      """
+      repo id\s+repo name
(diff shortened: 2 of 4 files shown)


dnf-behave-tests/features/environment.py
```diff
@@ -297,6 +297,8 @@ def after_scenario(context, scenario):
 
 
 def before_feature(context, feature):
+    if 'disable_installroot' in feature.tags:
+        context.config.userdata["installroot"] = "/"
```
lukash: I'd call the tag `@no_installroot`. It better describes the fact you want to run the test only when tests without installroot or destructive tests were requested. I do think the tag is the way to do this (right, @m-blaha?).

There's a `DESTRUCTIVE_TAGS` list at the beginning of `environment.py`. How about generalizing this a bit and make both "@destructive` and `@no_installroot` work as both `Feature` and `Scenario` tags and include the `@no_installroot` tag in the destructive list? Then you wouldn't need to specify the `@destructive` tag on every scenario, which feels rather redundant. It will require a bit of juggling with the variables in different scopes, but shouldn't be too hard.

dnf-behave-tests/features/microdnf/repolist-disabled.feature
```diff
@@ -0,0 +1,40 @@
...
+   When I execute microdnf with args "repolist --enabled"
+   Then the exit code is 0
+    And stdout is empty
+
+
+@destructive
+Scenario: Repolist with "disabled"
+   When I execute microdnf with args "repolist --disabled"
+   Then the exit code is 0
+    And stdout contains "dnf-ci-fedora\s+dnf-ci-fedora"
```
lukash: Could you use a single `stdout is` step to check the output if at all possible (for all the tests)?

dnf-behave-tests/features/environment.py
m-blaha: The major problem here is that we should not tamper with global namespace `context.config`. This will cause that all subsequent features will have installroot set to '/'.
I'm not sure that `@no_installroot` always means that the test is destructive, but I'm definitely agree that one `@destructive` decorator on `Feature` level should be enough.

lukash: > I'm not sure that @no_installroot always means that the test is destructive

Not necessarily, that's true. But most of the time, even if the command doesn't modify anything (like this case), there's presumably some setup, like the main config, repos, etc., to the point I'd quite like to just declare them destructive to make it clear...

Also, I think we've got a lot more work to do with regards to the no_installroot tests, as atm. they all modify the environment for all the following tests, which is bad. I'm thinking we'll need to ideally run each (no_installroot) test in a separate container to maintain the isolation...

> The major problem here is that we should not tamper with global namespace context.config. This will cause that all subsequent features will have installroot set to '/'.

So how about fixing the design a bit... The `DNFContext` class should actually be called `ScenarioContext`, as it is per-scenario configuration with per-scenario lifetime. Then we can add a `FeatureContext` as an analogy, with a per-feature lifetime. Then we can have a config dict being passed down from `Context` -> `FeatureContext` -> `ScenarioContext` with a possibility to override values?

m-blaha: In fact we do have this. Behave itself creates those feature/scenario contexts (see https://behave.readthedocs.io/en/latest/api.html#behave.runner.Context) and that (`context.feature.something`) is what I suggested to jrohel to use.

The problem with not-enouth-separated scenarios exists. It could be solvable if we change the way how tests are run in containers. Now they are ran one feature file per container. There are some options in `behave` how to achieve running single scenario only - although they are kind of clumsy:

-t <tag> - may be we could somehow manage to have scenarios automatically tagged by some unique identifier

-n <scenario name> 

Although tags and names could contain placeholders (see https://behave.readthedocs.io/en/latest/new_and_noteworthy_v1.2.5.html) we could expect some problems with `Scenario Outlines` which are used quite lot .

m-blaha: And of course running tests one scenario per container could have significant performance impact.

lukash: > Behave itself creates those feature/scenario contexts ... and that ... is what I suggested to jrohel to use.

Cool :slightly_smiling_face: haven't noticed those before.

> There are some options in behave how to achieve running single scenario only - although they are kind of clumsy:
>
> `-t <tag>` - may be we could somehow manage to have scenarios automatically tagged by some unique identifier

Doesn't seem probable that will work...

> `-n <scenario name> `

This one really is clumsy. I did notice it works if you just use a unique prefix of the string... I think. The question really is if it will work with some way we can list the scenarios... (and the placeholders etc., though behave might just be magically handling it)

> And of course running tests one scenario per container could have significant performance impact.

Indeed... But I doubt we have other real options. As I said, it's gonna be (IMO) quite a bit of work to do this and make it convenient enough to use... We don't need it now, but when we need it, we may be in a time constrained situation, so some planning might be worth it as well :wink:

dnf-behave-tests/features/microdnf/repolist-disabled.feature
```diff
@@ -0,0 +1,40 @@
...
+Scenario: Repolist with "--enabled"
+   When I execute microdnf with args "repolist --enabled"
+   Then the exit code is 0
+    And stdout is empty
+
+
+Scenario: Repolist with "--disabled"
+   When I execute microdnf with args "repolist --disabled"
+   Then the exit code is 0
+    And stdout contains "dnf-ci-fedora\s+dnf-ci-fedora test repository$"
```
lukash: Is the output variable in any way? I think not, so can you please use the `stdout is ...` step for all the scenarios?

jrohel:
I replaced `stdout contains` by more strict `stdout matches line by line`.

PR #695, opened 2019-11-19
Add tests for repoinfo command

j-mracek:


dnf-behave-tests/features/repoinfo.feature
```diff
@@ -0,0 +1,116 @@
...
+    And stdout is
+"""
+Repo-id            : dnf-ci-fedora
+Repo-name          : dnf-ci-fedora test repository
+Repo-revision      : 1550000000
+Repo-updated       : Mon Nov 18 20:00:44 2019
+Repo-pkgs          : 289
+Repo-available-pkgs: 289
+Repo-size          : 2.0 M
+Repo-baseurl       : file:///opt/behave/fixtures/repos/dnf-ci-fedora
```
m-blaha: although this url could be correct when run in the docker container, if you run the test locally you would get different path - in my case `file:///home/mblaha/src/ci-dnf-stack/dnf-behave-tests/fixtures/repos/dnf-ci-fedora`

dnf-behave-tests/features/repoinfo.feature
```diff
@@ -0,0 +1,116 @@
...
+Repo-expire        : 172800 second(s) (last: Mon Nov 18 20:00:44 2019)
```
m-blaha: again - the date will vary and on top of that also the number of seconds format depends on the locale setting. I see `172,800 second(s)` here.

dnf-behave-tests/features/repoinfo.feature
```diff
@@ -0,0 +1,116 @@
...
+Repo-filename      : /tmp/dnf_ci_installroot_ww3liqzs/etc/yum.repos.d/dnf-ci-fedora.repo
```
m-blaha: The installroot directory is unique for each tests run.

dnf-behave-tests/features/repoinfo.feature
```diff
@@ -0,0 +1,116 @@
...
+   Then the exit code is 0
+    And stdout is
+"""
+Repo-id            : dnf-ci-fedora
+Repo-name          : dnf-ci-fedora test repository
+Repo-revision      : 1550000000
+Repo-updated       : Mon Nov 18 20:00:44 2019
+Repo-pkgs          : 289
+Repo-available-pkgs: 0
+Repo-size          : 2.0 M
```
m-blaha: The repo size is calculated from available packages => I'm getting 0 here instead of 2.0 M

j-mracek:
@m-blaha Any suggestion how to resolve variable outputs?

m-blaha:
You can use "Then stdout matches line by line" step which checks that each line of stdout matches respective line in regular expressions. The drawback is that you need to escape all dots and parenthesis in the pattern (as the pattern is a list of regular expressions).

dnf-behave-tests/features/repoinfo.feature
```diff
@@ -0,0 +1,116 @@
...
+        | key     | value |
+        | enabled | 0     |
+
+
+Scenario: Repolist without arguments
+   When I execute dnf with args "repoinfo"
+   Then the exit code is 0
+    And stdout is
+"""
+Repo-id            : dnf-ci-fedora
```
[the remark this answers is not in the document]
j-mracek: I am sorry, but an empty line was not fixed yet.

dnf-behave-tests/features/repoinfo.feature
```diff
@@ -0,0 +1,121 @@
...
+"""
+
+Repo-id            : dnf-ci-fedora
+Repo-name          : dnf-ci-fedora test repository
+Repo-revision      : 1550000000
+Repo-updated       : .*
+Repo-pkgs          : 289
+Repo-available-pkgs: 289
+Repo-size          : 2.0 M
+Repo-baseurl       : .*/opt/behave/fixtures/repos/dnf-ci-fedora
```
m-blaha: The Repo-baseurl strongly depends on the directory the test was run from. Currently the test passes when run using dnf-testing.sh in the container. But if you ran it directly using behave, the path would change - for example I have "Repo-baseurl       : file:///home/mblaha/src/ci-dnf-stack/dnf-behave-tests/fixtures/repos/dnf-ci-fedora" there.
The same for following tests.

PR #696, opened 2019-11-20
Test reposync respects excludes, but not modular excludes (RhBug:1750…

pkratoch:
…273)

### DIFF
--- a/dnf-behave-tests/features/plugins-core/reposync.feature
+++ b/dnf-behave-tests/features/plugins-core/reposync.feature
@@ -102,6 +102,50 @@ Scenario: Reposync downloads packages from all streams of modular repository eve
     And file "//{context.dnf.tempdir}/dnf-ci-fedora-modular/x86_64/nodejs-11.0.0-1.module_2311+8d497411.x86_64.rpm" exists
 
 
+@not.with_os=rhel__eq__8
+@bz1686602
+Scenario: Reposync respects excludes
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/plugins-core/reposync.feature
```diff
@@ -102,6 +102,52 @@ Scenario: Reposync downloads packages from all streams of modular repository eve
...
+
+@not.with_os=rhel__eq__8
+Scenario: Reposync respects excludes, but not modular excludes
+  Given I use repository "dnf-ci-fedora-modular" as http
+   When I execute dnf with args "reposync --download-path={context.dnf.tempdir} --setopt=includepkgs=postgresql"
+   Then the exit code is 0
+    And stdout contains ": postgresql-6\.1-1\."
+    And stdout contains ": postgresql-6\.1-1\."
+    And stdout contains ": postgresql-9\.6\.8-1\."
+    And stdout contains ": postgresql-9\.6\.8-1\."
```
m-blaha: Why are these duplicated lines needed?

pkratoch: They are not. I originally distinguished between "srt" and "x86_64" packages, but then it turned out the name is wrapped when too long (which I don't think is a bug). I'll remove the additional lines.

pkratoch: Done.

PR #697, opened 2019-11-20
Improved excludes report

j-mracek:


dnf-behave-tests/features/config.feature
```diff
@@ -38,7 +38,7 @@ Scenario: Test with dnf.conf in installroot (dnf.conf is taken from installroot)
     """
    When I execute dnf with args "install filesystem"
    Then the exit code is 1
-    And stdout contains "No match for argument: filesystem"
+    And stdout contains "All matches were filtered out by exclude filtering for argument: filesystem"
     And stderr contains "Error: Unable to find a match"
```
lukash: I'd still prefer using an exact match for these, unless you have a reason not to? Otherwise looks good.

j-mracek: Changed

PR #698, opened 2019-11-21
Tests for "module repoquery" and "module remove --all"

m-blaha:


dnf-behave-tests/features/module/remove.feature
```diff
@@ -166,3 +166,43 @@ Given I use repository "dnf-ci-fifthparty"
         | remove                    | leia-0:1.0-1.x86_64    |
         | remove                    | luke-0:1.0-1.x86_64    |
 
+
+Scenario: Remove command respects --all switch to remove all packages
+  Given I successfully execute dnf with args "install nodejs-docs"
+   When I execute dnf with args "module remove nodejs --all"
```
j-mracek: What about to add step `"module remove nodejs"` before step with all. The new step should not produce any transaction.

m-blaha: Actually it would remove a lot (everything except the nodejs-docs package) - the `default` profile is installed in Background section using `module install nodejs/default`.

j-mracek: Sorry, the note was not relevant.

PR #700, opened 2019-11-22
Test for downgrade/upgrade exit codes

m-blaha:


dnf-behave-tests/features/downgrade.feature
```diff
@@ -41,3 +41,82 @@ Scenario: Downgrade RPM that requires downgrade of dependency
...
+    """
+    <REPOSYNC>
+    Dependencies resolved.
+    Nothing to do.
+    Complete!
+    """
+    And stderr is
+    """
+    Package flac of lowest version already installed, cannot downgrade it.
+    """
```
lukash: Wouldn't this be better split into multiple scenarios? The main difference being the scenarios would be run independently, here the first fail causes the rest not to be run. Since this even tests two different commands (upgrade/downgrade), I think it would be better to split into smaller pieces.

For the common setup these tests have, `Given I use repository "dnf-ci-fedora-updates"` is shared with all the scenarios, so that can go into `Background`. And then `"dnf-ci-fedora"` is only required for a couple of tests it seems.

Otherwise LGTM.

m-blaha: One of the goals for this test was to make sure that upgrade/downgrade commands behave consistently - that they return the same exit code in the same situations. So I would like to keep them together. But I agree the scenario could be split into three (four) independent tests - non existent package, not installed package, already on highest (lowest) version. The background bit of course could be done.

m-blaha: Eventually I've split them into multiple scenarios.

PR #703, opened 2019-11-25
Fix problem when CI runs with yum command

j-mracek:
Some of outputs is changed according to binary that is used. When yum
command is used, then outputs refer to yum.


dnf-behave-tests/features/steps/cmd.py
```diff
@@ -122,6 +120,9 @@ def step_impl(context):
 def step_impl(context):
     context.dnf._set("disable_plugins", False)
 
+@behave.given("I set dnf command to \"{command}\"")
+def step_impl(context, command):
+    context.dnf._set("dnf_command", command)
```
lukash: Why not just do `context.dnf_command = command`? the `_set` method just looks like a roundabout way to do the same...

kontura:
Since @j-mracek is unavailable I rebased his change and addressed @lukash's note in https://github.com/rpm-software-management/ci-dnf-stack/pull/710.

PR #705, opened 2019-11-25
Test for history new command

m-blaha:
Requires: https://github.com/rpm-software-management/libdnf/pull/854
Requires: https://github.com/rpm-software-management/dnf/pull/1541


lukash:
Can you add a test for a failure to remove or is there no way to make it fail in the tests? :slightly_smiling_face:

PR #706, opened 2019-11-25
Common steps

lukash:
This PR prepares the CI stack for multiple test suites next to the dnf one we have now in `dnf-behave-tests/features`. In preparation for _that_, the first few commits move some code around to consolidate the modules a bit. Then a new `common` directory is introduced and the code that is certain to be shared is moved over to there. Then there are some commits moving code out of `environment.py`, as we will want to reuse that code as well, or at least prepare to do so.

Notable exception is the `VersionedActiveTagMatcher`, which should be trivial to move later, when it's needed.


lukash:
The remaining sider issues are either unavoidable (in the case of module imports for the common steps) or copy-pasted, so I'm not planning to fix them. But if you want me to fix some, @m-blaha, let me know.

m-blaha:
Good work Lukas, thanks for it!

PR #710, opened 2019-11-26
rhel-8.2.0 fixes

kontura:
I removed unnecessary @not.with_os=rhel__eq__8 and included rebased change from https://github.com/rpm-software-management/ci-dnf-stack/pull/703.

Our gating should pass with these changes.

### DIFF
--- a/dnf-behave-tests/features/help.feature
+++ b/dnf-behave-tests/features/help.feature
@@ -1,6 +1,7 @@
 Feature: Help command
 
-Scenario: General help
+Scenario: General help (dnf)
+  Given I set dnf command to "dnf"
    When I execute dnf with args "--help"
    Then the exit code is 0
     And stdout contains "List of Main Commands"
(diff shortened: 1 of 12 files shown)


dnf-behave-tests/features/help.feature
```diff
@@ -12,9 +13,30 @@ Scenario: General help
...
+   When I execute dnf with args "help"
+   Then the exit code is 0
+    And stdout contains "List of Main Commands"
+   When I execute dnf with args "unknown-command"
+   Then the exit code is 1
+    And stderr is
+   """
+   No such command: unknown-command. Please use /usr/bin/yum --help
+   It could be a YUM plugin command, try: "yum install 'dnf-command(unknown-command)'"
+   """
```
m-blaha: It would make more sense to me to be able to run the whole test suite either with `yum` or `dnf` command.
So instead of making specific scenarios for yum/dnf make the scenarios more generic, i.e:
leave out Given I set dnf command to "yum" step
and change stderr checking to
```
And stderr is
   """
   No such command: unknown-command. Please use /usr/bin/{context.dnf.dnf_command} --help
   It could be a YUM plugin command, try: "{context.dnf.dnf_command} install 'dnf-command(unknown-command)'"
   """
```
The problem with upper case `YUM/DNF` string remains - it could be solved e.g. by adding uppercase variant to the context.dnf.

What do you think?

kontura: I would also prefer this approach, but I think the `context.dnf.dnf_command` can also be a path to the program right? It could be something like `/usr/bin/dnf-3` so we would have to process that.

lukash: Yes, majority of the time I'm running the tests against a locally built dnf stack and I'm using a wrapper script to call it, and then call behave like: `behave -Ddnf_command=my-dnf ...`

I agree with @m-blaha it would be better to run the whole stack with either yum or dnf, but the reason for that is that there is a requirement that DNF says "YUM" in its output if it was run through a yum symlink. I'd like to point out this doesn't make sense and is misleading, because the tool the user is using is still DNF, not YUM. The fact we have to add all this complexity for this rather silly reason, while we also have plenty other work to do that is much more important, is quite frustrating.

For an actual solution, this is what dnf currently does:
```
MAIN_PROG = argparse.ArgumentParser().prog if argparse.ArgumentParser().prog == "yum" else "dnf"
```
I think its as good a hack as it gets :confused: We could do the same in the CI stack...

m-blaha: OK, in order not to delay the gating I tend to merge it and we could improve it in a subsequent PR

lukash: Yeah. If it's blocking the gating, certainly merge it.

kontura: Ok, great. Thank you. I will later create a PR that will change it as discussed.

PR #712, opened 2019-11-27
Add tests for installing and removing multilib packages (RhBug:1745878)

pkratoch:


kontura:
The tests pass, I think its good to have this documented and linked to the bz.

PR #717, opened 2019-12-02
updateinfo show <advisory> of the running kernel after a kernel update

lucab85:
updateinfo show <advisory> of the running kernel after a kernel update

https://bugzilla.redhat.com/show_bug.cgi?id=1728004

### DIFF
--- a/dnf-behave-tests/features/updateinfo.feature
+++ b/dnf-behave-tests/features/updateinfo.feature
@@ -312,3 +312,33 @@ Examples:
     # yum compatibility
     | list bugzillas      |
     | list bzs            |
+
+
+@bz1728004
+Scenario: updateinfo show <advisory> of the running kernel after a kernel update
+   When I execute dnf with args "install kernel"
+   Then Transaction is following
+        | Action        | Package                                  |
+        | install       | kernel-core-0:4.18.16-300.fc29.x86_64    |
+        | install       | kernel-modules-0:4.18.16-300.fc29.x86_64 |
+        | install       | kernel-0:4.18.16-300.fc29.x86_64         |
+  Given I use repository "dnf-ci-fedora-updates"
+    And I execute dnf with args "updateinfo list kernel"
+   Then the exit code is 0
+    And stdout is
+    """
+    <REPOSYNC>
+    FEDORA-2019-348e185000 bugfix kernel-4.19.15-300.fc29.x86_64
+    """
+   When I execute dnf with args "update kernel"
+   Then Transaction is following
+        | Action        | Package                                  |
+        | install       | kernel-core-0:4.19.15-300.fc29.x86_64    |
+        | install       | kernel-modules-0:4.19.15-300.fc29.x86_64 |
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/features/updateinfo.feature
```diff
@@ -312,3 +312,32 @@ Examples:
...
+
+@bz1728004
+Scenario: updateinfo show <advisory> of the running kernel after a kernel update
+   When I execute dnf with args "install kernel"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                                  |
+        | install       | kernel-core-0:4.18.16-300.fc29.x86_64    |
+        | install       | kernel-modules-0:4.18.16-300.fc29.x86_64 |
+        | install       | kernel-0:4.18.16-300.fc29.x86_64         |
```
lukash: These 3 steps seem to be the test setup, so you can use `Given I successfully execute dnf ...` and then you don't need the `Then the exit code is 0`. In theory you shouldn't need the `And Transaction is following` step either, but I guess it makes the test more robust in case someone e.g. inadvertently changes the test data.

dnf-behave-tests/features/updateinfo.feature
```diff
@@ -312,3 +312,32 @@ Examples:
...
+  Given I use repository "dnf-ci-fedora-updates"
+    And I execute dnf with args "updateinfo list kernel"
+   Then the exit code is 0
+    And stdout matches line by line
```
lukash: I think you can use the `stdout is` step here, even though the difference is mall in this case. One advantage of `stdout is` is it prints a colored diff in case of mismatch, this one doesn't (even though it could and should, just that nobody implemented it yet :slightly_smiling_face: )

dnf-behave-tests/features/updateinfo.feature
```diff
@@ -312,3 +312,32 @@ Examples:
...
+   When I execute dnf with args "update kernel"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                                  |
+        | install       | kernel-core-0:4.19.15-300.fc29.x86_64    |
+        | install       | kernel-modules-0:4.19.15-300.fc29.x86_64 |
+        | install       | kernel-0:4.19.15-300.fc29.x86_64         |
+   When I execute dnf with args "updateinfo list kernel"
+   Then the exit code is 0
+    And stdout does not contain "FEDORA-2019-348e185000"
```
lukash: You should also use `stdout is` or a more loose positive match test, the negative match is quite fragile...

lucab85: I agree but unfortunately it's the only match possible

lukash: How is that?

(if you say something's not possible, you'll save a communication round trip if you include the reasoning straight away :wink: )

lucab85: This code follow the AC because checks that the previous kernel advisory is not present

lucab85: Code corrected

PR #718, opened 2019-12-03
Run tests without networking

m-blaha:
During testing PR are tests ran in containers without network access.
This patch unifies the behaviour also for dnf-testing.sh.


lukash:
While I agree the tests should run without networking, won't this cause issues for someone who uses the `-{r,R}` switches and needs to e.g. install an additional package to investigate a failure (e.g. strace, gdb, upstream version of DNF to compare the behavior)? Maybe we should add a switch to turn the networking on?

lukash:
This one's covered by the rewrite of the script to python: #768; closing.

PR #688, opened 2019-11-14
Add test for repoquery --unsatisfied when kernel has missing dependencies

pkratoch:

PR #720, opened 2019-12-05
Add steps to fake kernel release

pkratoch:
Separated from PR https://github.com/rpm-software-management/ci-dnf-stack/pull/688 so that it can be merged sooner.

### DIFF
--- a/dnf-behave-tests/features/steps/cmd.py
+++ b/dnf-behave-tests/features/steps/cmd.py
@@ -71,6 +71,38 @@ def faketime_today(context, when):
     context.execute_steps('when I move the clock backward to "{}"'.format(when))
 
 
+@behave.step("I fake kernel release to {release}")
+def i_fake_kernel_release(context, release):
+    """Override uname() system call to return {release} as release.
+
+    This is useful for faking running kernel, because libdnf determines running kernel in the
+    following way:
+        1. makes uname syscall to get operation system release
+        2. searches for file "/boot/vmlinuz-<release>" (<release> from previous step)
+        3. searches for RPM that provides that file
+           -> this RPM is considered the running kernel
+
+    Note that libdnf only checks running kernel when not running in installroot, therefore,
+    @no_installroot tag is needed.
+
+    By faking uname.release, all that remains to create fake running-kernel package is for the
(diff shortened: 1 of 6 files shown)


lukash:
LGTM, I'll just defer to @m-blaha for the changes to the Dockerfiles, as I'm not entirely sure they won't break anyhting.

dnf-behave-tests/features/steps/cmd.py
```diff
@@ -71,6 +71,30 @@ def faketime_today(context, when):
     context.execute_steps('when I move the clock backward to "{}"'.format(when))
 
 
+@behave.step("I fake kernel release to {release}")
+def i_fake_kernel_release(context, release):
+    """Override uname() system call to return {release} as release.
+
+    Libdnf determines running kernel as an RPM package that provides the following file:
+        "/boot/vmlinuz-" + uname.release
+    By faking uname.release, all that remains to create fake running-kernel package is for the
```
j-mracek: Please could you add note that @no_installroot and @destructive marks are needed for scenario where fake uname is used?

lukash: The step doesn't do anything destructive itself...

I think the need for `@no_installroot` comes from the file needing to be present on the host? This is probably what should be mentioned here.

I also think this:
> Libdnf determines running kernel as an RPM package that provides the following file:
>    "/boot/vmlinuz-" + uname.release

Should be extended by mentioning that dnf does the `uname` syscall and uses the result to look for the file (the `uname.release` above), to make it more clear what exactly is going on.

j-mracek: fake uname does not work with installroot (libdnf implementation), therefore you need to use @no_installroot

pkratoch: Thank you both for feedback. ``@no_installroot`` is among the destructive tags by default. But I added a note that ``@no_installroot`` is necessary to fake running kernel. I also tried to clarify how libdnf checks for running kernel, so please review if it's understandable.

lukash: I think it's clear now :+1:

j-mracek:
LGTM

PR #723, opened 2019-12-09
Test for removing package when other version is locked

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1780370

### DIFF
--- a/dnf-behave-tests/features/plugins-core/versionlock-lock.feature
+++ b/dnf-behave-tests/features/plugins-core/versionlock-lock.feature
@@ -96,6 +96,20 @@ Scenario: I can remove the installed locked version of the package
         | remove        | wget-0:1.19.5-5.fc29.x86_64           |
 
 
+@bz1780370
+Scenario: I can remove installed package when other version is locked
+  Given I successfully execute dnf with args "install wget-1.19.5"
+    And I create file "/etc/dnf/plugins/versionlock.list" with
+    """
+    wget-0:2.0.0-1.fc29.*
+    """
+   When I execute dnf with args "remove wget"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                               |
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/plugins-core/versionlock-lock.feature
```diff
@@ -96,6 +96,20 @@ Scenario: I can remove the installed locked version of the package
         | remove        | wget-0:1.19.5-5.fc29.x86_64           |
 
 
+@bz1780370
```
lukash: Shouldn't you rather tag with the original bug instead of the new one? (and are you going to mark the new bug as a duplicate?)

dnf-behave-tests/features/plugins-core/versionlock-lock.feature
```diff
@@ -96,6 +96,20 @@ Scenario: I can remove the installed locked version of the package
...
+Scenario: I can remove installed package when other version is locked
+  Given I successfully execute dnf with args "install wget-1.19.5"
+    And I create file "/etc/dnf/plugins/versionlock.list" with
+    """
+    wget-0:1.19.6-5.fc29.*
```
lukash: How about making the version more visibly different? Squinting required to discern this one...

dnf-behave-tests/features/plugins-core/versionlock-lock.feature
m-blaha: The new bug is not strictly a duplicate - although they have the same root cause the bugs refer to different commands - downgrade / remove.

dnf-behave-tests/features/plugins-core/versionlock-lock.feature
m-blaha: Done.

PR #708, opened 2019-11-26
Update countme.feature for dropped whitelist

dmnks:
Requires: https://github.com/rpm-software-management/libdnf/pull/855

PR #725, opened 2019-12-10
Allow RHEL in User-Agent

lucab85:
Allow RHEL in User-Agent 

https://bugzilla.redhat.com/show_bug.cgi?id=1777255

### DIFF
--- a/dnf-behave-tests/features/countme.feature
+++ b/dnf-behave-tests/features/countme.feature
@@ -44,6 +44,42 @@ Feature: Better user counting
             | header     | value  |
             | User-Agent | libdnf |
 
+    @bz1676891
+    @destructive
+    @fixture.httpd
+    Scenario: User-Agent header is sent (RHEL)
+        Given I am running a system identified as the "Red Hat Enterprise Linux 8.2 generic"
+          And I use repository "dnf-ci-fedora" as http
+          And I start capturing outbound HTTP requests
+         When I execute dnf with args "makecache"
+         Then every HTTP GET request should match:
+            | header     | value                                                        |
+            | User-Agent | libdnf (Red Hat Enterprise Linux 8.2; generic; Linux.x86_64) |
+
+    @bz1676891
+    @destructive
+    @fixture.httpd
+    Scenario: User-Agent header is sent (RHEL)
+        Given I am running a system identified as the "Red Hat Enterprise Linux 8.2 (Ootpa)"
+          And I use repository "dnf-ci-fedora" as http
+          And I start capturing outbound HTTP requests
+         When I execute dnf with args "makecache"
+         Then every HTTP GET request should match:
+            | header     | value                                                        |
+            | User-Agent | libdnf (Red Hat Enterprise Linux 8.2; generic; Linux.x86_64) |
+
+    @bz1676891
+    @destructive
+    @fixture.httpd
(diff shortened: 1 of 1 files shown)


dmnks:
This is no longer needed (was fixed by https://github.com/rpm-software-management/ci-dnf-stack/pull/708 already).

dnf-behave-tests/features/countme.feature
```diff
@@ -44,6 +44,42 @@ Feature: Better user counting
             | header     | value  |
             | User-Agent | libdnf |
 
+    @bz1676891
+    @destructive
+    @fixture.httpd
+    Scenario: User-Agent header is sent (RHEL)
+        Given I am running a system identified as the "Red Hat Enterprise Linux 8.2 generic"
```
dmnks: Although this PR is already closed, let me just clarify:
"generic" is the default value of `VERSION_ID` in `/etc/os-release` on RHEL, so you wouldn't need to specify it here (but it wouldn't hurt of course).

dnf-behave-tests/features/countme.feature
```diff
@@ -44,6 +44,42 @@ Feature: Better user counting
...
+         When I execute dnf with args "makecache"
+         Then every HTTP GET request should match:
+            | header     | value                                                        |
+            | User-Agent | libdnf (Red Hat Enterprise Linux 8.2; generic; Linux.x86_64) |
+
+    @bz1676891
+    @destructive
+    @fixture.httpd
+    Scenario: User-Agent header is sent (RHEL)
+        Given I am running a system identified as the "Red Hat Enterprise Linux 8.2 (Ootpa)"
```
dmnks: The codename (`(Ootpa)`) is not part of the NAME field in `/etc/os-release`, so it shouldn't be included here either (libdnf wouldn't use it in the User-Agent).

dnf-behave-tests/features/countme.feature
```diff
@@ -44,6 +44,42 @@ Feature: Better user counting
...
+    @destructive
+    @fixture.httpd
+    Scenario: User-Agent header is sent (RHEL)
+        Given I am running a system identified as the "Red Hat Enterprise Linux"
+          And I use repository "dnf-ci-fedora" as http
+          And I start capturing outbound HTTP requests
+         When I execute dnf with args "makecache"
+         Then every HTTP GET request should match:
+            | header     | value  |
+            | User-Agent | libdnf |
```
dmnks: This is actually incorrect; libdnf no longer uses a whitelist so "Red Hat Enterprise Linux" would appear here (along with the version which would have to be specified in the step above for the test to work).

dnf-behave-tests/features/countme.feature
```diff
@@ -44,6 +44,42 @@ Feature: Better user counting
...
+         When I execute dnf with args "makecache"
+         Then every HTTP GET request should match:
+            | header     | value                                                        |
+            | User-Agent | libdnf (Red Hat Enterprise Linux 8.2; generic; Linux.x86_64) |
+
+    @bz1676891
+    @destructive
+    @fixture.httpd
+    Scenario: User-Agent header is sent (RHEL)
+        Given I am running a system identified as the "Red Hat Enterprise Linux"
```
dmnks: The step implementation actually expects a version in this string, so you would have to use "Red Hat Enterprise Linux 8.1" or similar.

PR #726, opened 2019-12-10
Install an RPM without null lines

lucab85:
Install an RPM without null lines

https://bugzilla.redhat.com/show_bug.cgi?id=1691353

### DIFF
--- a/dnf-behave-tests/features/microdnf/install7.feature
+++ b/dnf-behave-tests/features/microdnf/install7.feature
@@ -1,0 +1,20 @@
+@no_installroot
+Feature: Install package
+
+
+Background:
+Given I delete file "/etc/dnf/dnf.conf"
+  And I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+@bz1691353
+Scenario: Install an RPM without null lines
+  Given I use repository "dnf-ci-fedora"
+   When I execute microdnf with args "install lame"
+   Then the exit code is 0
+    And microdnf transaction is
+        | Action        | Package                                   |
+        | install       | lame-0:3.100-4.fc29.x86_64                |
+        | install       | lame-libs-0:3.100-4.fc29.x86_64           |
+  And stdout does not contain "Installing: (null)"
+


dnf-behave-tests/features/microdnf/install.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Install package
+
+@bz1691353
+Scenario: Install an RPM without null lines
+  Given I use repository "dnf-ci-fedora"
+   When I execute dnf with args "install glibc"
```
m-blaha: Isn't this test supposed to test microdnf?

dnf-behave-tests/features/microdnf/install.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Install package
```
m-blaha: All microdnf tests should be decorated by @no_installroot

dnf-behave-tests/features/microdnf/install.feature
lucab85: You're right, I'll adjust the test. It is useful also usage of [PR#731](https://github.com/rpm-software-management/ci-dnf-stack/pull/731)

dnf-behave-tests/features/microdnf/install7.feature
```diff
@@ -0,0 +1,14 @@
+@no_installroot
+Feature: Install package
+
+@bz1691353
+Scenario: Install an RPM without null lines
+  Given I use repository "dnf-ci-fedora"
+   When I execute microdnf with args "install lame"
```
m-blaha: The test is run on the host - not in the install root. That said, all other repositories already enabled on host (like fedora, updates...) are used which makes the test considerably slower (need to sync all metadata for those repositories) and unable to run without the network (which is the way how gating tests are run).
You can solve this by removing all current repo files as first steps (see also other microdnf/install*.feature files in the master branch):
```
Background:
Given I delete file "/etc/dnf/dnf.conf" 
  And I delete file "/etc/yum.repos.d/*.repo" with globs
```

lucab85: Thank you for the explanation.

PR #727, opened 2019-12-10
microdnf install command on packages with xml:base and option

lucab85:
microdnf install command on packages with xml:base and option

https://bugzilla.redhat.com/show_bug.cgi?id=1734350
https://bugzilla.redhat.com/show_bug.cgi?id=1769831
https://bugzilla.redhat.com/show_bug.cgi?id=1771012
https://bugzilla.redhat.com/show_bug.cgi?id=1779757

### DIFF
--- a/dnf-behave-tests/features/microdnf/install-nodocs-1.feature
+++ b/dnf-behave-tests/features/microdnf/install-nodocs-1.feature
@@ -1,0 +1,18 @@
+@no_installroot
+Feature: microdnf install command on packages
+
+
+Background:
+Given I delete file "/etc/dnf/dnf.conf"
+  And I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+Scenario: Install a documentation package from local repodata
+Given I use repository "microdnf-install-nodocs"
+ When I execute microdnf with args "install man-pages"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | man-pages-0:4.16-3.fc29.x86_64            |
+  And file "/usr/share/doc/man-pages/README" exists
+
--- a/dnf-behave-tests/features/microdnf/install1.feature
+++ b/dnf-behave-tests/features/microdnf/install1.feature
@@ -1,0 +1,21 @@
+@no_installroot
+Feature: microdnf install command on packages
+
+
+Background:
+Given I delete file "/etc/dnf/dnf.conf"
+  And I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+@bz1734350
+@bz1779757
+Scenario: Install package from local repodata with local packages
+#1. local repo with local packages
+Given I use repository "dnf-ci-fedora"
+ When I execute microdnf with args "install kernel"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | kernel-core-0:4.18.16-300.fc29.x86_64     |
+      | install       | kernel-modules-0:4.18.16-300.fc29.x86_64  |
+      | install       | kernel-0:4.18.16-300.fc29.x86_64          |
--- a/dnf-behave-tests/features/microdnf/install2.feature
+++ b/dnf-behave-tests/features/microdnf/install2.feature
@@ -1,0 +1,23 @@
+@no_installroot
+Feature: microdnf install command on packages
+
+
+Background:
+Given I delete file "/etc/dnf/dnf.conf"
+  And I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+@bz1734350
+Scenario: Install package from local repodata with local xml:base
+#2. local repo with local packages (different package location specified using xml:base)
+Given I copy repository "dnf-ci-fedora" for modification
+  And I use repository "dnf-ci-fedora"
+  And I execute "createrepo_c --baseurl file://{context.dnf.installroot}/xml_base/dnf-ci-fedora /{context.dnf.repos[dnf-ci-fedora].path}"
+  And I copy directory "{context.dnf.repos[dnf-ci-fedora].path}" to "/xml_base/dnf-ci-fedora"
+ When I execute microdnf with args "install abcde"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | flac-0:1.3.2-8.fc29.x86_64                |
+      | install       | abcde-0:2.9.2-1.fc29.noarch               |
+  And file "/xml_base/dnf-ci-fedora/noarch/abcde-2.9.2-1.fc29.noarch.rpm" exists
--- a/dnf-behave-tests/features/microdnf/install3.feature
+++ b/dnf-behave-tests/features/microdnf/install3.feature
@@ -1,0 +1,22 @@
+@no_installroot
+Feature: microdnf install command on packages
+
+
+Background:
+Given I delete file "/etc/dnf/dnf.conf"
+  And I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+@bz1734350
+Scenario: Install package from local repodata with xml:base pointing to remote packages
+#3. local repo with remote packages (different package location specified using xml:base)
+Given I make packages from repository "dnf-ci-fedora" accessible via http
+  And I copy repository "dnf-ci-fedora" for modification
+  And I execute "createrepo_c --baseurl http://localhost:{context.dnf.ports[dnf-ci-fedora]} /{context.dnf.repos[dnf-ci-fedora].path}"
+  And I use repository "dnf-ci-fedora"
+ When I execute microdnf with args "install abcde"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | flac-0:1.3.2-8.fc29.x86_64                |
+      | install       | abcde-0:2.9.2-1.fc29.noarch               |
--- a/dnf-behave-tests/features/microdnf/install-nodocs-2.feature
+++ b/dnf-behave-tests/features/microdnf/install-nodocs-2.feature
@@ -1,0 +1,19 @@
+@no_installroot
+Feature: microdnf install command on packages
+
+
+Background:
+Given I delete file "/etc/dnf/dnf.conf"
+  And I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+@bz1769831
+Scenario: Install package with option from local repodata with local packages
+Given I use repository "microdnf-install-nodocs"
+ When I execute microdnf with args "--setopt=tsflags=nodocs install man-pages"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | man-pages-0:4.16-3.fc29.x86_64            |
+  And file "/usr/share/doc/man-pages/README" does not exist
+
--- a/dnf-behave-tests/features/microdnf/install-nodocs-3.feature
+++ b/dnf-behave-tests/features/microdnf/install-nodocs-3.feature
@@ -1,0 +1,19 @@
+@no_installroot
+Feature: microdnf install command on packages
+
+
+Background:
+Given I delete file "/etc/dnf/dnf.conf"
+  And I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+@bz1771012
+Scenario: Install package with --nodocs option from local repodata with local packages
+Given I use repository "microdnf-install-nodocs"
+  And I execute microdnf with args "--nodocs install man-pages"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | man-pages-0:4.16-3.fc29.x86_64            |
+  And file "/usr/share/doc/man-pages/README" does not exist
+
--- a/dnf-behave-tests/features/microdnf/install4.feature
+++ b/dnf-behave-tests/features/microdnf/install4.feature
@@ -1,0 +1,28 @@
+@no_installroot
+Feature: microdnf install command on packages
+
+
+Background:
+Given I delete file "/etc/dnf/dnf.conf"
+  And I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+@bz1734350
+Scenario: Install packages from local repodata that have packages with xml:base pointing to a remote as well as local packages
+#4. local repo with local and remote packages; installing both at the same time.
+Given I make packages from repository "dnf-ci-fedora" accessible via http
+  And I copy repository "dnf-ci-fedora" for modification
+  And I copy repository "dnf-ci-thirdparty" for modification
+  And I execute "createrepo_c --baseurl http://localhost:{context.dnf.ports[dnf-ci-fedora]} /{context.dnf.repos[dnf-ci-fedora].path}"
+  And I execute "mergerepo_c --repo file://{context.dnf.repos[dnf-ci-fedora].path} --repo file://{context.dnf.repos[dnf-ci-thirdparty].path}" in "{context.dnf.installroot}"
+  And I configure a new repository "merged-repo" with
+      | key     | value                                        |
+      | baseurl | file://{context.dnf.installroot}/merged_repo |
+  When I execute microdnf with args "install abcde alternator"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | FlacBetterEncoder-0:1.0-1.x86_64          |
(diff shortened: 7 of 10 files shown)


dnf-behave-tests/features/microdnf/install-xml-base.feature
```diff
@@ -0,0 +1,73 @@
+@no_installroot
+@desctructive
```
lukash: You don't need the `@destructive` tag, `@no_installroot` already marks the tests as destructive.

dnf-behave-tests/features/microdnf/install-xml-base.feature
```diff
@@ -0,0 +1,73 @@
...
+Feature: microdnf install command on packages with xml:base
+
+@bz1734350
+Scenario: Install packages from local repodata with local packages
+#1. local repo with local packages
+Given I use repository "dnf-ci-fedora"
+ When I execute microdnf with args "install setup"
```
lukash: Since you're running the tests directly on the system and not in an installroot, there are the system repositories configured and they are being synced when you run microdnf, which takes a long time and may fail in CI due to networking not being available.

You should at least delete the repo configs like it is done e.g. here: https://github.com/rpm-software-management/ci-dnf-stack/blob/master/dnf-behave-tests/features/microdnf/repolist.feature

But, deleting the system repos when not running in an installroot should be part of the test environment setup somewhere in `environment.py`. If you'd like to try to figure out how to do that, that would be great. Actually, it should be pretty simple, when the test is tagged `@no_installroot`, just delete the system repos configuration, I think.

dnf-behave-tests/features/microdnf/install-xml-base.feature
```diff
@@ -0,0 +1,73 @@
+@no_installroot
+@desctructive
+Feature: microdnf install command on packages with xml:base
+
+@bz1734350
+Scenario: Install packages from local repodata with local packages
```
lukash: This is a regular package installation test and you're testing various combinations of repository urls, not just `xml:base`, so (unless you decide to split the tests into multiple files - see my other comments), why not just name the file `microdnf/install.feature`?

dnf-behave-tests/features/microdnf/install-xml-base.feature
```diff
@@ -0,0 +1,73 @@
...
+
+
+@bz1734350
+Scenario: Install packages from local repodata with local xml:base
+#2. local repo with local packages (different package location specified using xml:base)
+Given I copy repository "dnf-ci-fedora" for modification
+  And I use repository "dnf-ci-fedora"
+  And I execute "createrepo_c --baseurl file://{context.dnf.installroot}/xml_base/dnf-ci-fedora /{context.dnf.repos[dnf-ci-fedora].path}"
+  And I copy directory "{context.dnf.repos[dnf-ci-fedora].path}" to "/xml_base/dnf-ci-fedora"
+ When I execute microdnf with args "install setup"
```
lukash: You are repeatedly installing the same package, "setup". It is important to note that since the microdnf tests don't run in an installroot, and since all tests from a single feature file are run in the same container, you are modifying an environment shared between the tests, therefore the `setup` package is already installed in this test (unless you specifically run this single scenario, in which case it won't be installed...).

This is actually seemingly subtle, but quite a big problem. It can hit you in ways you don't expect, it's hard to write tests in such environment and they are much less robust. @m-blaha is looking into how to run each `@no_installroot` scenario in a different container (even though it will have a performance impact) I think, not sure what's the status? @m-blaha will you have a solution soon or do we still need to worry about this?

Anyway, I think if you want to keep these tests in a single feature file, you should pick a different package to install in each scenario. That way they _shouldn't_ be influencing each other (much). An alternative is to split the scenarios into more feature files, that way they will be run in separate containers. In an extreme you could put one scenario per feature file to ensure they don't influence each other.

dnf-behave-tests/features/microdnf/install-xml-base.feature
```diff
@@ -0,0 +1,73 @@
...
+@bz1734350
+Scenario: Install packages from remote repodata with xml:base pointing to packages on different remote
+#6. remote repo with remote packages (different package location (different url) specified using xml:base)
+Given I make packages from repository "dnf-ci-fedora" accessible via http
+  And I copy repository "dnf-ci-fedora" for modification
+  And I execute "createrepo_c --baseurl http://localhost:{context.dnf.ports[dnf-ci-fedora]} /{context.dnf.repos[dnf-ci-fedora].path}"
+  And I use repository "dnf-ci-fedora" as http
+ When I execute microdnf with args "install setup"
+ Then the exit code is 0
+  And stderr is empty
```
lukash: Since you are writing quite extensive tests for the microdnf install command, may I ask you if you could add a test for the `--nodocs` option? I need the tests for a bug of mine - the simple install test you're adding is actually enough, but it's better to add a test for the option as well just to make sure it's covered...

lucab85: OK https://bugzilla.redhat.com/show_bug.cgi?id=1769831

dnf-behave-tests/features/microdnf/install.feature
```diff
@@ -0,0 +1,86 @@
+@no_installroot
+Feature: microdnf install command on packages with xml:base
```
lukash: This still mentions `xml:base`, while the tests are for installing in general.

dnf-behave-tests/features/microdnf/install.feature
```diff
@@ -0,0 +1,86 @@
...
+  Given I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+@bz1734350
+Scenario: Install package from local repodata with local packages
+#1. local repo with local packages
+Given I use repository "dnf-ci-fedora"
+ When I execute microdnf with args "install kernel"
+ Then the exit code is 0
+  And stderr is empty
```
lukash: Hmm, we should _definitely_ check the installation happened and the output of the command as well (in all the tests). There's a step for dnf to do that: `Transaction is following`, which checks both the dnf output and the rpm transaction that took place. We'll probably need to add a step like `microdnf transaction is following`, which will share the rpm transaction bit, but will parse microdnf output to verify it. Let me know if you can do that or you need some help with it.

dnf-behave-tests/features/microdnf/install.feature
```diff
@@ -0,0 +1,86 @@
...
+
+
+@bz1769831
+Scenario: Install package with option from local repodata with local packages
+Given I use repository "dnf-ci-fedora"
+ When I execute microdnf with args "--setopt=tsflags=nodocs install glibc"
```
lukash: Thanks for this. Could you add another one with the more straightforward `--nodocs` and tag it with `@bz1771012`?

Also (sorry about adding more work), you should check the option did what it was meant to, which, if I understand correctly, doesn't install documentation files from RPMs. Which I think means you need an RPM with a file in `/usr/share/man/` and check it wasn't installed if this option was used.

dnf-behave-tests/features/microdnf/install.feature
```diff
@@ -0,0 +1,86 @@
+@no_installroot
+Feature: microdnf install command on packages with xml:base
+
+
+Background: Using repositories dnf-ci-fedora and dnf-ci-thirdparty-updates
+  Given I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+@bz1734350
```
lukash: Can you add `@bz1779757` here? It's the same crash as the `--nodocs` ones, just this one wasn't reported wtih the `--nodocs` option, so let's tag the simple test with it... Thanks!

dnf-behave-tests/features/microdnf/install.feature
lucab85: I agree and I need support to update "transaction.py" to match microdnf transactions.

Example:
```
[...]
    And Transaction is following                                         # features/steps/transaction.py:144 0.133s
      | Action  | Package                                  |
      | install | kernel-core-0:4.18.16-300.fc29.x86_64    |
      | install | kernel-modules-0:4.18.16-300.fc29.x86_64 |
      | install | kernel-0:4.18.16-300.fc29.x86_64         |
        File "/usr/lib/python3.7/site-packages/behave/model.py", line 1329, in run
          match.run(runner.context)
        File "/usr/lib/python3.7/site-packages/behave/matchers.py", line 98, in run
          self.func(context, *args, **kwargs)
        File "features/steps/transaction.py", line 146, in then_Transaction_contains
          check_transaction(context, 'contains')
        File "features/steps/transaction.py", line 126, in check_transaction
          check_dnf_transaction(context, mode)
        File "features/steps/transaction.py", line 94, in check_dnf_transaction
          dnf_transaction = parse_transaction_table(lines)
        File "/opt/behave/features/steps/lib/dnf.py", line 119, in parse_transaction_table
          table_begin = find_transaction_table_begin(lines)
        File "/opt/behave/features/steps/lib/dnf.py", line 93, in find_transaction_table_begin
          raise RuntimeError("Transaction table start not found")
      RuntimeError: Transaction table start not found

Last Command: microdnf install kernel

Last Command stdout:
Package                                 Repository       Size
Installing:                                                  
 kernel-4.18.16-300.fc29.x86_64         dnf-ci-fedora  6.2 kB
 kernel-core-4.18.16-300.fc29.x86_64    dnf-ci-fedora 52.3 kB
 kernel-modules-4.18.16-300.fc29.x86_64 dnf-ci-fedora 48.2 kB
Transaction Summary:
 Installing:        3 packages
 Reinstalling:      0 packages
 Upgrading:         0 packages
 Removing:          0 packages
 Downgrading:       0 packages
Running transaction test...
Installing: kernel-core;4.18.16-300.fc29;x86_64;dnf-ci-fedora
Installing: kernel-modules;4.18.16-300.fc29;x86_64;dnf-ci-fedora
Installing: kernel;4.18.16-300.fc29;x86_64;dnf-ci-fedora
Complete.

```

lukash:
I've created a PR for the microdnf transaction check step: https://github.com/rpm-software-management/ci-dnf-stack/pull/731

@lucab85 please check it out, I've tried your scenarios on it and they're not doing very well :slightly_smiling_face:  here's your branch with that, the update is simple so you may not find it useful but anyway: https://github.com/lukash/ci-dnf-stack/tree/microdnf-transaction-727

Hope the step works for you, hit me up in case of problems, there might be a bug or two :slightly_smiling_face:

dnf-behave-tests/features/microdnf/install2.feature
```diff
@@ -0,0 +1,19 @@
+@no_installroot
```
lukash: I know it's not easy since you're forced to split the feature file, but the names `install[1-8].feature` aren't great. They don't really tell what is inside. e.g. you've got a test for `--setopt=tsflags=nodocs` here in `install2.feaure` and `--nodocs` in `install8.feature`, while they're obviously related.

We should at least try to make it a bit less messy... come up with some better naming scheme?

(unless @m-blaha has a positive update on running the scenarios in separate containers, which would save us from this?)

dnf-behave-tests/features/microdnf/install1.feature
```diff
@@ -0,0 +1,22 @@
...
+#1. local repo with local packages
+Given I use repository "dnf-ci-fedora"
+ When I execute microdnf with args "install kernel"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | kernel-core-0:4.18.16-300.fc29.x86_64     |
+      | install       | kernel-modules-0:4.18.16-300.fc29.x86_64  |
+      | install       | kernel-0:4.18.16-300.fc29.x86_64          |
+
```
lukash: You now have one too many newline at the ends of the files... :) If configured properly, your editor should add a single newline automatically, as it is customary on linux.

dnf-behave-tests/features/microdnf/install3.feature
```diff
@@ -0,0 +1,25 @@
+@no_installroot
+Feature: microdnf install command on packages
+
+
+Background:
+  Given I delete file "/etc/dnf/dnf.conf"
+    And I delete file "/etc/yum.repos.d/*.repo" with globs
+
+
+
```
lukash: Three newlines here and in some other feature files. Two is enough...

dnf-behave-tests/fixtures/specs/dnf-ci-fedora/man-pages-4.16-3.fc29.spec
```diff
@@ -0,0 +1,23 @@
+Name: man-pages
```
lukash: Please don't add packages to `dnf-ci-fedora`. We were just discussing this - the repository is used for a large number of tests, which get influenced if you change its data (however improbably it is that this will silently break something, the chance is there). The solution should be to create your own repository (which we've agreed should be named e.g. `microdnf-install-nodocs` for this test, and the feature files could then be named like `microdnf/install-nodocs-1.feature` and `microdnf/install-nodocs-2.feature` in this case).

dnf-behave-tests/features/microdnf/install8.feature
```diff
@@ -0,0 +1,19 @@
...
+
+@bz1771012
+Scenario: Install package with --nodocs option from local repodata with local packages
+Given I use repository "dnf-ci-fedora"
+  And I execute microdnf with args "--nodocs install man-pages"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | man-pages-0:4.16-3.fc29.x86_64            |
+  And file "/usr/share/doc/man-pages/README" does not exist
```
lukash: I think it would be a good idea to install the package also without the flag and check the file is there, because there's a solid chance the file won't be there by a mistake in the scenario, like a typo...

dnf-behave-tests/features/microdnf/install-nodocs-1.feature
```diff
@@ -0,0 +1,18 @@
...
+
+
+Scenario: Install a documentation package from local repodata
+Given I use repository "microdnf-install-nodocs"
+ When I execute microdnf with args "install man-pages"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | man-pages-0:4.16-3.fc29.x86_64            |
+  And file "/usr/share/doc/man-pages/README" exist
```
lukash: This doesn't work, it needs to be `exists`, not `exist`.

dnf-behave-tests/features/microdnf/install-nodocs-1.feature
```diff
@@ -0,0 +1,18 @@
...
+
```
lukash: Still empty lines at ends of some of the feature files (sorry about nitpicking :slightly_smiling_face:)

dnf-behave-tests/features/microdnf/install-nodocs-1.feature
```diff
@@ -0,0 +1,18 @@
+@no_installroot
+Feature: microdnf install command on packages
+
+
+Background:
+  Given I delete file "/etc/dnf/dnf.conf"
+    And I delete file "/etc/yum.repos.d/*.repo" with globs
```
lukash: All the Backgrounds have the steps indented by two more spaces than the scenarios, can you unify it?

dnf-behave-tests/features/microdnf/install2.feature
```diff
@@ -0,0 +1,23 @@
...
+  And I use repository "dnf-ci-fedora"
+  And I execute "createrepo_c --baseurl file://{context.dnf.installroot}/xml_base/dnf-ci-fedora /{context.dnf.repos[dnf-ci-fedora].path}"
+  And I copy directory "{context.dnf.repos[dnf-ci-fedora].path}" to "/xml_base/dnf-ci-fedora"
+ When I execute microdnf with args "install abcde"
+ Then the exit code is 0
+  And microdnf transaction is
+      | Action        | Package                                   |
+      | install       | flac-0:1.3.2-8.fc29.x86_64                |
+      | install       | abcde-0:2.9.2-1.fc29.noarch               |
+  And file "/xml_base/dnf-ci-fedora/noarch/abcde-0:2.9.2-1.fc29.noarch" exists
```
lukash: This fails for me, does it pass for you?

dnf-behave-tests/features/microdnf/install3.feature
lukash: Still the three newlines... :slightly_smiling_face: Here and in `install[4-6].feature`.

dnf-behave-tests/features/microdnf/install2.feature
lucab85: I fixed it

PR #728, opened 2019-12-11
Add test for correct detection of the latest

j-mracek:


pkratoch:
Hi, the changes to "dnf-ci-multicontext-hybrid-multiversion-modular" repository are breaking a test in plugins-core/reposync.feature. Can you please fix that?

Btw, we have discussed that it would be best to create special repo for each feature file, so that changes don't affect other tests. It's also more readable. For example, in your scenario, you are enabling postgresql:9.6, but it doesn't have anything to do with the test, it's just a setup, and it makes it more confusing to read. Separate repository with only the right modules and RPMs would remove this problem.

pkratoch:
I can't reproduce the failure anymore. I probably run the tests against wrong RPMs or something. Sorry about the false alarm and I will merge this.

PR #731, opened 2019-12-12
Add a step for checking microdnf transaction

lukash:
Tested somewhat rudimentarily, but it should work at least for the basic cases for which we need it now.

I'd like to wait for @lucab85 to use this in his #727 before merging it, that should test this a bit more - hence the `blocked` tag.


lukash:
The step works for @lucab85, so it can be merged.

PR #738, opened 2019-12-18
Test aliases backslash processing (RhBug:1680482)

pkratoch:


dnf-behave-tests/features/alias-command.feature
```diff
@@ -228,3 +228,16 @@ Scenario: One disabled config does not affect others
         | Action        | Package                               |
         | install       | dnf-ci-package-0:1.0-1.x86_64         |
   Given I successfully execute dnf with args "remove dnf-ci-package"
+
+
+@bz1680482
+Scenario: Backslash ends the recursive processing and the '\' is stripped
+   When I execute dnf with args "alias add install='\install'"
+   Then the exit code is 0
```
lukash: You could use:
```gherkin
   Given I successfully execute dnf with args "alias add install='\install'"
```
Instead of the two lines above.

dnf-behave-tests/features/alias-command.feature
```diff
@@ -228,3 +228,16 @@ Scenario: One disabled config does not affect others
...
+  Given I use repository "alias-command"
+   When I execute dnf with args "install dnf-ci-packageA"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                               |
+        | install       | dnf-ci-packageA-0:1.0-1.x86_64        |
+  Given I successfully execute dnf with args "remove dnf-ci-packageA"
```
lukash: What's this line here for? A `Given` line at the end of a test doesn't make sense? (I see it's also in the test above.. maybe a forgotten leftover and then a copy-paste? :slightly_smiling_face: )

dnf-behave-tests/features/alias-command.feature
pkratoch: Fixed.

dnf-behave-tests/features/alias-command.feature
pkratoch: Because of ``@no_installroot`` tag. The features run in separate containers, but not individual scenarios. Looking at microdnf tests, there it was resolved by splitting the scenarios into multiple features (install1, install2...) or by removing the package in the Background (install_weak_deps.feature). I suppose I could either move the remove step to Background or try to come up with testing the aliases on some command that does not alter the system.

lukash: Oh, right.

So I'd prefer testing on a command that doesn't alter the system. Adding extra steps to work around the issue clutters the tests while we should really be solving the root cause. (and after solving the root cause the clutter will stay, unless someone goes through the scenarios carefully)

pkratoch: Ok, I changed that for all scenarios. I still use install command, because it outputs which package cannot be found and that's handy in the scenarios for recursion to check if original arguments were really used. (For example, info command just outputs "Error: No matching Packages to list".)

PR #741, opened 2019-12-19
Test that sack is not leaking file descriptors

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1594016
https://bugzilla.redhat.com/show_bug.cgi?id=1781360


lukash:
Tbh. I think this should go in unit tests. We don't have the right setup in the CI stack to test the API like this (which is obvious when you look at the new `I execute python script` test and then the scenario being just that step and an exit code check).

Are unit tests officially part of gating? If not, they just simply should be. They are part of the packaging process AFAIK, you can't build a package without running them?

If we need this in, we can make it an exception, but really we should work towards having the API tests in the right place and certainly not mix them with the dnf CLI tests...

PR #750, opened 2020-01-03
Change microdnf install tests to not depend on wget

lucab85:
`wget` can happen to be installed in the container already when the tests are run, which breaks checking the transaction


lukash:
The commit message doesn't say much, why not describe why you're actually making the change? Also the linked bug seems irrelevant...

PR #757, opened 2020-01-09
Add additional createrepo_c tests

kontura:
I have a bunch more createrepo_c tests locally but I will be creating smaller PRs with them to make the review process easier and also because I have to integrate them into the current form of our CI which takes time. 

Tests for:
https://bugzilla.redhat.com/show_bug.cgi?id=1762697
and
https://github.com/rpm-software-management/createrepo_c/issues/130


dnf-behave-tests/common/output.py
```diff
@@ -140,6 +140,6 @@ def then_stderr_is(context):
 
 @behave.then("stderr matches line by line")
 def then_stderr_matches_line_by_line(context):
-    out_lines = context.cmd_stderr.split('\n')
-    regexp_lines = context.text.split('\n')
+    out_lines = context.cmd_stderr.strip().split('\n')
+    regexp_lines = context.text.strip().split('\n')
```
lukash: tbh. I'm a bit torn on the `strip()` here. Ultimately we'd like to check for the correct number of line breaks too. Once we put this in, we probably won't look back... but yeah, we're stripping the output for `stdout` already...

I assume you've had a particular reason to do this?

dnf-behave-tests/createrepo_c/update-update-md-path.feature
```diff
@@ -0,0 +1,122 @@
...
+  And the exit code is 0
+  And stdout matches line by line
+      """
+      Directory walk started
+      Directory walk done - 1 packages
+      Temporary output repo path: ./.repodata/
+      Preparing sqlite DBs
+      Pool started \(with 5 workers\)
+      Pool finished
+
```
lukash: ... You've got this extra line here... And there's a `strip()` on both `expected` and `found` in the step, so this happens to work. But seems like we're too loose, should we do something about it?

kontura: I see, I have the extra line in all the stdouts... I agree that the `expected` should be without the strip, that just allows us to be more sloppy.

dnf-behave-tests/common/output.py
kontura: I also think we probably should not strip either the stdout or stderrr. I did this because I was getting the output from stderr surrounded by two empty ones, but now that I think about it I just treated the symptoms.. I will further look into it.

kontura: It turned out the empty new line really is in the output of createrepo_c, thanks for catching this.

kontura:
I have kind of completely reversed the `strip` commit, now I have removed the `strip` calls from the `stdout matches line by line` step. I ran the suite with this change and all the tests have passed so it should be fine.

lukash:
> I have kind of completely reversed the `strip` commit, now I have removed the `strip` calls from the `stdout matches line by line` step. I ran the suite with this change and all the tests have passed so it should be fine.

Neat, thanks! I'm rerunning the tests to double-check, then I'll merge.

PR #758, opened 2020-01-10
Test modifyrepo_c can work with metadata which is symlink

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1776399


m-blaha:
Thanks for review. Moved to its own feature file (though I wanted it to have more generic name so other potential future modifyrepo_c tests could be added here), symlink step used.

kontura:
Great, thank you.

PR #760, opened 2020-01-13
Avoid interference with packages installed in container

m-blaha:
The wget package may or may not be upgradable in the testing
container.
(See failing test for PR
https://github.com/rpm-software-management/dnf/pull/1566)


kontura:
This passes locally and it should fix failing CI, thank you.

PR #761, opened 2020-01-14
Remove test data in separate image layer

pkratoch:
The ssl certificates are sometimes incorrectly generated when old
fixtures/certificates/testcerts is present. This is probably caused by
a bug in podman that cache is sometimes reused even when files have
changed (with --no-cache option, the ssl certificates are generated
correctly).

Deleting the directory in a separate RUN command (therefore, in a
separate layer) works around the issue.

### DIFF
--- a/Dockerfile.f29
+++ b/Dockerfile.f29
@@ -112,6 +112,11 @@ RUN echo -e "\
 destructive=yes\n\
 os=$OSVERSION" > /opt/behave/behave.ini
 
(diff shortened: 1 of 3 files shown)


m-blaha:
Thanks!

lukash:
Oh please, can you also add a comment explaining why the extra RUN is there? This may cause someone wondering about the why and potentially remove it, thinking it's not doing anything?

PR #765, opened 2020-01-15
Fix wrong path for filesystem package in dnf-ci-fedora repo

lucab85:
- correct feature path prevent input/output error
- variable support in "I execute in dnf shell {command}" step

### DIFF
--- a/dnf-behave-tests/features/shell-install.feature
+++ b/dnf-behave-tests/features/shell-install.feature
@@ -42,7 +42,7 @@ Scenario: Using dnf shell, fail to install local file when goal is not empty
   Given I use repository "dnf-ci-fedora"
    When I open dnf shell session
     And I execute in dnf shell "install setup-0:2.12.1-1.fc29.noarch"
-   When I execute in dnf shell "install {context.dnf.fixturesdir}/repos/dnf-ci-fedora/noarch/filesystem-0:3.9-2.fc29.x86_64.rpm"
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/features/shell-install.feature
```diff
@@ -42,7 +42,7 @@ Scenario: Using dnf shell, fail to install local file when goal is not empty
   Given I use repository "dnf-ci-fedora"
    When I open dnf shell session
     And I execute in dnf shell "install setup-0:2.12.1-1.fc29.noarch"
-   When I execute in dnf shell "install {context.dnf.fixturesdir}/repos/dnf-ci-fedora/noarch/filesystem-0:3.9-2.fc29.x86_64.rpm"
+   When I execute in dnf shell "install {context.dnf.fixturesdir}/repos/dnf-ci-fedora/x86_64/filesystem-3.9-2.fc29.x86_64.rpm"
```
m-blaha: Thanks for the catch. But to fix the issue you also have to patch "@behave.step("I execute in dnf shell \"{command}\"") step in features/steps/shell.py file.
Currently it contains the line:

context.shell_session.sendline(command)

which means that the command string is passed to the shell without any change.

It should be changed to something like
context.shell_session.sendline(command.format(context=context))
which will first replace {context.dnf.fixturesdir} with the real path and then pass the result to the shell.

The test will still keep failing on dnf without https://github.com/rpm-software-management/dnf/pull/1546 patch, but now it will be failing the "right" way :)

lucab85: Done

PR #768, opened 2020-01-16
Rewrite dnf-testing.sh to Python

m-blaha:
It should be as compatible as possible with the bash version. Before merging must be approved by QE that it does not break their workflow.
The main new feature is ability to run destructive scenarios in separate containers and added script to run createrepo_c tests.


dnf-testing.py
```diff
@@ -0,0 +1,335 @@
+
```
lukash: I wouldn't mind having a shebang here. Is python 2 vs python 3 a problem on fedora, can't use a single shebang?

I don't like wrappers around wrappers...

Also, how about naming this better? It runs the tests in a container... I can't think of anything short and punchy, `run-tests-in-container` isn't that great :slightly_smiling_face:

dnf-testing.py
```diff
@@ -9,10 +9,12 @@
...
 
-DESTRUCTIVE_TAGS = set([
-    "@destructive",
-    "@no_installroot",
-])
+PROGPATH = os.path.abspath(os.path.dirname(sys.argv[0]))
+
+# add behave tests root to python path so the `consts` module can be imported
+sys.path.append(os.path.join(PROGPATH, 'dnf-behave-tests'))
+import consts
```
lukash: I'm not sure about adding this much python module hackery... I kind of considered the one in `dnf-behave-tests/common` a necessary evil. Maybe this won't cause issues but maybe we'd be better off just duplicating the two tags here?

dnf-testing.sh
```diff
@@ -9,4 +15,11 @@ if [ ! $PYTHON ]; then
     exit 1
 fi
 
-exec "$PYTHON" dnf-testing.py "$@"
+FEATURESDIR="features"
+if [[ $PROG_NAME == "createrepo_c-testing.sh" ]]; then
+    FEATURESDIR="createrepo_c"
+fi
+
+exec "$PYTHON" dnf-testing.py --featuresdir=$FEATURESDIR "$@"
```
lukash: This is kind of complex and again, wrappers wrappers. Seems unnecessary to me, why not just have (using the current script name even though I suggested finding a better one):
```
# dnf-testing.py dnf ...
# dnf-testing.py createrepo_c ...
```

dnf-testing.py
```diff
@@ -1,13 +1,12 @@
+#!/usr/bin/env python
```
lukash: Why have the refactors as a separate commit in the PR and not just squash them?

m-blaha:
> Just browsing through for now.
> 
> Do you have any (rough) numbers of the overhead of running each scenario in a separate container?

Yes, at first I tried to run tho whole suite one scenario per container but it didn't even finish. If I recall it correctly podman started to fail on creating lock files.
But I did some measurements on smaller portion of tests and running scenario per container was roughly 4 times slower then feature per container. Of course it strongly depends on the number of scenarios in feature files.

dnf-testing.py
m-blaha: Not only on Fedora. The script is run in various environments - Fedora, RHEL7, RHEL8 with different versions of python installed. That is why single shebang is not usable and one of the reasons the shell wrapper exists.

m-blaha: But I will think about some cooler name for the script.

dnf-testing.py
m-blaha: I tried to avoid duplicity here (although in first versions the DESTRUCTIVE_TAGS was duplicated here). Having tests correctly recognized as destructive is crucial and I just wanted to narrow the space for an error by forgetting to edit it in both places.

dnf-testing.py
m-blaha: Yeah, that could be cleaned a bit...

lukash:
> running scenario per container was roughly 4 times slower then feature per container

Oh god, that's terrible... :frowning:

m-blaha:
> > running scenario per container was roughly 4 times slower then feature per container
> 
> Oh god, that's terrible...

Just to make myself clear - the "4 times slower" is not the case for this PR. It was true if all the scenarios were run each in its own container. This PR performs the same as original dnf-testing.sh because it runs only destructive scenarios in separate containers. All others are run the same old way - one feature file per container.

lukash:
Ok, I see... so the impact is not so big now, but it will still be a major factor if we want to run more non-installroot tests, which we should.

dnf-testing.py
```diff
@@ -0,0 +1,351 @@
...
+
+class MultilineHelpFormatter(argparse.HelpFormatter):
+    '''
+    Help formatter that allows multiline help texts.
+    '''
+    def _split_lines(self, text, width):
+        if '\n' in text:
+            return text.splitlines()
+        return super(MultilineHelpFormatter, self)._split_lines(text, width)
+
```
lukash: I suspect sider is not picking up this as Python code to check, is it? I'd guess there should be two linebreaks here, also the "hacked" import of consts.py in that other commit should raise a warning... Do you know how to make it pick this file up?

dnf-testing.py
```diff
@@ -0,0 +1,351 @@
...
+        help="Specified Image ID or name if do not want to run the last built image")
+    parser.add_argument(
+        "-d", "--devel", action="store_true", default=False,
+        help="Share local feature/ with docker")
+    parser.add_argument(
+        "-f", "--file", metavar="FILE", dest="docker_file",
+        default="Dockerfile", help="Path to Dockerfile to use")
+    parser.add_argument(
+        "-p", "--podman", action="store_true", default=False,
+        help="Force using podman instead of docker")
```
lukash: I wonder if we'd like to make podman the default. Granted we've had some issues lately with podman being broken, but otherwise, for development, maybe most of us would prefer to use podman as the default and only fall back to docker if podman isn't present? For CI this shouldn't make a difference?

dnf-testing.py
```diff
@@ -0,0 +1,351 @@
...
+def command_line_parser():
+    '''
+    Initialize and return command line arguments parser.
+    '''
+    parser = argparse.ArgumentParser(
+        description="Functional tests for DNF",
+        formatter_class=MultilineHelpFormatter)
+    parser.add_argument(
+        "--featuresdir", default="features",
+        help="Directory with *.feature files")
```
lukash: So... how about calling this option `"-s", "--suite"`, as in the test suite to run, either `dnf` or `createrepo_c`? I'd even argue it should have no default and be mandatory to specify which suite to run - they are in effect equal, even though `dnf` is what we want to run most of the time. (I'm not opposed to the default if others want to have it though)

I'd also move `dnf-behave-tests/features` to `dnf-behave-tests/dnf`, not necessarily as part of this PR, but I think it will make for a more clear naming. I don't think it should break anything?

dnf-testing.sh
```diff
@@ -1,196 +1,12 @@
-#!/bin/bash
-set -euo pipefail
+#!/usr/bin/bash
```
lukash: Hey, this actually doesn't work for me :slightly_smiling_face: I have `/bin/bash` but no `/usr/bin/bash`...

dnf-testing.py
```diff
@@ -0,0 +1,351 @@
...
+    @property
+    def docker_bin(self):
+        '''
+        Autodetection whether podman or docker is gonna be used
+        '''
+        if self._docker_bin is None:
+            if self.command_line_args.podman:
+                self._docker_bin = 'podman'
+            else:
+                if execute(['command', '-v', 'docker'])[0] \
```
lukash: Does this work for you? `command` is a bash built-in and you need to call `subprocess.Popen(..., shell=True)` for this to work. If I run this script with `-p` (telling it to use podman), it works, without `-p` I get a traceback on this. After fixing this one I'm getting another traceback when trying to run with docker, please test it and if it works for you, we can discuss privately so that we don't spam here...

(my podman is currently acting up, I can only test with docker for now)

dnf-behave-tests/launch-test
```diff
@@ -12,6 +13,7 @@ while [[ $# -gt 0 ]]; do
     case "$1" in
         -r|-R) RESERVE="$1"; shift;;
         -t|--tag|--tags) TAGS="$TAGS --tags $2"; shift 2;;
+        -n) NAMES+=("-n" "$2"); shift 2;;
```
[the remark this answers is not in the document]
m-blaha: Yes, you can pass -n switch multiple times: -n "^The first name$" -n "^The other name$"

dnf-behave-tests/consts.py
```diff
@@ -0,0 +1,18 @@
...
+
+import os
+
+FIXTURES_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "fixtures"))
+
+DEFAULT_DNF_COMMAND = "dnf"
+DEFAULT_CONFIG = os.path.join(FIXTURES_DIR, "dnf.conf")
+DEFAULT_REPOS_LOCATION = os.path.join(FIXTURES_DIR, "repos")
+DEFAULT_RELEASEVER="29"
+DEFAULT_PLATFORM_ID="platform:f29"
```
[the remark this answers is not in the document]
m-blaha: OK, I left only FIXTURES_DIR and DESTRUCTIVE_TAGS in shared consts module. Also the occurrence in createrepo_c/environment.py is now replaced.

dnf-testing.py
m-blaha: Done.

dnf-testing.py
```diff
@@ -0,0 +1,351 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
m-blaha: Done.

dnf-testing.py
m-blaha: The --featuresdir is renamed to --suite. Also moving features/ to dnf/ makes sense, but I'd rather do this in another PR.

dnf-testing.py
m-blaha: Right, I changed the logic so that the podman is the primary choice and docker is a fallback.

dnf-testing.py
m-blaha: Yes, it does. But after some digging on internet I found this: https://unix.stackexchange.com/questions/429400/command-command-not-found-in-python-subprocess
Is this the traceback you are getting? I tried to fix.

dnf-testing.sh
m-blaha: Given that there is /bin -> /usr/bin symlink on Fedora/RHEL I changed the shebang to /bin/bash.

dnf-testing.py
lukash: I don't remember any more, but it's working for me now.

lukash: Actually, with `--docker`, it doesn't:
```
  File "./dnf-testing.py", line 369, in <module>
    BehaveRunner()()
  File "./dnf-testing.py", line 134, in __call__
    command()
  File "./dnf-testing.py", line 363, in command_build
    returncode = subprocess.call(command)
  File "/usr/lib/python3.7/subprocess.py", line 339, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib/python3.7/subprocess.py", line 800, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.7/subprocess.py", line 1551, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'sudo docker': 'sudo docker'
```

The issue is with using `"sudo docker"` as the binary for `subprocess.call()`, if I replace it with two items in the list: `["sudo", "docker", ...]`, it works.

lukash:
@m-blaha can you see my review from today with 4 comments above? I didn't submit it but I restarted my browser and it seems it got posted...

Anyway, I was also thinking about the name of the script(s), what about `podman-test.py`? I'd still drop the createrepo_c specific link (and support for that in the shell script) and just use `podman-test.py -s createrepo_c` instead?

Also I'd still personally squash all the incremental changes to the new script into a single commit, the rest seems like development noise that doesn't need to be part of history, but up to you.

I've gone through the code and tested the script too, I think barring my current notes it's good to merge.

m-blaha:
@lukash yes I can see them. I'm about to look into them.

m-blaha:
@lukash hopefully all issues you mentioned are fixed. When working on it I found another issues and got stucked on fixing them. So sorry for the delay.

PR #769, opened 2020-01-16
Add @xfail tag to failing tests

kontura:
On RHEL 8 createrepo_c is compiled without support for zchunk and
modules therefore we have to disable tests that use them.


lukash:
Shouldn't you have used the os-specific tags? You've just turned the tests off completely...

(I know the tags need to be moved to the common module...)

kontura:
I have thought about it, but since it disabled them only in this branch (rhel-8.2.0) I think is fine for now.

Although once we will be running createrepo_c gating with the master branch we might need them for the next releases. But it doesn't seem necessary to have it in this already branched rhel-8.2.0.

lukash:
@kontura ah, right, I haven't noticed this is for the `rhel-8.2.0` branch...

PR #770, opened 2020-01-16
Replace duplicate tests for yum/dnf name with more general parametrized form

kontura:
Updating the matching of "dnf"/"yum" in output as discussed here: https://github.com/rpm-software-management/ci-dnf-stack/pull/710

### DIFF
--- a/dnf-behave-tests/features/environment.py
+++ b/dnf-behave-tests/features/environment.py
@@ -119,6 +119,13 @@ class DNFContext(object):
             self.delete_installroot = True
 
         self.dnf_command = userdata.get("dnf_command", DEFAULT_DNF_COMMAND)
+
+        self.prog_name = "dnf"
+        self.prog_name_upper = "DNF"
+        if "yum" in self.dnf_command.lower():
+            self.prog_name = "yum"
+            self.prog_name_upper = "YUM"
+
         self.config = userdata.get("config", DEFAULT_CONFIG)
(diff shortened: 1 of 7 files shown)


lukash:
Okay, this looks good and thanks for coming back to it.

But as it is, you are removing the yum tests, are we already explicitly running the test suite against yum or does it remain to be implemented?

kontura:
So the original motivation for this was failing gating, it used to fail because it is run with yum on rhel and the outputs expected dnf. The suite can be run with yum using `./dnf-testing run --command yum`.

I agree with your and @m-blaha's comments from #710 that its preferable to be able to run the whole suite with either dnf or yum, rather than to have couple specific scenarios for yum.

lukash:
> So the original motivation for this was failing gating, it used to fail because it is run with yum on rhel and the outputs expected dnf.

Ok, but I seem to recall another reason for the tests was to actually make sure that dnf does say "yum" and not "dnf" in its output if run through the `yum` command. I'm not entirely sure and looking for the commits led me to #710, so maybe I'm misremembering. @j-mracek, you were making the changes, do we need to explicitly test that dnf says "yum" if you use the `yum` symlink or are we good here?

j-mracek:
If there was a test that always used yum prefix it must stay there. YUM prefix is very important for us and functionality must be tested always in upstream.

kontura:
Ok, its no problem, we can keep it as it is.

PR #772, opened 2020-01-17
Basic smoke tests for swidtags plugin (RhBug:1689178)

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1689178

### DIFF
--- a/dnf-behave-tests/features/plugins-third/swidtags.feature
+++ b/dnf-behave-tests/features/plugins-third/swidtags.feature
@@ -1,0 +1,68 @@
+# the swidtags plugin is available only in Fedora
+@use.with_os=fedora__ge__30
+Feature: Smoke test for swidtags dnf plugin
+
+
+Background: Enable the swidtags plugin
+  Given I enable plugin "swidtags"
+
+
+@bz1689178
+Scenario: Run swidtags without command prints usage
+   When I execute dnf with args "swidtags"
+   Then the exit code is 2
+    And stderr is
+    """
+    usage: dnf swidtags [-c [config file]] [-q] [-v] [--version]
+                        [--installroot [path]] [--nodocs] [--noplugins]
+                        [--enableplugin [plugin]] [--disableplugin [plugin]]
+                        [--releasever RELEASEVER] [--setopt SETOPTS]
+                        [--skip-broken] [-h] [--allowerasing] [-b | --nobest] [-C]
+                        [-R [minutes]] [-d [debug level]] [--debugsolver]
+                        [--showduplicates] [-e ERRORLEVEL] [--obsoletes]
+                        [--rpmverbosity [debug level name]] [-y] [--assumeno]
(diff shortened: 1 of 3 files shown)


dnf-behave-tests/features/plugins-third/swidtags.feature
```diff
@@ -0,0 +1,68 @@
...
+                        [--noautoremove] [--nogpgcheck] [--color COLOR]
+                        [--refresh] [-4] [-6] [--destdir DESTDIR] [--downloadonly]
+                        [--comment COMMENT] [--bugfix] [--enhancement]
+                        [--newpackage] [--security] [--advisory ADVISORY]
+                        [--bz BUGZILLA] [--cve CVES]
+                        [--sec-severity {{Critical,Important,Moderate,Low}}]
+                        [--forcearch ARCH]
+                        {{sync,regen,purge}}
+    dnf swidtags: error: the following arguments are required: swidtagscmd
+    """
```
lukash: Is this scenario useful? It will require updates when the common CLI switches are changed or the help text changes, which isn't relevant to this particular test...

dnf-behave-tests/features/plugins-third/swidtags.feature
```diff
@@ -0,0 +1,68 @@
...
+  Given I use repository "dnf-ci-fedora"
+    # enable locally generated SWID tags
+    And I create and substitute file "/etc/dnf/dnf.conf" with
+      """
+      [main]
+      gpgcheck=1
+      installonly_limit=3
+      clean_requirements_on_remove=True
+      pluginconfpath={context.dnf.installroot}/etc/dnf/plugins
+      """
```
lukash: Every time I see this, I recall we should really do the automatic config generation... :slightly_smiling_face:

dnf-behave-tests/features/plugins-third/swidtags.feature
m-blaha: I consider the scenario useful. There was a bug in command line arguments parsing that would have been catched if we had had this test. Although, I can change it to test that only last line of stderr matches.

dnf-behave-tests/features/plugins-third/swidtags.feature
m-blaha: True :)

dnf-behave-tests/features/plugins-third/swidtags.feature
lukash: I am torn here. On one side I imagine having this test for every plugin and then changing something in the `--help` output... we don't want the redundancy. On the other hand, if you only test for the last line, there can be undesired text in the output which it won't catch.

So.. keep it the way it is? :slightly_smiling_face:

m-blaha: The same here. Anyway, it is the only test in the suite that checks full output of the --help switch.

PR #773, opened 2020-01-20
Add another part of createrepo_c tests

kontura:
Adding tests for modifyrepo_c with compressed input, zchunk and drpms.

I have also extracted the VersionedActiveTagMatcher class into common, so that createrepo_c can use it (its useful mainly for zchunk and modules, because those are turned off on rhel)

### DIFF
--- a/dnf-behave-tests/createrepo_c/environment.py
+++ b/dnf-behave-tests/createrepo_c/environment.py
@@ -17,6 +17,7 @@ from behave import model
 from behave.formatter.ansi_escapes import escapes
 
 from common.lib.cmd import print_last_command
+from common.lib.tag_matcher import VersionedActiveTagMatcher
 
 FIXTURES_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "fixtures"))
 
@@ -58,6 +59,8 @@ def after_step(context, step):
 
 
 def before_scenario(context, scenario):
+    if context.tag_matcher.should_exclude_with(scenario.effective_tags):
+        scenario.skip(reason="DISABLED ACTIVE-TAG")
     context.tempdir_manager = TempDirManager(context.config.userdata)
 
     context.scenario.default_tmp_dir = context.tempdir_manager.tempdir
@@ -85,7 +88,7 @@ def after_tag(context, tag):
 
 
 def before_all(context):
-    pass
+    context.tag_matcher = VersionedActiveTagMatcher({"os": context.config.userdata.get("os", None)})
 
 
 def after_all(context):
--- a/dnf-behave-tests/createrepo_c/steps/lib/file.py
+++ b/dnf-behave-tests/createrepo_c/steps/lib/file.py
@@ -27,14 +27,8 @@ def get_checksum_regex(type_str):
 
 
 def get_compression_suffix(type_str):
-    if type_str == "gz":
-        return ".gz"
-    if type_str == "zck":
-        return ".zck"
-    if type_str == "xz":
-        return ".xz"
-    if type_str == "bz2":
-        return ".bz2"
+    if type_str in ("gz", "zck", "xz", "bz2"):
+        return "." + type_str
     if type_str == "-":
         return ""
     raise ValueError("Unknown compression type: " + type_str)
@@ -47,6 +41,9 @@ def decompression_iter(filepath, compression_type, blocksize=65536):
         return file_as_blockiter(lzma.open(filepath, 'rb'), blocksize)
     if compression_type == "bz2":
         return file_as_blockiter(bz2.open(filepath, 'rb'), blocksize)
+    if compression_type == "zck":
+        decompress_file_by_extension_to_dir(filepath, os.path.dirname(filepath))
+        return file_as_blockiter(open(filepath[:-4], 'rb'), blocksize)
 
     return file_as_blockiter(open(filepath, 'rb'), blocksize)
 
@@ -83,10 +80,7 @@ def decompress_file_by_extension_to_dir(compressed_filepath, dest_dir):
 
     basename = os.path.basename(compressed_filepath)
     dst = os.path.join(dest_dir, basename)
-    if compressed_filepath.endswith(".bz2"):
-        dst = dst[:-4]
-    else: # .gz and .xz are both 3 chars long
-        dst = dst[:-3]
+    dst = os.path.splitext(dst)[0] #remove compression extension
 
     open(dst, "wb").write(content)
     return dst
--- a/dnf-behave-tests/createrepo_c/steps/lib/file.py
+++ b/dnf-behave-tests/createrepo_c/steps/lib/file.py
@@ -84,3 +84,21 @@ def decompress_file_by_extension_to_dir(compressed_filepath, dest_dir):
 
     open(dst, "wb").write(content)
     return dst
+
+
+def create_compressed_file_with_contents(filename, compression, contents, encoding="utf-8"):
+    fullname = filename + get_compression_suffix(compression)
+    if os.path.exists(fullname):
+        raise ValueError("File: " + fullname + " already exists")
+
+    if compression == "gz":
+        with gzip.open(fullname, 'wt') as f:
+            f.write(contents)
+    elif compression == "xz":
+        with lzma.open(fullname, 'wt') as f:
+            f.write(contents)
+    elif compression == "bz2":
+        with bz2.open(fullname, 'wt') as f:
+            f.write(contents)
+    else:
+        raise ValueError("Unknown compression type: " + compression)
--- a/dnf-behave-tests/createrepo_c/steps/repodata.py
+++ b/dnf-behave-tests/createrepo_c/steps/repodata.py
@@ -179,9 +179,7 @@ def repodata_in_path_is(context, path):
             if compression_suffix and filepath.endswith(compression_suffix):
                 filepath = filepath[:-(len(compression_suffix))]
             if tmp:
-                if filepath.endswith(".zck"):
-                    assert("ZCK" in str(tmp))
-                elif filepath.endswith(".sqlite"):
+                if filepath.endswith(".sqlite"):
                     assert("SQLite" in str(tmp))
                 elif filepath.endswith(".xml"):
                     assert("xml" in str(tmp))
@@ -194,7 +192,7 @@ def repodata_in_path_is(context, path):
                     raise
         except (AssertionError, IOError):
             raise AssertionError("Cannot decompress File: " + repodata_file + " using"
-                                 " copression type: " + compression_type)
+                                 " compression type: " + compression_type)
 
     if len(files) > 0:
         raise AssertionError("repodata directory contains additional metadata files:\n{0}".format('\n'.join(files)))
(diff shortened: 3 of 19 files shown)


dnf-behave-tests/createrepo_c/environment.py
```diff
@@ -85,7 +91,9 @@ def after_tag(context, tag):
 
 
 def before_all(context):
-    pass
+    os = context.config.userdata.get("os", None)
+    if os:
+        active_tag_value_provider["os"] = os
```
lukash: Couldn't you just do something like:
```python
context.tag_matcher = VersionedActiveTagMatcher({"os": context.config.userdata.get("os", UNDEFINED)})
```
Where `UNDEFINED` is some default value representing an unidentified os. If the tag matcher won't work with this (I haven't checked...), just replace this oneliner with constructing the dict and adding the "os" item only if it's defined...

Then you can just use `context.tag_matcher` in `before_scenario()`.

This is cleaner and saves us the global variables, which don't hurt in this case but it's good practice to avoid global variables if you can.

dnf-behave-tests/createrepo_c/steps/lib/file.py
```diff
@@ -95,3 +95,21 @@ def decompress_file_by_extension_to_dir(compressed_filepath, dest_dir):
...
+        with gzip.open(fullname, 'wt') as f:
+            f.write(contents)
+    elif compression == "xz":
+        with lzma.open(fullname, 'wt') as f:
+            f.write(contents)
+    elif compression == "bz2":
+        with bz2.open(fullname, 'wt') as f:
+            f.write(contents)
+    else:
+        raise ValueError("Unknown compression type: " + compression)
```
lukash: Is there any reason why not put the body of this function straight to the step in `dnf-behave-tests/createrepo_c/steps/cmd.py`? It doesn't seem probable this function will be useful in another place and then the indirection just makes it harder to read the code and has no benefit?

dnf-behave-tests/createrepo_c/steps/lib/file.py
```diff
@@ -83,7 +88,7 @@ def decompress_file_by_extension_to_dir(compressed_filepath, dest_dir):
 
     basename = os.path.basename(compressed_filepath)
     dst = os.path.join(dest_dir, basename)
-    if compressed_filepath.endswith(".bz2"):
+    if compressed_filepath.endswith(".bz2") or compressed_filepath.endswith(".zck"):
         dst = dst[:-4]
     else: # .gz and .xz are both 3 chars long
         dst = dst[:-3]
```
lukash: Seems you could use https://docs.python.org/3/library/os.path.html#os.path.splitext for this...

dnf-behave-tests/createrepo_c/steps/repodata.py
```diff
@@ -176,19 +176,19 @@ def repodata_in_path_is(context, path):
...
             tmp = next(decompression_iter(filepath, compression_type, blocksize=100))
+            if filepath.endswith(compression_suffix):
+                    filepath = filepath[:-(len(compression_suffix))]
             if tmp:
-                if filepath.endswith(".zck"):
-                    assert("ZCK" in str(tmp))
-                elif filepath.endswith(".sqlite"):
-                    assert("SQLITE" in str(tmp))
+                if filepath.endswith(".sqlite"):
+                    assert("SQLite" in str(tmp))
```
lukash: You've changed the case on the "SQLITE" string, is it intentional? ("SQLITE" before, "SQLite" after)

dnf-behave-tests/createrepo_c/steps/lib/repodata.py
```diff
@@ -55,3 +55,32 @@ def verify_repomd_item_with_file(item, filepath, decompressed_filepath):
     if (a < b - 1 or a > b + 1):
         raise AssertionError("Timestamp of File: " + filepath + " doesn't match"
                              "timestamp in the repomd.xml:\n" + str(a) + " vs. " + str(b))
+
+def keys_do_not_differ(prim, flist, oth):
+    if set(prim.keys()) != set(flist.keys()):
+        raise AssertionError("Primary and Filelists have different package sets.")
```
lukash: These assertions will be rather unhelpful when someone starts getting them. One way to do it is include the difference in the assertion text itself. For larger data that are better printed in multiline and with something to highlight the differences to the reader this gets a bit impractical.

How about using my great `print_lines_diff()`? :slightly_smiling_face:
https://github.com/rpm-software-management/ci-dnf-stack/blob/a7aa1a30866ae5ae6a19ec0f4cc1de0a3f2467ee/dnf-behave-tests/common/lib/diff.py#L15

It's a bit makeshift (and I don't know how the data you are comparing look like, so it may not display them well), but it does the job in coloring the diff so it's easier to see what's actually different. You can tweak it too if you want; just make sure you don't break the current usage - because it only gets called when a scenario fails, so you won't know it unless you test it explicitly.

dnf-behave-tests/createrepo_c/steps/lib/repodata.py
```diff
@@ -55,3 +55,32 @@ def verify_repomd_item_with_file(item, filepath, decompressed_filepath):
...
+            raise AssertionError("Primary repodata are different.\n"
+                                 "Difference: %s" % (diff))
+        diff = flist1.diff(flist2)
+        if diff:
+            raise AssertionError("Filelists repodata are different.\n"
+                                 "Difference: %s" % (diff))
+        diff = oth1.diff(oth2)
+        if diff:
+            raise AssertionError("Other repodata are different.\n"
+                                 "Difference: %s" % (diff))
```
lukash: Oh, one more thing, as I mentioned before, I see no benefit in putting functions in a different ("lib") python module if you only use them in one place anyway. I'd prefer to put these at the top of the module you use them in, to save the reader having to go find them in a different place.

dnf-behave-tests/createrepo_c/steps/lib/file.py
kontura: I didn't want to import all the compression libraries again. But I can move it if you don't think is worth it?

dnf-behave-tests/createrepo_c/steps/lib/file.py
kontura: Thanks, that's nifty. :+1:

dnf-behave-tests/createrepo_c/steps/repodata.py
kontura: Yea, there was a bug actually. it always ended in the `else` branch and just passed. :-( Even if the string was wrongly capitalized. I will extract it into a separate commit and fix the dangerous default `pass`.

dnf-behave-tests/createrepo_c/steps/lib/file.py
lukash: Yeah, ok... probably reason enough to leave it this way...

dnf-behave-tests/createrepo_c/steps/lib/repodata.py
kontura: The original lines are basically completely different, so `print_lines_diff` would almost always display everything red. But I could print which packages are missing from which file, that should be easy and potentially helpful.

lukash: Ok.. I have very little idea of how the data you're working with here actually look like, but you could just print a diff of the keys you're comparing for the assert (`set(prim.keys()) != set(flist.keys())`), as that is what really matters and so that's what the person running the tests should see.

You can see I've added support for diffing lists and tuples to the `print_lines_diff` function some time ago, I think it can compare sets just the same, I just didn't add it as I didn't have a use for it.

dnf-behave-tests/createrepo_c/environment.py
kontura: I have updated it this way also in dnf `environment.py`.

kontura:
Thank you for the review @lukash, I have pushed an update which addresses your comments.

dnf-behave-tests/createrepo_c/steps/repodata.py
```diff
@@ -176,16 +176,22 @@ def repodata_in_path_is(context, path):
                                  " not match suffix of File: " + repodata_file)
         try:
             tmp = next(decompression_iter(filepath, compression_type, blocksize=100))
+            if compression_suffix and filepath.endswith(compression_suffix):
+                    filepath = filepath[:-(len(compression_suffix))]
```
lukash: This line is over-indented, right? Weird the pep8 check didn't catch it...

dnf-behave-tests/createrepo_c/steps/lib/file.py
```diff
@@ -35,6 +35,8 @@ def get_compression_suffix(type_str):
         return ".xz"
     if type_str == "bz2":
         return ".bz2"
+    if type_str == "zck":
+        return ".zck"
```
lukash: Hmm, the `type_str == "zck"` check is already present on line 32 :slightly_smiling_face: 

Seems you could just do:
```python
if type_str in ("gz", "zck", "xz", "bz2"):
    return "." + type_str
# ...
```

dnf-behave-tests/createrepo_c/steps/repodata.py
```diff
@@ -32,11 +33,46 @@
       "md_ns": "http://linux.duke.edu/metadata/repo"}
 
 
+def keys_do_not_differ(prim, flist, oth):
+    if set(prim.keys()) != set(flist.keys()):
+        print_lines_diff(prim.keys(), flist.keys())
```
lukash: You're converting the `prim.keys()` to a set for the check itself. I'm not sure if it's "just in case" or actually needed (i.e. there's either a possibility of duplicate values or perhaps the order can differ?), but in either case the diff that is printed will be confusing, because either there will be duplicities or the order will be wrong.

I wouldn't mind doing `print_lines_diff(typle(set(prim.keys())), tuple(set(flist.keys())))` as it doesn't matter much here (though I'm not entirely sure the order will be stable this way), but I think you can just as easily extend `print_lines_diff()` to support sets?

dnf-behave-tests/createrepo_c/steps/repodata.py
```diff
@@ -69,6 +105,12 @@ def repodata_are_consistent(context, path):
             filelists_path = decompressed_p
         elif "other" == item.name:
             other_path = decompressed_p
+        elif "primary_zck" == item.name:
+            prim_zck_path = decompressed_p
+        elif "filelists_zck" == item.name:
+            filelists_zck_path = decompressed_p
+        elif "other_zck" == item.name:
+            other_zck_path = decompressed_p
```
lukash: Hmm, the convention is to have the variable part on the left and constant part on the right when doing comparisons, I find this strangely twisted :slightly_smiling_face: 

The code is already there, so up to you if you wanna flip all the conditions (don't flip just those that are added, that would just make it worse...)

dnf-behave-tests/createrepo_c/steps/repodata.py
kontura: I just removed the set conversion, I don't remember why I put it there since duplicates and or different order would actually indicate a problem. Thanks. :-)

kontura:
I have pushed another update.

lukash:
@kontura thanks and sorry about the delay. There are conflicts now, can you rebase?

The code is good to merge from my side (under the assumption the tests still pass, they did when I tested it two weeks ago :slightly_smiling_face: ), feel free to merge it yourself if I don't come around to do it...

Thanks!

kontura:
I have rebased and checked that the tests still pass.
Thanks for the review, merging.

PR #801, opened 2020-03-03
Add test for check for TraceBack for group remove command

j-mracek:
The issue is triggered when module/group is removed, package is not
remove and only reason is changed to dep but following operation will
remove it anyway.


pkratoch:
Can you please add "@not.with_os=rhel__eq__8" tag to the test?

j-mracek:
@pkratoch added

PR #807, opened 2020-03-17
Repoquery --whatrequires with srpms (buildrequries) (RhBug:1812596)

kontura:
https://bugzilla.redhat.com/show_bug.cgi?id=1812596


kontura:
These tests should also work with older (rhel-8.2.0) implementation of repoquery, therefore no `@not.with_os=rhel__eq__8` is needed.

PR #810, opened 2020-03-20
Add test for creation SWDB in memory for normal user

j-mracek:


lukash:
This one got added in #816.

PR #811, opened 2020-03-23
Add dependency on glibc-all-langpacks to test encoding

pkratoch:
Without this, encoding.feature was failing due to:
"Failed to set locale, defaulting to C"


kontura:
@janblazek I think you will also have to update other QA dockerfiles with this change, at least for 8.3 once the tests in encoding.feature are enabled.

PR #812, opened 2020-03-26
Sort search output (RhBug:1811802)

inknos:
dnf: https://github.com/rpm-software-management/dnf/pull/1610/
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1811802

### DIFF
--- a/dnf-behave-tests/features/search-sort.feature
+++ b/dnf-behave-tests/features/search-sort.feature
@@ -1,0 +1,91 @@
+Feature: Sort search command output
+
+Background:
+  Given I use repository "search-sort"
+
+@bz1811802
+Scenario: sort alphanumerically
+  When I execute dnf with args "search name"
+  Then the exit code is 0
+   And stdout is
+   """
+   ========================== Name Exactly Matched: name ==========================
+   name.src : Summary
+   name.x86_64 : Summary
+   ========================= Name & Summary Matched: name =========================
+   name-summary.src : Summary of name
+   name-summary.x86_64 : Summary of name
+   name-summary-description.src : Summary of name
+   name-summary-description.x86_64 : Summary of name
+   name-summary-description-url.src : Summary of name
+   name-summary-description-url.x86_64 : Summary of name
+   name-summary-url.src : Summary of name
+   name-summary-url.x86_64 : Summary of name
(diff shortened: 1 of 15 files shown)


dnf-behave-tests/features/search-alpha.feature
```diff
@@ -0,0 +1,30 @@
...
+  Then the exit code is 0
+   And stdout is
+   """
+   ======================== Name & Summary Matched: fedora ========================
+   fedora-release.noarch : Fedora release files
+   fedora-release.src : Fedora release files
+   =========================== Summary Matched: fedora ============================
+   basesystem.noarch : The skeleton package which defines a simple Fedora system
+   basesystem.src : The skeleton package which defines a simple Fedora system
+   """
```
lukash: This doesn't seem like it necessarily tests the output is actually sorted...

In case we'd keep the `--alpha` option (see my https://github.com/rpm-software-management/dnf/pull/1610#issuecomment-606622177), the proper way to test it would be to call `search` without `--alpha` that would have a different order than when sorted.

In case we ditch the option, I guess this is fine, though it would be somewhat preferable to have at least two packages listed in one of the groups :slightly_smiling_face:

dnf-behave-tests/features/search-sort.feature
```diff
@@ -0,0 +1,92 @@
+Feature: Sort search command output
+
+Background:
+  Given I use repository "dnf-ci-search"
```
lukash, line 4: Could you name the repository according to the name of the feature file, i.e. `search-sort`? The `dnf-ci-` prefix is legacy for repositories that could conflict with the ones on the host system (although those shouldn't be there for the tests). For custom repositories for a given feature file, the convention is to name them after the feature file name.

dnf-behave-tests/features/search-sort.feature
```diff
@@ -0,0 +1,92 @@
...
+
+@bz1811802
+Scenario: sort alphanumerically
+  When I execute dnf with args "search name"
+  Then the exit code is 0
+   And stdout is
+
```
lukash, line 11: Nitpick: remove the linebreak :slightly_smiling_face:

PR #814, opened 2020-04-03
[needs-restarting] add opt using config file (RhBug:1810123)

inknos:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1810123


dnf-behave-tests/features/plugins-core/needs-restarting-conf-files.feature
```diff
@@ -0,0 +1,122 @@
+Feature: Add package to needs-restarting using config files
+
+Background:
+    Given I enable plugin "needs_restarting"
+      And I use repository "dnf-ci-fedora"
+      And I move the clock backward to "before boot-up"
+      And I execute dnf with args "install lame kernel basesystem glibc wget abcde"
```
lukash, line 7: Looks like this list of packages is unnecessary, you're only ever using wget and abcde from the list?

While not necessary, I'd prefer if you set up a single repository with the 3 packages you're using and used only that in the feature file, to make it a bit cleaner... I'm not requiring it though.

dnf-behave-tests/features/plugins-core/needs-restarting-conf-files.feature
```diff
@@ -0,0 +1,122 @@
...
+      """
+      wget
+      """
+      Given I do not set config file
+
+    @bz1810123
+    Scenario: restarting needed after updating package in conf files
+        Given I execute dnf with args "upgrade wget"
+         When I execute dnf with args "needs-restarting -r"
+         Then the exit code is 1
```
lukash, line 21: Unrelated to your PR, so consider this a rhetorical question :slightly_smiling_face: but is it really returning a non-zero exit status when there are packages to restart? That doesn't make sense...

dnf-behave-tests/features/plugins-core/needs-restarting-conf-files.feature
```diff
@@ -0,0 +1,122 @@
...
+          And stdout is
+          """
+          Core libraries or services have been updated since boot-up:
+            * wget
+
+          Reboot is required to fully utilize these updates.
+          More information: https://access.redhat.com/solutions/27943
+
```
lukash, line 29: Is this empty line in the output?

dnf-behave-tests/fixtures/specs/needs-restarting/test-1.0-1.fc29.spec
```diff
@@ -0,0 +1,21 @@
+Name:           test
+Version:        1.0.0
+Release:        1
```
lukash, line 3: Nitpick: the release actually doesn't contain "fc29", which you have in the filename, you could remove it, but whatever... :slightly_smiling_face:

dnf-behave-tests/features/plugins-core/needs-restarting-conf-files.feature
```diff
@@ -0,0 +1,122 @@
...
+      And I use repository "dnf-ci-fedora-updates"
+    Given I create directory "/etc/dnf/plugins/needs-restarting.d"
+      And I create and substitute file "/etc/dnf/plugins/needs-restarting.d/wget.conf" with
+      """
+      wget
+      """
+      Given I do not set config file
+
+    @bz1810123
+    Scenario: restarting needed after updating package in conf files
```
lukash, line 18: The alignment in the whole file looks quite weird. We don't really have a consistent way to align in feature files, sadly... but this is quite a bit different. I prefer something like:
```
Scenario: test
Given something
  And something else
 When X
 Then foo is
      """
      hello
      """
```

dnf-behave-tests/features/plugins-core/needs-restarting-conf-files.feature
```diff
@@ -0,0 +1,122 @@
...
+      And I move the clock backward to "before boot-up"
+      And I execute dnf with args "install lame kernel basesystem glibc wget abcde"
+      And I move the clock forward to "the present"
+      And I use repository "dnf-ci-fedora-updates"
+    Given I create directory "/etc/dnf/plugins/needs-restarting.d"
+      And I create and substitute file "/etc/dnf/plugins/needs-restarting.d/wget.conf" with
+      """
+      wget
+      """
+      Given I do not set config file
```
lukash, line 15: Almost forgot, please rebase, this step was removed. Not sure why you have it here exactly but I think you just won't need it anymore (the config is now generated inside the installroot automatically).

Also, you're using multiple `Given`s here, you can just use `And` to be more consistent.

dnf-behave-tests/features/plugins-core/needs-restarting-conf-files.feature
```diff
@@ -0,0 +1,142 @@
+Feature: Add package to needs-restarting using config files
+
+Background:
+Given I enable plugin "needs_restarting"
+  And I use repository "needs-restarting"
+  And I move the clock backward to "before boot-up"
+  And I execute dnf with args "install wget abcde"
+  And I move the clock forward to "the present"
+  And I use repository "needs-restarting-updates"
+Given I create directory "/etc/dnf/plugins/needs-restarting.d"
```
lukash, line 10: This `Given` should be `And`.

dnf-behave-tests/features/plugins-core/needs-restarting-conf-files.feature
```diff
@@ -0,0 +1,142 @@
...
+  Reboot is required to fully utilize these updates.
+  More information: https://access.redhat.com/solutions/27943
+  """
+
+@bz1810123
+Scenario: conf files can be more than one
+Given I create and substitute file "/etc/dnf/plugins/needs-restarting.d/abcde.conf" with
+"""
+abcde
+"""
```
lukash, line 35: I'm not really sure at what indent we are putting the text blocks, but this is inconsistent with the other text blocks in this file, please unify it (more occurrences below).

dnf-behave-tests/features/plugins-core/needs-restarting-conf-files.feature
```diff
@@ -0,0 +1,142 @@
...
+  And I execute dnf with args "upgrade wget abcde"
+ Then the exit code is 0
```
lukash, line 37: This actually is a setup step to test the `needs-restarting` functionality. You should use this instead of the two above lines:
```
Given I successfully execute dnf with args "upgrade wget abcde"
```

(Or, in this case `Given` is the active step so you should use `And`)

More occurrences of this below.

dnf-behave-tests/fixtures/specs/needs-restarting-updates/abcde-2.9.3-1.fc29.spec
```diff
@@ -0,0 +1,22 @@
+Name:           abcde
+Version:        2.9.3
```
lukash, line 2: I prefer to use simple versions for the test packages, like `1.0`, so that you don't need to squint when you have multiple versions of a package like 3.14.1.1 and 3.14.7.1 and try to figure out which is which :sunglasses: But feel free to keep it here, not much of an issue in this case.

dnf-behave-tests/fixtures/specs/needs-restarting-updates/abcde-2.9.3-1.fc29.spec
```diff
@@ -0,0 +1,22 @@
...
+Summary:        A Better CD Encoder
+
+# cddb-tool is Public Domain, otherwise GPLv2+
+License:        GPLv2+ and Public Domain
+URL:            https://abcde.einval.com/
+
+BuildArch:      noarch
+Requires:       wget
+Recommends:     flac
+Suggests:       lame
```
lukash, line 13: There are attributes on the packages like the above which are irrelevant to the test and therefore superfluous (also the lengthy `%description`s). Again, not hurting anything and they are not too long, so feel free to keep them but in general it's good practice to keep the test data minimal to the test case.

PR #815, opened 2020-04-08
Move invalid_utf8_char from context.dnf to context

pkratoch:
Although this is used only in dnf tests, the character must be also
replaced when checking context.cmd in case that the dnf command
contained it. Moreover, it is not a dnf-specific thing, so it makes
more sense in the context rather than context.dnf.

### DIFF
--- a/dnf-behave-tests/common/lib/cmd.py
+++ b/dnf-behave-tests/common/lib/cmd.py
@@ -6,6 +6,8 @@ from __future__ import print_function
 from behave.formatter.ansi_escapes import escapes
 import subprocess
 
+import consts
+
 
 def run(cmd, shell=True, cwd=None):
     """
@@ -48,13 +50,13 @@ def run_in_context(context, cmd, can_fail=False, **run_args):
 
 
 def assert_exitcode(context, exitcode):
-    cmd = context.cmd.replace(context.dnf.invalid_utf8_char, "\\udcfd")
+    cmd = context.cmd.replace(consts.INVALID_UTF8_CHAR, "\\udcfd")
     assert context.cmd_exitcode == int(exitcode), \
         "Command has returned exit code {0}: {1}".format(context.cmd_exitcode, cmd)
 
 
 def print_last_command(context):
-    cmd = context.cmd.replace(context.dnf.invalid_utf8_char, "\\udcfd")
+    cmd = context.cmd.replace(consts.INVALID_UTF8_CHAR, "\\udcfd")
     if getattr(context, "cmd", ""):
         print(
             "%sLast Command: %s%s" %
--- a/dnf-behave-tests/consts.py
+++ b/dnf-behave-tests/consts.py
@@ -10,3 +10,5 @@ DESTRUCTIVE_TAGS = [
     "destructive",
     "no_installroot",
 ]
+
+INVALID_UTF8_CHAR = '\udcfd'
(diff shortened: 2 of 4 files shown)


lukash:
`context.dnf` is basically a scenario context... in a DNF test suite - everything there is related to DNF tests, it's not like it should define whether it goes to `context.dnf` or not :slightly_smiling_face: 

We'd probably be better off renaming it just for the sake of renaming, even though it's wide-spread and any name we come up with will be longer... :roll_eyes:

pkratoch:
> `context.dnf` is basically a scenario context... in a DNF test suite - everything there is related to DNF tests, it's not like it should define whether it goes to `context.dnf` or not slightly_smiling_face

Not everything is related to DNF tests. There are also createrepo_c tests that don't have the dnf cotext. I should have mentioned it in the commit message explicitly.

> We'd probably be better off renaming it just for the sake of renaming, even though it's wide-spread and any name we come up with will be longer... roll_eyes

I don't understand this note. What should we rename or not rename?

lukash:
> Not everything is related to DNF tests. There are also createrepo_c tests that don't have the dnf cotext. I should have mentioned it in the commit message explicitly.

Oh, you've added the constant to the `createrepo_c` context too, I've missed that. Not sure I like it that much (as the steps in `common` are now dependent on the same, duplicate variable being present in two places). I guess it's not a big problem, but there should probably be a better place to put it. Like into some module in `common`, or after we merge https://github.com/rpm-software-management/ci-dnf-stack/pull/768 there will be the `consts.py` file into which it could go.

> > We'd probably be better off renaming it just for the sake of renaming, even though it's wide-spread and any name we come up with will be longer... roll_eyes
> 
> I don't understand this note. What should we rename or not rename?

This is mostly just a rant, but... The `DNFContext` should be renamed to e.g. `ScenarioContext` and subsequently `context.dnf` to `context.scenario_context` or a shorter version of the same... Because really everything under the `features` directory is related to dnf tests, naming it "dnf" doesn't mean anything, the real difference is that it has a per-scenario lifetime, unlike `context`, which lives for the whole run I think. This is good to know and not apparent to someone new to the code.

pkratoch:
> Oh, you've added the constant to the `createrepo_c` context too, I've missed that. Not sure I like it that much (as the steps in `common` are now dependent on the same, duplicate variable being present in two places). I guess it's not a big problem, but there should probably be a better place to put it. Like into some module in `common`, or after we merge #768 there will be the `consts.py` file into which it could go.

The `consts.py` sounds like a good place. @kontura can this PR wait for #768?

> This is mostly just a rant, but... The `DNFContext` should be renamed to e.g. `ScenarioContext` and subsequently `context.dnf` to `context.scenario_context` or a shorter version of the same... Because really everything under the `features` directory is related to dnf tests, naming it "dnf" doesn't mean anything, the real difference is that it has a per-scenario lifetime, unlike `context`, which lives for the whole run I think. This is good to know and not apparent to someone new to the code.

Ah, I get it now. Yeah, I agree that `context.scenario_context` or similar would be more descriptive.

kontura:
Yes, I think it can. As far as I know the createrepo_c tests are only run by us manually.

pkratoch:
Actually, I think it has to be in the context after all, so that it can be used in the encoding.feature. But at least it doesn't have to be in the context of createrepo_c tests.

PR #816, opened 2020-04-14
Swdb error tests, unprivileged user test steps and improvements

lukash:
Some generic improvements to the behave steps, as well as a step to run dnf under an unprivileged user.

Following that are tests for error handling of Swdb for an unprivileged user (permissions, database missing).

I've adopted @j-mracek's https://github.com/rpm-software-management/ci-dnf-stack/pull/810 and updated for the new step. @j-mracek please have a look, I've preserved your authorship even though I've modified the commit quite a bit. Hope you're happy with it :slightly_smiling_face: 

@dmach please feel free to check the test cases cover all the possible situations that can occur with Swdb, I couldn't think of more, seems there aren't that many in the end.

The tests are for https://github.com/rpm-software-management/libdnf/pull/935

### DIFF
--- a/dnf-behave-tests/common/cmd.py
+++ b/dnf-behave-tests/common/cmd.py
@@ -5,7 +5,7 @@ from __future__ import print_function
 
 import behave
 
-from common.lib.cmd import run_in_context
+from common.lib.cmd import assert_exitcode, run_in_context
 
 
 @behave.step("I set working directory to \"{working_dir}\"")
@@ -18,9 +18,21 @@ def when_I_execute_command_in_directory(context, command, directory):
     run_in_context(context, command.format(context=context), cwd=directory.format(context=context))
 
 
+@behave.step("I successfully execute \"{command}\" in \"{directory}\"")
+def when_I_successfully_execute_command_in_directory(context, command, directory):
+    when_I_execute_command_in_directory(context, command, directory)
+    assert_exitcode(context, 0)
+
+
 @behave.step("I execute \"{command}\"")
 def when_I_execute_command(context, command):
     run_in_context(context, command.format(context=context))
+
+
+@behave.step("I successfully execute \"{command}\"")
+def when_I_successfully_execute_command(context, command):
+    when_I_execute_command(context, command)
+    assert_exitcode(context, 0)
 
 
 @behave.step("I set LC_ALL to \"{value}\"")
--- a/dnf-behave-tests/features/alias-command.feature
+++ b/dnf-behave-tests/features/alias-command.feature
@@ -3,10 +3,7 @@
 Feature: Test for alias command
 
 Background:
-  Given I delete directory "/etc/dnf/aliases.d/"
-    And I delete file "/etc/yum.repos.d/*.repo" with globs
-    And I delete directory "/var/lib/dnf/modulefailsafe/"
-    And I use repository "alias-command"
+  Given I use repository "alias-command"
 
 
 Scenario: Add alias
--- a/dnf-behave-tests/features/environment.py
+++ b/dnf-behave-tests/features/environment.py
@@ -177,6 +177,12 @@ def before_scenario(context, scenario):
             context.config.userdata.get('destructive', 'no') == 'yes'):
         scenario.skip(reason="DESTRUCTIVE")
 
+    # if we are skipping a scenario, don't create the environment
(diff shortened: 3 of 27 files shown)


dnf-behave-tests/common/cmd.py
```diff
@@ -18,11 +18,23 @@ def when_I_execute_command_in_directory(context, command, directory):
     run_in_context(context, command.format(context=context), cwd=directory.format(context=context))
 
 
+@behave.given("I successfully execute \"{command}\" in \"{directory}\"")
```
dmach, line 21: This should be for all steps, no just `.given`.

In the past I advocated against the "I successfully execute" step, because it's not aligned with what I expect from behave. It merges given/when steps together with then. I would prefer to keep it as it is.

dnf-behave-tests/features/alias-command.feature
```diff
@@ -3,10 +3,7 @@
 Feature: Test for alias command
 
 Background:
-  Given I delete directory "/etc/dnf/aliases.d/"
-    And I delete file "/etc/yum.repos.d/*.repo" with globs
-    And I delete directory "/var/lib/dnf/modulefailsafe/"
-    And I use repository "alias-command"
```
dmach, line 9: You're removing similar lines from many places.
Is it safe? I did not go deep in the code review, that's why I'm asking even if it's maybe obvious.

lukash: What this does is cleans up the environment (the host system, since it's for `@no_installroot`) for the tests. I've moved this preparation to `environment.py` (it's in this commit), where I move these files to a backup location (so that if this gets run on a host system on accident, the user doesn't lose the files).

So yeah I think it's safe and shares code that always needs to be executed when running `@no_installroot`.

dnf-behave-tests/common/cmd.py
lukash: > This should be for all steps, no just `.given`.

I can do that :slightly_smiling_face: 

> In the past I advocated against the "I successfully execute" step, because it's not aligned with what I expect from behave. It merges given/when steps together with then. I would prefer to keep it as it is.

Though I don't really understand this paragraph... Should I also change the other "successfully" variations of the steps (those that are already there) to work for all behave steps? To be consistent and so that people don't get confused that some work for `Then` and some don't?

dnf-behave-tests/features/alias-command.feature
lukash: FTR it may be a bit easier to run the "cleaning" code in `environment.py` on a host system by accident, but hence it does a backup - I still think it's the best way as opposed to having to run all the steps in each destructive test (and forgetting to do that).

dnf-behave-tests/features/steps/cmd.py
```diff
@@ -111,6 +111,24 @@ def when_I_execute_dnf_with_args(context, args):
     run_in_context(context, cmd, can_fail=True)
 
 
+@behave.step("I execute dnf with args \"{args}\" as an unprivileged user")
+def when_I_execute_dnf_with_args(context, args):
+    """
+    Creates an unpriviged user if it doesn't exist (therefore the test is
+    destructive!). Runs the dnf command under this user.
+    """
```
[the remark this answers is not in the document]
lukash: Right :) that's what I thought but the other steps we have that are in effect destructive also don't have this check, so I just followed suit. Should we add this check?

dnf-behave-tests/features/environment.py
```diff
@@ -53,9 +67,10 @@ def __init__(self, userdata, force_installroot=False, no_installroot=False):
             self.installroot = tempfile.mkdtemp(prefix="dnf_ci_installroot_")
             # if we're creating installroot, ensure /etc/yum.repos.d exists,
             # otherwise dnf picks repo configs from the host
```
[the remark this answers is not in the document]
lukash: Indeed, I've removed it - there are now two cases (installroot and no installroot) and the comment would be confusing trying to describe them both.

dnf-behave-tests/features/steps/cmd.py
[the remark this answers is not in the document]
m-blaha: I'm not blocking this PR on it. But it would be nice to have this fixed in separate PR in future.

PR #820, opened 2020-04-29
Remove 'required' argument to make it compatible for rhel8 with pytho…

janblazek:
Remove 'required' argument to make it compatible for RHEL 8 with python 3.6.


m-blaha:
Oh, I missed that `required` keyword argument is not available until Python 3.7. Let's merge this quick fix and I will try to provide final solution.

PR #821, opened 2020-04-29
Creating various arch packages, mergerepo_c arch tests

kontura:
In order to test mergerepo_c --archlist --arch-expand I need packages of multiple different arches.


dnf-behave-tests/fixtures/specs/build.sh
```diff
@@ -63,11 +63,19 @@ for path in $DIR/*/*.spec; do
         echo "Building $path..."
         eval $RPMBUILD_CMD --target=$ARCH $path
 
-        # make lz4 multilib
-        if [[ "$SPEC_NAME" =~ "multilib" ]]; then
+        # make packages of different arches
+        if [[ "$SPEC_NAME" =~ "i686" ]]; then
```
lukash, line 67: The reason there was the "multilib" word in the spec file name is that it is building for i686 in addition to $ARCH (i.e. x86_64). Now this gets a bit confusing as the name always contains a specific architecture but the packages built are always for that architecture plus x86_64.

So I'm wondering what's the best and still simple way to do this - what about building all the architectures you're adding when "multilib" is in the name? This would add those extra arch packages to the tests which already use the multilib keyword so I'm not sure how good that is... Any other idea?

kontura: We could also only build the explicitly mentioned architectures with none defaulting to x86_64. So for the packages that require both we could do e.g. mid-a1.x86_64.i686.spec.

This is also not super clear but its more scalable, then again we might not need that. For the current situation I am fine with both honestly.

lukash: Okay... so maybe it's actually best to keep it the way you've done it, just put a comment here explicitly describing it's generating the architecture in the file name in addition to "x86_64" on purpose?

kontura: I meant I am fine with both the .multilib solution and any of the arch explicitly specified solutions, just in case we didn't understood each other. If we did just disregard this. :slightly_smiling_face: 

Will add the comment.

kontura: Comment updated.

PR #822, opened 2020-04-29
Tests that versionlocked package cannot get obsoleted

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1627124

### DIFF
--- a/dnf-behave-tests/features/plugins-core/versionlock-lock.feature
+++ b/dnf-behave-tests/features/plugins-core/versionlock-lock.feature
@@ -234,3 +234,40 @@ Scenario: Check-update command does not report updates filtered out by the versi
     """
     <REPOSYNC>
     """
+
+
+# PackageB-Obsoleter obsoletes PackageB < 3.0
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/plugins-core/versionlock-lock.feature
```diff
@@ -234,3 +234,40 @@ Scenario: Check-update command does not report updates filtered out by the versi
     """
     <REPOSYNC>
     """
+
+
+# PackageA-Obsoleter obsoletes PackageB < 3.0
```
kontura, line 239: Is this comment perhaps leftover from some other version of the test? Or how is PackageA-Obsoleter relevant here?

m-blaha: Yes, it's a typo. Fixed.

kontura: Thanks!

PR #823, opened 2020-04-30
Remove @not.with_os=rhel__eq__8 tags and protected packages

kontura:


dnf-behave-tests/features/steps/cmd.py
```diff
@@ -174,6 +174,11 @@ def step_impl(context):
     context.dnf._set("disable_plugins", False)
 
 
+@behave.given("I use hosts protected packages settings")
+def step_impl(context):
+    context.dnf.setopts.pop("protected_packages", None)
```
m-blaha, line 179: Why are you removing `protected_packages` key from setopts?

kontura: Because if we want to use hosts protected packages we cannot override `protected_packages` option using setopt in any way.

If we set it to empty string it doesn't protect any package overriding hosts setting (which this commit does by default in the tests).

m-blaha: I see, OK. But this should be eventually moved to generated dnf.conf file or something.

lukash: Instead of using host's protected packages, you should instead set `yum` as a protected package in the specific test that needs it.

It can be done through the generated config (as well as resetting the protected package list) and is imo a better solution - I've got the generated config implemented in a branch and am just waiting for https://github.com/rpm-software-management/ci-dnf-stack/pull/816 to be merged before submitting a PR.

kontura: ok, we can move it to the config file, but I think that if we set `yum` as protected manually it will defeat the purpose of the test. This [specific test](https://github.com/rpm-software-management/ci-dnf-stack/blob/master/dnf-behave-tests/features/remove-protected.feature) doesn't test that the protected functionality works, but that dnf when installed protects `yum` (meaning [this](https://github.com/rpm-software-management/dnf/blob/master/etc/dnf/protected.d/yum.conf) exists).  

Now that I think about it, it should rather be `@no_installroot` test. :)

m-blaha: I'm taking the PR 816 review.

lukash: @kontura I see, then it's fine I guess... though I'm not sure we need such a test, you might as well check the contents of that config file as installed from the package is "yum"... :slightly_smiling_face:

kontura:
I have updated the PR. 

@lukash we have introduced a problem by generating the configs. Microdnf no_installroot tests were trying to back up the generated config but assert was failing since original configs were already backed up. 
I made a commit where the configs are deleted instead, but I am not sure why did you create them in the first place? If the reason was to have the original configs accessible for debugging in the container, we could back up only the first configs (container originals) and delete all the subsequent ones instead of deleting them all?

I have also unified the protected packages features and added the config option `,` for protected packages inside installroots.

And finally removed some duplicate microdnf tests which are not needed anymore. 

@m-blaha if you can please also rereview.

lukash:
@kontura I'm backuping them instead of deleting for the case someone runs it on the host system by accident. Each of these tests should run in a clean environment so the backuped files shouldn't be there. Either I'm missing a case where this can happen and is correct or there is a bug where the tests are re-run in the same environment?

Since @m-blaha rewrote `dnf-testing.sh` to python each destructive test should run in a new container.

kontura:
@lukash I have dropped that commit, because after rebuilding the container I can no longer reproduce it, tried it several times. I must have messed something up locally before, not sure what. You are right it runs the destructive tests separately.

Sorry for the confusion, this should be ready for review now.

m-blaha:
All tests are passing with RHEL 8, merging.

PR #827, opened 2020-05-07
Create the config file from the steps

lukash:
Creates the dnf.conf in the default path inside installroot from behave steps.

@kontura it uses an empty value to unset a config option. We may need to use something else (like "<unset>" or any special placeholder?) for it to work with the protected packages option, since I think you need to set it to an empty value?

### DIFF
--- a/dnf-behave-tests/features/steps/config.py
+++ b/dnf-behave-tests/features/steps/config.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+import behave
+
+from common.lib.behave_ext import check_context_table
+from lib.config import write_config
+
+
+@behave.step("I configure dnf with")
+def step_configure_dnf(context):
+    """
+    Merges the new configuration values with what is stored in DNFContext and
+    writes the config file at <installroot>/etc/dnf/dnf.conf.
+
+    [main] is the default section, you can add more sections like so:
+
+    Given I configure dnf with
+          | key          | value |
+          | best         | False |
+          | [my_section] |       |
+          | foo          | bar   |
+    """
+    check_context_table(context, ["key", "value"])
+    section = "[main]"
+    context.dnf.config.setdefault(section, {})
+    for k, v in context.table:
+        if k.startswith("[") and v == "":
+            section = k
+            context.dnf.config.setdefault(section, {})
+
+        context.dnf.config[section][k] = v
+
+    write_config(context)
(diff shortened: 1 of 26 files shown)


dnf-behave-tests/features/steps/config.py
```diff
@@ -0,0 +1,36 @@
...
+          | [my_section] |       |
+          | foo          | bar   |
+    """
+    check_context_table(context, ["key", "value"])
+    section = "[main]"
+    context.dnf.config.setdefault(section, {})
+    for k, v in context.table:
+        if k.startswith("[") and v == "":
+            section = k
+            context.dnf.setdefault(section, {})
```
kontura, line 32: I think there is a missing `.config` -> `context.dnf.config.setdefault(section, {})`

lukash: Thanks, fixed.

kontura:
> We may need to use something else (like "" or any special placeholder?) for it to work with the protected packages option

Yes, if we don't want to protect any package, special placeholder you mean something like: `EMPTY_VALUE` and then check for that? 
If we don't want to do special handling we could use just `,` which I am not sure but I think resets given list option, but that is less readable.

lukash:
> Yes, if we don't want to protect any package, special placeholder you mean something like: `EMPTY_VALUE` and then check for that?

Yes. I'd rather have the empty "value" column represent an empty option in the config though. And use a special placeholder like `<unset>` to remove the option from the config altogether.

> If we don't want to do special handling we could use just `,` which I am not sure but I think resets given list option, but that is less readable.

Maybe we can do that. Putting an empty config option in the config file seems weird in general, perhaps it makes sense mostly with lists and we could handle this rare case with it. Could you try whether it works in your protected packages tests?

kontura:
I have tried it and the `,` could be used like this, the tests work. I would use it at least for now, so we can move the builds ahead.

Later on we should fix protected_packages in installroot which will fix the root cause and this will go away anyway.

lukash:
> I have tried it and the `,` could be used like this, the tests work. I would use it at least for now, so we can move the builds ahead.

Okay.

> Later on we should fix protected_packages in installroot which will fix the root cause and this will go away anyway.

How will we fix them? I understood this is the way we want it?

kontura:
The protected_packages are by default: `dnf`,  `glob:/etc/yum/protected.d/*.conf`  and  `glob:/etc/dnf/protected.d/*.conf`
I would expect that when using installroot we should load the default from `glob:/installroot/etc/dnf/protected.d/*.conf`.

PR #829, opened 2020-05-12
Test config option protect_running_kernel (RhBug:1698145)

pkratoch:
Requires:
- libdnf: https://github.com/rpm-software-management/libdnf/pull/957
- dnf: https://github.com/rpm-software-management/dnf/pull/1625

### DIFF
--- a/dnf-behave-tests/features/microdnf/protect-running-kernel.feature
+++ b/dnf-behave-tests/features/microdnf/protect-running-kernel.feature
@@ -1,0 +1,69 @@
+@no_installroot
+Feature: Protect running kernel
+
+
+Background: Install fake kernel
+  Given I use repository "protect-running-kernel"
+    And I successfully execute microdnf with args "install dnf-ci-kernel"
+    And I fake kernel release to "1.0"
+
+
+@bz1698145
+Scenario: Running kernel is protected
+   When I execute microdnf with args "remove dnf-ci-kernel"
+   Then the exit code is 1
+    And stderr is
+        """
+        error: Could not depsolve transaction; 1 problem detected:
+         Problem: The operation would result in removing the following protected packages: dnf-ci-kernel
+        """
+
+
+@bz1698145
+Scenario: Running kernel is not protected with config protect_running_kernel=False
+  Given I configure dnf with
+        | key                       | value            |
+        | protect_running_kernel    | 0                |
+   When I execute microdnf with args "remove dnf-ci-kernel"
+   Then the exit code is 0
+    And microdnf transaction is
+        | Action        | Package                             |
+        | remove        | dnf-ci-kernel-0:1.0-1.x86_64        |
+
+
+@bz1698145
+Scenario: Running kernel is protected when in protected_packages even with config protect_running_kernel=False
+  Given I configure dnf with
+        | key                       | value            |
+        | protected_packages        | dnf-ci-kernel    |
+   When I execute microdnf with args "remove dnf-ci-kernel"
+   Then the exit code is 1
+    And stderr is
+        """
+        error: Could not depsolve transaction; 1 problem detected:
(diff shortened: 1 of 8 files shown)


dnf-behave-tests/features/microdnf/protected-running-kernel1.feature
```diff
@@ -0,0 +1,21 @@
...
+
+@bz1698145
+Scenario: Running kernel is protected
+  Given I use repository "repoquery-unsatisfied"
+   When I execute microdnf with args "install dnf-ci-kernel"
+   Then the exit code is 0
+    And microdnf transaction is
+        | Action        | Package                            |
+        | install       | dnf-ci-kernel-0:1.0-1.x86_64       |
+        | install       | dnf-ci-systemd-0:1.0-1.x86_64      |
```
lukash, line 13: I think instead of the three steps above you can just use:
```gherkin
Given I successfully execute microdnf with args "install dnf-ci-kernel"
```
(if on accident nothing would get installed and `microdnf` exited with 0, the tests would surely fail later on?)

You can also now put multiple microdnf tests into a single .feature file, the runner now launches them in separate containers.

And if you do that, you can put the shared `Given` steps into `Background` to save on repetition :slightly_smiling_face:

dnf-behave-tests/features/protected-running-kernel.feature
```diff
@@ -0,0 +1,60 @@
...
+@bz1698145
+@no_installroot
+Scenario: Running kernel is protected
+  Given I use repository "repoquery-unsatisfied"
+   When I execute dnf with args "install dnf-ci-kernel"
+   Then the exit code is 0
+    And DNF Transaction is following
+        | Action        | Package                             |
+        | install       | dnf-ci-kernel-0:1.0-1.x86_64        |
+        | install-dep   | dnf-ci-systemd-0:1.0-1.x86_64       |
```
lukash, line 13: Same here, you can do a single
```gherkin
Given I successfully execute dnf with args "install dnf-ci-kernel"
```

And put the shared `Given` steps into `Background`.

lukash:
Also a nitpick, the .feature files could be called `protect_running_kernel` to match the option name exactly...

dnf-behave-tests/features/microdnf/protected-running-kernel1.feature
pkratoch: Done.

dnf-behave-tests/features/protected-running-kernel.feature
pkratoch: Done.

dnf-behave-tests/features/microdnf/protect-running-kernel.feature
```diff
@@ -0,0 +1,45 @@
+@no_installroot
+Feature: Protect running kernel
+
+
+Background: Install fake kernel
+  Given I use repository "repoquery-unsatisfied"
```
lukash, line 6: The tests are actually using a repository belonging to another feature file, I think we should avoid this and likely create a dedicated repo for these tests? Or, we could create a "shared" repo named accordingly, e.g. `kernel-protected`. Although, the `repoquery-unsatisfied` actually tests quite a special case that's just related to a kernel package.

pkratoch: I'll create a dedicated repository. I also just came up with one more test that will require additional package, so it will be better that way.

pkratoch:
I originally though that the step ``microdnf transaction is`` checks only the transaction report from microdnf and not RPMDB, because it didn't complain about missing ``dnf-ci-kernel`` removal in context table... but it turned out that the bug was in RPMDB check.

So, I added fix for the RPMDB check, step ``only microdnf transaction is`` (if you think of better name, let me know!) and the additional tests (mainly to test that the protect-running-kernel option works also for other commands, not just "remove").

PR #832, opened 2020-05-14
Tests for handling group name being None

lukash:
Marking as WIP to discuss how it actually behaves when the name is None: In `group list` "None" is printed in the output and in case of installing the group is not listed in the transaction (see the last commit of this PR).


lukash:
Rebased on the updates to the transaction check step.

PR #833, opened 2020-05-15
copr plugin tests

lukash:
Basic set of tests for the copr plugin, including a test for https://github.com/rpm-software-management/dnf-plugins-core/pull/388.

Also adds a step for creating a stand-alone http server and printing of server failures to the output.

@dturecek @praiskup @xsuchy

### DIFF
--- a/dnf-behave-tests/features/plugins-core/copr.feature
+++ b/dnf-behave-tests/features/plugins-core/copr.feature
@@ -1,0 +1,203 @@
+@destructive
+Feature: Test the COPR plugin
+
+Background:
+Given I enable plugin "copr"
+  And I create directory "/{context.dnf.tempdir}/copr"
+  And I start http server "copr" at "{context.dnf.tempdir}/copr"
+  And I create and substitute file "//etc/dnf/plugins/copr.conf" with
+      """
+      [main]
+      distribution = Fedora
+      releasever = 30
+      [testhub]
+      hostname = localhost
+      protocol = http
+      port = {context.dnf.ports[copr]}
+      """
+  And I create and substitute file "/{context.dnf.tempdir}/copr/coprs/testuser/testproject/repo/fedora-30/dnf.repo" with
+      """
+      [copr:localhost:testuser:testproject]
+      name=Copr test project
+      baseurl=http://project_base_url/
+      type=rpm-md
+      """
+
+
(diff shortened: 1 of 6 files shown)


dnf-behave-tests/features/plugins-core/copr.feature
```diff
@@ -0,0 +1,142 @@
...
+      [copr:localhost:testuser:testrepo]
+      name=Copr test repo
+      baseurl=http://repo_base_url/
+      type=rpm-md
+      """
+
+
+Scenario: Test enabling and disabling a repository
+ When I execute dnf with args "copr enable testcopr/testuser/testrepo"
+ Then the exit code is 0
```
xsuchy, line 29: Nope. It is called as:
```
  copr enable rhscl/perl516 epel-6-x86_64
  copr enable ignatenkobrain/ocltoys
```
Similary for remove/disable.

dturecek: This is actually correct. However, in terminology we use, this should probably be `copr enable testhub/testuser/test{copr, project}`. The hub needs to be specified here because Fedora copr would be used otherwise.

dturecek:
The tests look good to me.

dnf-behave-tests/features/plugins-core/copr.feature
lukash: I'll change it, @dturecek, but I don't understand the `{copr, project}` part, are those two words interchangeable or did you mean something else by this?

dturecek: Yes, we use both. But I think lately we are using `project` more, so I'd change it to `copr enable testhub/testuser/testproject`. Sorry for the confusion.

lukash: Ok, thanks and no worries. I'll update it in a bit (the last force-push updates the PR to work with @praiskup's https://github.com/rpm-software-management/dnf-plugins-core/pull/398 and I don't want to mix it up)

lukash: @dturecek I've fixed the naming, you can check whether it adheres to your conventions if you want...

kontura:
LGTM.

PR #834, opened 2020-05-18
Prefer platform-python if present

m-blaha:


janblazek:
Thank you, Marek. I have tried this on RHEL8 and it works as expected.

PR #837, opened 2020-05-21
Fix history longer than 80 scenario

inknos:
This scenario is failing for different installroot paths
Now it checks that the length is a 3 digit number

### DIFF
--- a/dnf-behave-tests/features/history-list.feature
+++ b/dnf-behave-tests/features/history-list.feature
@@ -194,12 +194,13 @@ Scenario: history lame (no transaction with such package)
 
 @bz1786335
 @bz1786316
+# TODO change this to actually verify stdout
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/history-list.feature
```diff
@@ -194,12 +194,13 @@ Scenario: history lame (no transaction with such package)
 
 @bz1786335
 @bz1786316
+# TODO change this to actually verify stdout
 Scenario: history longer than 80 characters
- When I execute dnf with args "history | head -1 | wc -c"
+ When I execute dnf with args "history | head -1 | wc -c | wc -c"
```
j-mracek, line 199: Please could you add a note? It should explain why you use the approach

PR #838, opened 2020-05-22
Add tests for various aliased/depricated/yum-compat commands (RhBug:1821524)

kontura:
https://bugzilla.redhat.com/show_bug.cgi?id=1821524

These are tests for https://github.com/rpm-software-management/dnf/pull/1611

### DIFF
--- a/dnf-behave-tests/features/yum-compat.feature
+++ b/dnf-behave-tests/features/yum-compat.feature
@@ -1,0 +1,161 @@
+Feature: Testing aliased/depricated commands compatible with yum
+
+
+Background:
+Given I use repository "simple-base"
+
+
+Scenario: install using install-n
+ When I execute dnf with args "install-n labirinto"
+ Then the exit code is 0
+  And Transaction is following
+      | Action        | Package                       |
+      | install       | labirinto-0:1.0-1.fc29.x86_64 |
+
+
+@bz1821524
+Scenario: using install-n doesn't allow additional NEVRA parts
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/features/yum-compat.feature
```diff
@@ -0,0 +1,138 @@
...
+      """
+
+
+Scenario: install using autoremove-n
+Given I successfully execute dnf with args "install vagare"
+ When I execute dnf with args "autoremove-n vagare"
+ Then the exit code is 0
+  And Transaction is following
+      | Action       | Package                       |
+      | remove       | labirinto-0:1.0-1.fc29.x86_64 |
```
pkratoch, line 105: Due to commit https://github.com/rpm-software-management/ci-dnf-stack/commit/d57c06d077774b492c7f59b237f46fc4c4c3d1dc, the action for labirinto should be ``remove-unused``.

pkratoch:
There is also ``install-nevra`` command that might deserve a test (it should fail if the argument is incomplete nevra).

Otherwise looks good.

kontura:
I have added the test for `install-nevra`, also I checked and all of these commands have the variants, so there is for example `autoremove-n`, `autorotemove-na` and `autorotemove-nevra`, I could add tests for all of them if you think its necessary.

pkratoch:
I think it's ok. The tests passed locally, so I am merging this.

PR #840, opened 2020-05-28
Fix using nonexistent module in module/install-errors.feature

pkratoch:
ModuleX does not exist in any of the repositories, therefore, the
scenarios tested non-existent module and not non-existent
stream/profile.


lukash:
Interestingly enough the error messages are the same :confused: 

Shouldn't we keep (add) a test for a non-existent module as well? Or did I just miss it?

pkratoch:
> Interestingly enough the error messages are the same confused

Yes, I presume that's why this happened in the first place.

> Shouldn't we keep (add) a test for a non-existent module as well? Or did I just miss it?

It's the first two scenarios in the same feature file. "A proper error message is displayed when I try to install a non-existent module"

PR #841, opened 2020-05-28
New "while IFS= read -r package" in sign.sh excpects a new line at the end

kontura:
PR https://github.com/rpm-software-management/ci-dnf-stack/pull/839
introduced a change where lines without newline at the end in
`../keyspecs/${KEY_NAME}/packages` are ignored. This broke a test.

Another option would be to make the loop in sign.sh more resilient to this problem, but I think its preferable to have all the `packages` files with the same format.


m-blaha:
I would prefer fixing the while loop. Something like this `while IFS= read -r package || [ -n "$package" ]; do` might do the job.

kontura:
Updated.

PR #842, opened 2020-05-29
DNF Transaction: distinguish userinstalled/dependency install and remove

lukash:
The transaction check was ignoring the reason for installation/removal. This adds distinguishing between userinstalled/dependency/weak dependency for install and similarly for remove.

It's a lot of changes, majority is just dumb replacement.


dmach:
@lukash have you also considered adding a new column?
Your approach merges action with reason.
Maybe keeping action and reason as 2 different columns would be cleaner.
An empty value would mean that we're not checking the value.

j-mracek:
LGTM!

lukash:
@dmach as discussed off-PR, I haven't considered it but it would mean splitting an information that is now stored in one place (the captions in dnf output under which the packages in the transaction are listed) into two places, so it doesn't make much sense to me. Also, I think it's good to always check the value (the exact reason), it doesn't cost much and makes the tests more exact and reliable.

@j-mracek thanks! Can ve merge then? :sunglasses:

j-mracek:
All test passed (locally), right now.

PR #844, opened 2020-06-04
Update @not.with_os=rhel__eq__8 tags for rhel-8.3.0 builds

inknos:


kontura:
One thing I forgot to mention, can you keep the:
```
# createrepo_c is compiled without support for zchunk on rhel 8
@not.with_os=rhel__eq__8
```
We didn't add zchunk support in the release. You can test the createrepo_c test suite by:
```
./container-test -s createrepo_c run
```

And also there already are new tests in comps-group.feature, so can you rebase the PR and add the `@not.with_os=rhel__eq__8` to them as well?

kontura:
Great, thanks!

PR #846, opened 2020-06-05
Add test for logfilelevel (RhBug:1802074)

kontura:
https://bugzilla.redhat.com/show_bug.cgi?id=1802074

Tests for:
https://github.com/rpm-software-management/libdnf/pull/973
https://github.com/rpm-software-management/dnf/pull/1631

### DIFF
--- a/dnf-behave-tests/features/logs.feature
+++ b/dnf-behave-tests/features/logs.feature
@@ -26,3 +26,63 @@ Scenario: dnf.rpm.log doesn't contain duplicate entries ()
         .* INFO --- logging initialized ---
         .* SUBDEBUG Erase: flac-1.3.3-3.fc29.x86_64
         """
+
+
+@bz1802074
+Scenario Outline: logfilelevel <level> controls verbosity of dnf.log
+Given I use repository "dnf-ci-fedora-updates"
+ When I configure dnf with
+      | key          | value   |
+      | logfilelevel | <level> |
+  And I execute dnf with args "install flac"
+  And I execute dnf with args "remove flac"
+ Then file "/var/log/dnf.log" contains lines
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/logs.feature
```diff
@@ -26,3 +26,42 @@ Scenario: dnf.rpm.log doesn't contain duplicate entries ()
...
+      | 1     | 48        | 7             | 4            |
+      | 2     | 48        | 7             | 4            |
+      | 3     | 64        | 7             | 4            |
+      | 4     | 64        | 7             | 4            |
+      | 5     | 64        | 7             | 4            |
+      | 6     | 64        | 7             | 4            |
+      | 7     | 92        | 7             | 4            |
+      | 8     | 92        | 7             | 4            |
+      | 9     | 92        | 7             | 4            |
+      | 10    | 92        | 71            | 23           |
```
lukash, line 67: Heh... I quite dislike this, because the numbers are going to change slightly on every change to what's logged during the dnf operations used... but at the same time it's a test.. and it's quite cleverly done :slightly_smiling_face: 

Though.. could we check via regex that there's a message with the expected level in the logfile instead? Then the test wouldn't depend on the exact number of lines...

kontura: ok, I have pushed an update. :slightly_smiling_face: 
I am also thinking about keeping only one level for each unique setting, this way it runs 20 tests in total which seems redundant.

lukash: Agreed, one test per debug level should be enough.

kontura: Updated.

PR #848, opened 2020-06-10
Return correct exit code from dnf-testing.sh

m-blaha:
The wrapper script allways returned 0 which is not correct.
Now it will return the exit code of the underlying container-test program.

### DIFF
--- a/dnf-testing.sh
+++ b/dnf-testing.sh
@@ -2,7 +2,7 @@
 
 PROG_PATH="$(dirname "$(realpath -s -- "$0")")"
 
-pushd "$PROG_PATH" > /dev/null || return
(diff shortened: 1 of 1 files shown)


dnf-testing.sh
```diff
@@ -20,5 +20,8 @@ if [ ! "$PYTHON" ]; then
 fi
 
 "$PYTHON" container-test "$@"
+EXITCODE=$?
 
-popd || exit
+popd > /dev/null || :
```
lukash, line 25: What is the `|| :` for? Would there be a more clear way to do it, this is just totally cryptic to anyone not knowing the exact semantics? From what I've quickly tried to find I would guess it actually doesn't do anything?

PR #851, opened 2020-06-11
Add dnf/yum command version for verbose updateinfo.feature

kontura:


dnf-behave-tests/features/updateinfo.feature
```diff
@@ -415,8 +415,9 @@ Scenario: updateinfo shows summary for advisories using direct commands (yum com
 
 
 @bz1801092
-Scenario: updateinfo lists advisories referencing CVE with dates in verbose mode
-  Given I successfully execute dnf with args "install glibc flac"
+Scenario: updateinfo lists advisories referencing CVE with dates in verbose mode (DNF version)
+  Given I set dnf command to "dnf"
```
lukash, line 419: Is this not the default?

kontura: It is on fedora, but on rhel the default is yum. So in order for the test to work on both we need to be explicit.

lukash: Oh, is it? Why? Seems more trouble than it's worth? (we've got those explicit yum tests, so I'm kind of lost in why we have both those and the default changed to yum on RHEL)

kontura: The default didn't change, the name of dnf always was yum or RHEL. This is the same [problem](https://github.com/rpm-software-management/ci-dnf-stack/pull/770) as other test have and they usually have both variants, so I wanted to be consistent with it at least. 

You would prefer to just keep the yum version here?

lukash: I don't know, I think I'm really not getting the general idea here. When I commented in the PR you've linked, I didn't know the whole suite is running with the command set to yum. In fact I even asked about it: https://github.com/rpm-software-management/ci-dnf-stack/pull/770#issuecomment-575602901

So what are we testing here? We know yum is just a symlink, so the only thing we are testing is that dnf says "yum" in the output instead of "dnf". I find it minor enough to not being worth it to add complexity to the scenarios. In fact I don't even thing this "feature" should be there... it's just misleading people. (_"Wait, am I running yum or dnf here?"_)

It's not really so important, but I'd like to have a clear idea of what we're doing, this just seems like we're adding arbitrary combinations? Or is it just me?

kontura: To me it doesn't feel like arbitrary combinations since we have just dnf and yum. It's like if dnf had a switch `--yum` that would change all outputs to have "yum" instead of "dnf" we would want to test that switch as well.

The way I understand it is even though yum is just a symlink it is quite important we don't show "dnf" in output when running as yum. By rejecting https://github.com/rpm-software-management/ci-dnf-stack/pull/770 we basically said it is important enough to have the added complexity and duplicated tests. Or at least I don't have any other idea how to do it, we probably don't want to hardcode dnf for rhel tests.

kontura: I just realized I could change it to `Scenario Outline` with variants for dnf and yum. That is a bit better, but the core confusion remains.

lukash: > To me it doesn't feel like arbitrary combinations since we have just dnf and yum. It's like if dnf had a switch `--yum` that would change all outputs to have "yum" instead of "dnf" we would want to test that switch as well.

I meant arbitrary combinations of ways to test (default dnf command, a step to override the command) and the commands we actually test.

Though I do see we actually have the `I set dnf command to "dnf"` variants where `"yum"` is tested.

> The way I understand it is even though yum is just a symlink it is quite important we don't show "dnf" in output when running as yum. By rejecting #770 we basically said it is important enough to have the added complexity and duplicated tests. ...

I guess...

So lets merge this then... If you wanna do the `Outline` I'll wait for that, otherwise I can merge as it is.

kontura: > So lets merge this then... If you wanna do the Outline I'll wait for that..

Nah, I would then want to change all the test that have this and its not worth it to me now, you can go ahead. :slightly_smiling_face:

PR #852, opened 2020-06-15
[history] test reverse history list output (RhBug:1846692)

inknos:
dnf: https://github.com/rpm-software-management/dnf/pull/1632
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1846692

### DIFF
--- a/dnf-behave-tests/features/history-list.feature
+++ b/dnf-behave-tests/features/history-list.feature
@@ -212,3 +212,32 @@ Scenario: history length is 80 chars when missing rows are queried
   """
   80
   """
+
+@bz1846692
+Scenario: history list --reverse
+ When I execute dnf with args "history list --reverse"
+ Then the exit code is 0
+  And stdout is history list
+      | Id | Command | Action  | Altered |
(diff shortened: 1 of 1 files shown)


lukash:
I personally think having so many tests for the `--reverse` switch is a bit overkill. I understand it was cheap (as in the time needed to copy the `history-list.feature` and add the `--reverse` option), but it seems unnecessary and the tests need maintenance as well as they take time each time the test suite is run. While the usefulness / cost threshold is quite low (even if the test is just a little bit useful, it's better to have it), i'd say here most of the tests are below the threshold for me :slightly_smiling_face: 

How about picking just a few of the tests, just enough to test the `--reverse` functionality, and drop the rest? And if there are just a few, I'd rather just put them into the `history-list.feature` file.

inknos:
> How about picking just a few of the tests, just enough to test the `--reverse` functionality, and drop the rest? And if there are just a few, I'd rather just put them into the `history-list.feature` file.

It is ok, I will reduce the number of tests.

PR #854, opened 2020-06-18
Tests for dnf debug-dump / dnf-debug-restore

m-blaha:
requires: https://github.com/rpm-software-management/dnf-plugins-core/pull/402

### DIFF
--- a/dnf-behave-tests/features/plugins-core/debug-dump.feature
+++ b/dnf-behave-tests/features/plugins-core/debug-dump.feature
@@ -1,0 +1,76 @@
+Feature: Test for debug plugin - dumping
+
+
+Scenario: dnf debug-dump dumps file with configuration
+  Given I enable plugin "debug"
+    And I use repository "debug-plugin"
+    And I successfully execute dnf with args "install kernel-4.19.1 kernel-4.20.1"
+   When I execute dnf with args "debug-dump {context.dnf.tempdir}/dump.txt"
+   Then the exit code is 0
+    And file "//{context.dnf.tempdir}/dump.txt" matches line by line
+    """
+    dnf-debug-dump version 1
+    %%%%SYSTEM INFO
+      uname: .*
+      rpm ver: .*
+      python ver: .*
+    %%%%DNF INFO
+      arch: .*
+      basearch: .*
+      releasever: .*
+      dnf ver: .*
+      enabled plugins: debug
+      global excludes: 
+    %%%%RPMDB PROBLEMS
+    %%%%RPMDB
+      kernel-0:4\.19\.1-fc29\.x86_64
+      kernel-0:4\.20\.1-fc29\.x86_64
+    %%%%REPOS
+    %debug-plugin - .*/fixtures/repos/debug-plugin
+      excludes: 
+      kernel-0:4.18.1-fc29.src
+      kernel-0:4.18.1-fc29.x86_64
+      kernel-0:4.19.1-fc29.src
+      kernel-0:4.19.1-fc29.x86_64
+      kernel-0:4.20.1-fc29.src
+      kernel-0:4.20.1-fc29.x86_64
+      test-remove-0:1-fc29.src
+      test-remove-0:1-fc29.x86_64
+      test-replace-0:1-fc29.src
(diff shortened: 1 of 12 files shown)


dnf-behave-tests/features/plugins-core/debug-dump.feature
```diff
@@ -0,0 +1,78 @@
...
+    %%%%REPOS
+    %debug-plugin - file:///home/mblaha/work/debug-restore-bz1844533/ci-dnf-stack/dnf-behave-tests/fixtures/repos/debug-plugin
+      excludes: 
+      kernel-0:4.18.1-fc29.src
+      kernel-0:4.18.1-fc29.x86_64
+      kernel-0:4.19.1-fc29.src
+      kernel-0:4.19.1-fc29.x86_64
+      kernel-0:4.20.1-fc29.src
+      kernel-0:4.20.1-fc29.x86_64
+      test-install-0:1-fc29.src
```
kontura, line 37: This is the only place where the test-install package is used and since there is a lot of other packages listed here I think we don't really need it.

dnf-behave-tests/features/plugins-core/debug-dump.feature
```diff
@@ -0,0 +1,78 @@
...
+      releasever: .*
+      dnf ver: .*
+      enabled plugins: debug
+      global excludes: 
+    %%%%RPMDB PROBLEMS
+    %%%%RPMDB
+      kernel-0:4\.19\.1-fc29\.x86_64
+      kernel-0:4\.20\.1-fc29\.x86_64
+    %%%%REPOS
+    %debug-plugin - file:///home/mblaha/work/debug-restore-bz1844533/ci-dnf-stack/dnf-behave-tests/fixtures/repos/debug-plugin
```
kontura, line 29: I believe this is your local path and it doesn't match in general. We could hard code the path from our container `%debug-plugin - file:///opt/behave/fixtures/repos/debug-plugin`, but I think we should use another regex here, since the tests can be run in different setups as well.

dnf-behave-tests/features/plugins-core/debug-dump.feature
```diff
@@ -0,0 +1,78 @@
...
+      test-remove-0:1-fc29.src
+      test-remove-0:1-fc29.x86_64
+      test-replace-0:1-fc29.src
+      test-replace-0:1-fc29.x86_64
+      test-replace-0:2-fc29.src
+      test-replace-0:2-fc29.x86_64
+      test-replace-0:3-fc29.src
+      test-replace-0:3-fc29.x86_64
+    %%%%RPMDB VERSIONS
+      all: 2:9a5b819daf813e63ecfd470f07bcacb534b506d5
```
kontura, line 48: This hash doesn't work when I tried to run it, I don't think we can rely on it being always the same, there is too many moving parts.

dnf-behave-tests/features/plugins-core/debug-dump.feature
```diff
@@ -0,0 +1,78 @@
...
+      releasever: .*
+      dnf ver: .*
+      enabled plugins: debug
+      global excludes: 
+    %%%%RPMDB PROBLEMS
+    %%%%RPMDB
+      kernel-0:4\.19\.1-fc29\.x86_64
+      kernel-0:4\.20\.1-fc29\.x86_64
+    %%%%RPMDB VERSIONS
+      all: 2:9a5b819daf813e63ecfd470f07bcacb534b506d5
```
kontura, line 77: Here as well.

dnf-behave-tests/features/plugins-core/debug-dump.feature
m-blaha: Right. It is a left-over of previous version.

dnf-behave-tests/features/plugins-core/debug-dump.feature
m-blaha: Oh, sorry. I'll fix it.

m-blaha:
Alesi, thanks for the review. The issues should be fixed now.

kontura:
The test pass now, merging.
Thanks!

PR #855, opened 2020-06-22
Add tests for up/downgrade to RPM with different epoch (RhBug:1845069)

pkratoch:


dnf-behave-tests/features/epoch.feature
```diff
@@ -0,0 +1,48 @@
+Feature: Test upgrading and downgrading to package with different epoch
+
+
+Scenario: Upgrade to RPM with same NVR but different epoch
+  Given I use repository "epoch0"
```
lukash, line 5: How about naming the repos `epoch-[1-3]`?

dnf-behave-tests/features/epoch.feature
```diff
@@ -0,0 +1,48 @@
...
+        | Action        | Package                |
+        | upgrade       | dummy-1:1.0-1.x86_64   |
+
+
+@xfail
+Scenario: Upgrade to RPM with same NVR but different epoch from RPM with no epoch
+  Given I use repository "epoch"
+    And I successfully execute dnf with args "install dummy"
+    And I use repository "epoch1"
+   When I execute dnf with args "downgrade dummy"
```
lukash, line 20: It says "Upgrade" in Scenario name, but here it does a downgrade...

Also, I've come to a conclusion the `@xfail` tests are really not that useful, since they are always ignored (as in not run) and therefore they rot. It would be great if we could fix the problems instead, as it seems the `@xfail`s are just "never" going to be addressed. Just ranting though, it's still better than nothing :slightly_smiling_face:

dnf-behave-tests/features/epoch.feature
pkratoch: It's named this way, so that the name of the repository corresponds to the epoch defined in spec. So, repositories ``epoch0`` and ``epoch1`` contain packages with epochs ``0`` and ``1``, respectively, and repository ``epoch`` contains package without explicit epoch definition. I find it easy to work with. Is there any reason for preferring ``epoch-[1-3]`` naming?

dnf-behave-tests/features/epoch.feature
pkratoch: > It says "Upgrade" in Scenario name, but here it does a downgrade...

Sorry about that, I fixed that. Also, this scenario is actually not failing, so I removed ``@xfail``.

> Also, I've come to a conclusion the `@xfail` tests are really not that useful, since they are always ignored (as in not run) and therefore they rot. It would be great if we could fix the problems instead, as it seems the `@xfail`s are just "never" going to be addressed. Just ranting though, it's still better than nothing slightly_smiling_face

I wrote these tests when investigating bug 1845069. Normally, I would write patch and then these tests wouldn't be tagged ``@xfail``, but since the bug was reassigned to rpm, I don't have control over when it's going to be fixed.

dnf-behave-tests/features/epoch.feature
lukash: I see, I missed that the package without an epoch is in a repository with no number in its name.

I was just thinking we should keep the convention of naming the repos according to the feature files and in case there are more, name them e.g. `epoch-something`, `epoch-something-else` etc. But your naming makes sense and I don't have a better idea so we can keep it as it is...

lukash:
Sorry, totally forgot about this. Can you rebase the PR? That should trigger a CI run.

PR #856, opened 2020-06-25
Add test for error in download --resolve (RhBug:1844925)

pkratoch:
Test for https://github.com/rpm-software-management/dnf-plugins-core/pull/400


dnf-behave-tests/features/plugins-core/download-binary.feature
```diff
@@ -42,6 +42,21 @@ Scenario: Download an existing RPM with dependencies
         | {context.dnf.tempdir}/setup-2.12.1-1.fc29.noarch.rpm      | file://{context.dnf.fixturesdir}/repos/dnf-ci-fedora/noarch/setup-2.12.1-1.fc29.noarch.rpm    |
 
 
+@bz1844925
```
m-blaha, line 45: This test will fail on RHEL8, right? It should be marked @not.with_os=rhel__eq__8

pkratoch: Right, on the other hand, branch rhel-8.3.0 already exists, so aren't we now in the mode of not adding the tag by default?

m-blaha: Fair enough.

PR #858, opened 2020-06-29
Tests for installroot as unpriveliged user (RhBug:1843280)

pkratoch:
Tests https://github.com/rpm-software-management/dnf/pull/1635

### DIFF
--- a/dnf-behave-tests/features/unprivileged-installroot.feature
+++ b/dnf-behave-tests/features/unprivileged-installroot.feature
@@ -1,0 +1,24 @@
+@destructive
+@no_installroot
+Feature: Installroot test as unprivileged user
+
+
+Scenario: Fail when installing into installroot as unprivileged user
+  Given I use repository "miscellaneous"
+   When I execute dnf with args "--releasever=32 --installroot=/home/testuser/f32 install dummy" as an unprivileged user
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/unprivileged-installroot-2.feature
```diff
@@ -0,0 +1,14 @@
...
+@destructive
+@no_installroot
+Scenario: Fail when missing permissions for installroot directory
+  Given I use repository "miscellaneous"
+   When I execute dnf with args "--releasever=32 --installroot=/var/lib/f32 install dummy" as an unprivileged user
+   Then the exit code is 1
+    And stderr is
+        """
+        Config error: [Errno 13] Permission denied: '/var/lib/f32': '/var/lib/f32'
+        """
```
lukash, line 14: AFAIK you can now put the two scenarios into a single .feature file and the runner will run them in separate containers if they are `@no_installroot` (you can also mark the whole `Feature` `@no_installroot` I think).

pkratoch: Hm, I guess I must have run it without the runner, because I thought for sure it was failing. All right then, I put it into one file.

PR #866, opened 2020-07-16
Add dnf.librepo.log to the log-rotate test (RhBug:1816573)

pkratoch:
Requires https://github.com/rpm-software-management/libdnf/pull/999 and https://github.com/rpm-software-management/dnf/pull/1639

### DIFF
--- a/dnf-behave-tests/features/log-rotate.feature
+++ b/dnf-behave-tests/features/log-rotate.feature
@@ -4,12 +4,14 @@ Feature: Log rotation
 @bz1702690
 Scenario: Size and number of log files respects log_size and log_rotate options
   Given I use repository "dnf-ci-fedora"
-    And I execute dnf with args "--setopt=log_size=512 --setopt=log_rotate=2 install glibc"
-    And I execute dnf with args "--setopt=log_size=512 --setopt=log_rotate=2 remove glibc"
-    And I execute dnf with args "--setopt=log_size=512 --setopt=log_rotate=2 install glibc"
-    And I execute dnf with args "--setopt=log_size=512 --setopt=log_rotate=2 remove glibc"
-    And I execute dnf with args "--setopt=log_size=512 --setopt=log_rotate=2 install glibc"
-    And I execute dnf with args "--setopt=log_size=512 --setopt=log_rotate=2 remove glibc"
+    And I execute dnf with args "--setopt=log_size=1024 --setopt=log_rotate=2 install glibc"
+    And I execute dnf with args "--setopt=log_size=1024 --setopt=log_rotate=2 remove glibc"
+    And I execute dnf with args "--setopt=log_size=1024 --setopt=log_rotate=2 install glibc"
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/features/log-rotate.feature
```diff
@@ -40,3 +41,17 @@ Scenario: Size and number of log files respects log_size and log_rotate options
...
    Then size of file "tmp_dnf.rpm.log.2" is less than "512"
+
+   When I execute "ls {context.dnf.installroot}/var/log | grep "dnf\.librepo\.log""
+   Then stdout is
+        """
+        dnf.librepo.log
+        dnf.librepo.log.1
+        dnf.librepo.log.2
+        """
+   When I execute "head -n1 {context.dnf.installroot}/var/log/dnf.librepo.log > {context.dnf.installroot}/tmp_dnf.librepo.log"
```
lukash, line 52: Hum, shouldn't it really be `head -n-1` in all the checks? AFAICS this outputs the first line only to the tmp file... :sunglasses: 

Also, I haven't run the tests on this yet, just a question, does this still pass? I'd think the format is now unified...
https://github.com/rpm-software-management/ci-dnf-stack/blob/662576c2d15c3727d520aac5b4da87076f1cf55d/dnf-behave-tests/features/logs.feature#L69

pkratoch: > Hum, shouldn't it really be `head -n-1` in all the checks? AFAICS this outputs the first line only to the tmp file... sunglasses

Oh my... You are absolutely right and when I fix it, the tests don't pass anymore, because there can be one message at the end of the file that spreads over multiple lines. I have to figure out how to make it work.

> Also, I haven't run the tests on this yet, just a question, does this still pass? I'd think the format is now unified...
> 
> https://github.com/rpm-software-management/ci-dnf-stack/blob/662576c2d15c3727d520aac5b4da87076f1cf55d/dnf-behave-tests/features/logs.feature#L69

I don't understand this comment. What format and how is it related?

lukash: > > Also, I haven't run the tests on this yet, just a question, does this still pass? I'd think the format is now unified...
> > https://github.com/rpm-software-management/ci-dnf-stack/blob/662576c2d15c3727d520aac5b4da87076f1cf55d/dnf-behave-tests/features/logs.feature#L69
> 
> I don't understand this comment. What format and how is it related?

The log file format. From the linked test case I figured the format of the line is a bit different (note the `.*` which is not present in the other logs) and I was wondering if this test still passes. But as you said in the libdnf PR the fomrat actually appears to be the same, so I'm not sure why the difference here. I'll try to run the tests later...

pkratoch: Oh, I see now. The difference is between librepo and hawkey logs. Hawkey log has the level string (e.g. "DEBUG") before the time. I ran this feature, too, and it passes.

pkratoch: I fixed the test. The "head" is not needed at all, but the log_size had to be increased, see the commit message.

PR #879, opened 2020-08-12
Parse "replacing" line in microdnf transaction table

pkratoch:
Microdnf started to report obsoleted, downgraded, upgraded and
reinstalled packages (i.e. the nevra that is removed) all with
keyword "replacing".

The context table can now contain actions "upgraded", "downgraded",
"reinstalled" and "obsoleted". These actions can be compared with
the RPMDB as usual, and for comparison with the microdnf output,
they are all considered as "replaced" actions.

### DIFF
--- a/dnf-behave-tests/features/steps/transaction.py
+++ b/dnf-behave-tests/features/steps/transaction.py
@@ -207,6 +207,7 @@ def parse_microdnf_transaction_table(lines):
     lines = lines[:transaction_end]
 
     label_re = re.compile(r"^([^ ].+):$")
+    replacing_re = re.compile(r"^replacing +(?P<nevra>[^ ]*)$")
     action = None
     result = []
     for line in lines:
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/features/steps/transaction.py
```diff
@@ -207,6 +207,7 @@ def parse_microdnf_transaction_table(lines):
     lines = lines[:transaction_end]
 
     label_re = re.compile(r"^([^ ].+):$")
+    replacing_re = re.compile(r"^(?P<label>replacing) +(?P<nevra>[^ ]*)$")
```
lukash, line 210: The `label` regex group is not used, so it can be removed from the regex... it's a good idea to keep the regexes as simple as possible, they're hard to read enough already :slightly_smiling_face:

pkratoch: A remnant from another regex. Fixed.

PR #885, opened 2020-08-25
Test upgrading groups and environments

pkratoch:


dnf-behave-tests/features/comps-group-upgrade.feature
```diff
@@ -0,0 +1,100 @@
...
+  Given I drop repository "comps-group-upgrade-2"
+    And I use repository "comps-group-upgrade-1"
+   When I execute dnf with args "group upgrade testgroup"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                            |
+        | install-group | mandatory-0:1.0-1.x86_64           |
+        | install-group | default-0:1.0-1.x86_64             |
+        | install-group | conditional-true-0:1.0-1.x86_64    |
+        | group-upgrade | testgroup                          |
```
lukash, line 77: Hi, did I miss something or do the tests actually not test upgrading packages to a new version? You've mark-installed in `Background` and here it actually installs the missing packages, but `dnf group upgrade` should also upgrade already installed packages, right?

dnf-behave-tests/fixtures/specs/comps-environment-upgrade-1/comps.xml
```diff
@@ -0,0 +1,33 @@
...
+   <name>testgroup</name>
+   <description>Testgroup for DNF CI testing</description>
+    <packagelist>
+      <packagereq type="mandatory">mandatory</packagereq>
+      <packagereq type="default">default</packagereq>
+      <packagereq type="optional">optional</packagereq>
+      <packagereq type="conditional" requires="dummy">conditional-true</packagereq>
+      <packagereq type="conditional" requires="nonexistent">conditional-false</packagereq>
+    </packagelist>
+  </group>
```
lukash, line 19: Is the group definition and the actual packages the same as in the group tests? Perhaps it would actually be better to share the two repositories for the tests?

you could likely even put them into the same feature file to maintain the naming convention of feature file and repo name...

On more of a side note, I think it would be a good idea to put the comps tests into their own directory under `features`, eventually...

dnf-behave-tests/features/comps-group-upgrade.feature
pkratoch: So far, I was only interested in picking the right packages to install when doing group upgrade (e.g. not installing reinstalled packages or packages that were already part of the group but were not installed) to determine if we can do group upgrade during system upgrade (see https://github.com/rpm-software-management/dnf-plugins-extras/pull/183). In the repositories that I defined, there are no packages in multiple versions, so nothing gets upgraded in these scenarios. But it's a good point, I should also add test the actual package upgrade.

dnf-behave-tests/fixtures/specs/comps-environment-upgrade-1/comps.xml
pkratoch: > Is the group definition and the actual packages the same as in the group tests? Perhaps it would actually be better to share the two repositories for the tests?

Not sure what you mean. The comps-group.feature uses dnf-ci-thirdparty and dnf-ci-fedora and it's more messy. But I can change it so that both features use the same group-specific repositories.

> you could likely even put them into the same feature file to maintain the naming convention of feature file and repo name...
> 
> On more of a side note, I think it would be a good idea to put the comps tests into their own directory under `features`, eventually...

Makes sense.

dnf-behave-tests/features/comps-group-upgrade.feature
pkratoch: Hm, the naming of packages is probably confusing, because there is e.g. mandatory and mandatory-2 that are different packages. Maybe it would be better to rename them to mandatory-A and mandatory-B.

dnf-behave-tests/fixtures/specs/comps-environment-upgrade-1/comps.xml
lukash: > > Is the group definition and the actual packages the same as in the group tests? Perhaps it would actually be better to share the two repositories for the tests?
> 
> Not sure what you mean. The comps-group.feature uses dnf-ci-thirdparty and dnf-ci-fedora and it's more messy. But I can change it so that both features use the same group-specific repositories.

I mean the `testgroup` comps group definition as well as the RPMs in these two commits appear to be identical (though I didn't compare them side by side): https://github.com/rpm-software-management/ci-dnf-stack/pull/885/commits/4b7d83b492fd3ed9d8912659477380faf913fdee, https://github.com/rpm-software-management/ci-dnf-stack/pull/885/commits/0ce0601d8c57b63190be1208ef2a864c5a596a4f

The only difference then being the presence of the env group, and so I think the whole two repositories could be shared between the comps group upgrade and comps env upgrade tests?

pkratoch: Oh, I see. Another difference is having different group in the environment in comps-environment-upgrade-2 to tests that new group will be installed. But yeah, it could be simplified.

dnf-behave-tests/features/comps-group-upgrade.feature
pkratoch: I added test for upgrading packages during group upgrade and renamed the packages. Hope it's better readable now.

dnf-behave-tests/fixtures/specs/comps-environment-upgrade-1/comps.xml
pkratoch: I unified the repositories and added both group and environment tests into one feature.

dnf-behave-tests/features/comps-upgrade.feature
```diff
@@ -0,0 +1,231 @@
...
+Scenario: Upgrade group when there are new package versions - upgrade packages
+  Given I successfully execute dnf with args "group install A-group"
+    And I use repository "comps-upgrade-2"
+   When I execute dnf with args "group upgrade A-group"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                            |
+        | upgrade       | A-mandatory-0:2.0-1.x86_64         |
+        | upgrade       | A-default-0:2.0-1.x86_64           |
+        | upgrade       | A-conditional-true-0:2.0-1.x86_64  |
```
lukash, line 52: This test is failing for me and I don't see versions 2.0 of the packages in the `comps-upgrade-2` repository, did you by chance forget to add them?

dnf-behave-tests/features/comps-upgrade.feature
```diff
@@ -0,0 +1,231 @@
...
+#     mandatory: A-mandatory
+#     default: A-default
+#     optional: A-optional
+#     conditional: A-conditional-true (if dummy), A-conditional-false (if nonexistent)
+#   AB-group:
+#     mandatory: A-mandatory
+#     default: A-default
+#     optional: A-optional
+#     conditional: A-conditional-true (if dummy), A-conditional-false (if nonexistent)
+#   testenvironment:
```
lukash, line 15: The environment is called "AB-environment", I assume you've had it named "testenvironment" previously?

This is the problem when you copy the information from somewhere else. For myself (taking into account I know where to look already), I prefer to look at the comps.xml instead here, you can't rely on the real data being the same as the contents of this comment... So I don't find this very useful (the xml data are easy to read too, if, say, the data were scattered across multiple files and it was hard to get the general idea, the value of this would be higher), but it's subjective, feel free to keep this if you find it helpful.

dnf-behave-tests/features/comps-upgrade.feature
pkratoch: Sorry, yes, I forgot to add them.

lukash: Thanks, works now. Are you planning on addressing https://github.com/rpm-software-management/ci-dnf-stack/pull/885#discussion_r480979965? I don't mind much merging this the way it is, but having the comment accurate would be better...

pkratoch: It's fixed now. I was thinking if the comment is useful for this feature file, but I can't tell, since I wrote the feature myself and it's still fresh in my mind. But in general, I find these comments useful, so I am keeping it.

PR #889, opened 2020-09-07
Tests for the system-upgrade plugin

lukash:
Requires: https://github.com/rpm-software-management/dnf-plugins-extras/pull/184 (without it it reboots the system the tests are run on).

### DIFF
--- a/dnf-behave-tests/features/plugins-extras/system-upgrade.feature
+++ b/dnf-behave-tests/features/plugins-extras/system-upgrade.feature
@@ -1,0 +1,119 @@
+@destructive
+Feature: Test the system-upgrade plugin
+
+Background:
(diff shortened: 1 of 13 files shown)


dnf-behave-tests/fixtures/specs/system-upgrade-f30/pkg-b-2.0-1.spec
```diff
@@ -0,0 +1,15 @@
+Name:           pkg-b
+Version:        1.0
```
pkratoch, line 2: The file name should be ``pkg-b-1.0-1.spec``.

lukash: Fixed, thanks.

PR #890, opened 2020-09-11
Tests for the system-upgrade plugin using Transaction Store/Replay

lukash:


dnf-behave-tests/features/plugins-extras/system-upgrade.feature
```diff
@@ -117,3 +117,61 @@ Given I successfully execute dnf with args "system-upgrade reboot"
...
+  And stderr is
+      """
+      [Errno 2] No such file or directory: '/var/lib/dnf/system-upgrade/system-upgrade-transaction.json'
+      """
+
+
+Scenario: Test system-upgrade downloading a package from a different repo
+Given I configure repository "system-upgrade-2-f$releasever" with
+      | key     | value |
+      | enabled | 0     |
```
lukash, line 150: @j-mracek I've tried to reproduce the case when the same package is in two repositories, and is downloaded from one but installed from the other. But so far I wasn't able to reproduce it, no matter what I do, it either works or ends up with a different error. For example here dnf errors out with:
```
Error: Cache-only enabled but no cache for 'system-upgrade-2-f30'
```
Because the `system-upgrade download` changes the cache directory and because this repo is disabled, it doesn't download it at all, so it ends up missing the data when attempting to run `system-upgrade upgrade`.

Do you have an idea how to reproduce the case? The thing is while the repo id is being stored in the transaction json, it is not used in any way as of now, so presumably, if the case described can happen at all, it should be failing.

j-mracek: With system-upgrade cache it is little bit complicated, because it overrides the origin path during runtime. It means that system-upgrade is incompatible with --installroot option.

lukash: It does work, just not very well. The cache is still overridden consistently for all system-upgrade commands (otherwise it wouldn't work at all), it just uses the same cache path on the host system for an installroot as well. I was thinking whether I should change the plugin to respect the installroot and use a path inside it as the cache. Didn't do it yet in the end, what do you think about that?

Also, this has no impact on the behavior of the download I've described in my previous comment, AFAICS?

j-mracek: It sounds good to respect an installroot for modifier of cache path

lukash: @j-mracek I've added the installroot prefix to the datadir (in the plugins-extras PR).

I've also tested the case with the same package in two repos according to your (off-PR) suggestion to change the order in which the repos are loaded by dnf. I was able to reproduce it only by hacking dnf to change the order, since Python 3.7 the order of items in a dict is [guaranteed to be deterministic](https://docs.python.org/3/whatsnew/3.7.html). Hence it seems to not be possible to reproduce the case in the CI stack and the question is whether we should address it in the code?

j-mracek: @lukash Do you plan any changes? (I suggest that the question was answered during our call.)

lukash: @j-mracek I've done the changes on Friday, and I've posted an update here: https://github.com/rpm-software-management/dnf/pull/1659#issuecomment-702787838

Sorry about the disconnect. The repo priorities you've suggested did indeed reproduce the issue, so the test is updated to use that and works properly now. I've added the use of `repo_id` to dnf, hence the update there, as I wanted you to take a look at that.

Otherwise I think the PRs are ready.

PR #891, opened 2020-09-16
Test group/environment upgrade during system upgrade

pkratoch:
Tests for: https://github.com/rpm-software-management/dnf-plugins-extras/pull/183


kontura:
Can you please add a tag `@use.with_os=fedora__ge__31` so its run only on fedora? otherwise LGTM.

pkratoch:
> Can you please add a tag `@use.with_os=fedora__ge__31` so its run only on fedora? otherwise LGTM.

Added, thanks.

kontura:
Tests pass locally.

PR #894, opened 2020-09-17
Test decoding local baseurls

lukash:


kontura:
LGTM and tests pass locally, thanks!

PR #896, opened 2020-10-06
Add REPOSYNC to output check of some updateinfo tests

lukash:
The reposync happens sometimes and fails the tests.


kontura:
Thanks for the fix!

PR #897, opened 2020-10-13
Add a test for installing x86_64 package not pulling in i686 through deps

lukash:
The test is a simplified reproducer from:
https://bugzilla.redhat.com/show_bug.cgi?id=1782899

Which is fixed in libsolv upstream and should be released in 0.7.16.

Also adds a test for a problem discovered when reproducing the above,
which was subsequently reported in libsolv too.


lukash:
Note this requires the fix in libsolv, which should be present in our CI (which builds libsolv from upstream git) and in our nightly builds. But it wasn't released in Fedora yet.

kontura:
LGTM and works.
Also tested with older version of libsolv and the test does indeed fail and installs `foo-1.0-1.i686`.

PR #899, opened 2020-10-15
Dockerfile: Install the system-upgrade plugin

lukash:
It is now being tested by the test suite.


kontura:
Seeing this I realized the `system-upgrade.feature` will cause troubles on rhel (in gating) since there is no system-upgrade plugin, can you please also add a tag to it so its run only on fedora? I think `@use.with_os=fedora__ge__32` should be fine?

lukash:
Right. I've used Fedora 31 or greater and I've verified the tag works for the whole feature file too...

kontura:
Great, thanks!

PR #902, opened 2020-10-26
test history size when output is redirected (RhBug:1852577)

inknos:
dnf: https://github.com/rpm-software-management/dnf/pull/1636/

### DIFF
--- a/dnf-behave-tests/features/history-list.feature
+++ b/dnf-behave-tests/features/history-list.feature
@@ -194,13 +194,15 @@ Scenario: history lame (no transaction with such package)
 
 @bz1786335
 @bz1786316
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/history-list.feature
```diff
@@ -194,17 +194,21 @@ Scenario: history lame (no transaction with such package)
...
   And stdout is
   """
-  4
+  80
   """
 
 @bz1786335
 @bz1786316
+@bz1852577
+@bz1906970
```
j-mracek, line 211: This is a problem. The scenario did not fail before introduction of the patch, therefore it is not correct.

PR #904, opened 2020-10-29
When expect fails catch the exception and show just dnf output

kontura:
This makes it much more readable.

### DIFF
--- a/dnf-behave-tests/features/steps/shell.py
+++ b/dnf-behave-tests/features/steps/shell.py
@@ -17,6 +17,19 @@ def stdout_from_shell(context):
     context.cmd_stdout = context.shell_session.before.decode().replace("\r\n", "\n")
 
 
+def expect_from_shell(context, shell_out):
+    try:
+        context.shell_session.expect(shell_out)
+    except pexpect.exceptions.EOF:
+        stdout_from_shell(context)
+        raise AssertionError("expected: '" + str(shell_out) + "' was not found in dnf shell output,"
+                             " it terminated unexpectedly")
+    except pexpect.exceptions.TIMEOUT:
+        stdout_from_shell(context)
+        raise AssertionError("expected: '" + str(shell_out) + "' was not found in dnf shell output,"
+                             " it timed out")
+
+
 @behave.step("I open dnf shell session")
 def when_I_open_dnf_shell(context):
     cmd = " ".join(context.dnf.get_cmd(context)) + " shell"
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/steps/shell.py
```diff
@@ -42,12 +46,20 @@ def when_I_execute_in_shell(context, command):
     context.shell_session.sendline(command.format(context=context))
 
     if command.strip() == "quit" or command.strip() == "exit":
-        context.shell_session.expect(pexpect.EOF)
+        try:
+            context.shell_session.expect(pexpect.EOF)
+        except pexpect.exceptions.EOF:
+            stdout_from_shell(context)
+            raise AssertionError("dnf shell unexpected output")
```
lukash, line 53: I don't think you'll get the `EOF` exception here, since you're expecting `EOF`. The other exception you can get is `TIMEOUT` (https://pexpect.readthedocs.io/en/stable/_modules/pexpect/exceptions.html).

The way I understand it you either get what you `expect()` or it waits for it until it times out. Might be worth it catching that too with some descriptive message like "expected prompt '> ' was not found in dnf shell output, timed out" (for the case where you expect the prompt)

dnf-behave-tests/features/steps/shell.py
```diff
@@ -27,8 +27,12 @@ def when_I_open_dnf_shell(context):
...
     # to turn off TTY echo; the echo is still there though, so removing the delay
     context.shell_session.delaybeforesend = None
-    
-    context.shell_session.expect('> ')
+
+    try:
+        context.shell_session.expect('> ')
+    except pexpect.exceptions.EOF:
+        stdout_from_shell(context)
+        raise AssertionError("dnf shell unexpected output")
```
lukash, line 35: I think a better error text would be "dnf shell terminated unexpectedly" or something along those lines?

dnf-behave-tests/features/steps/shell.py
```diff
@@ -42,12 +46,20 @@ def when_I_execute_in_shell(context, command):
...
         stdout_from_shell(context)
         context.shell_session = None
         return
 
-    # previously, there was timeout=600 added in commit 415d980eb34e8fd0487f1be3da14a8c279b74993
-    # but it's probably not needed anymore
-    context.shell_session.expect("\r\n[^ \r-]*> ")
+    try:
+        # previously, there was timeout=600 added in commit 415d980eb34e8fd0487f1be3da14a8c279b74993
+        # but it's probably not needed anymore
```
lukash, line 60: The default timeout is 30, I think you can remove this comment.

dnf-behave-tests/features/steps/shell.py
kontura: I see, yes you are right. Ok I have extracted it into a small wrapper for the exceptions and added more descriptive messages.

lukash:
Nice, thanks.

PR #905, opened 2020-11-03
Add tests for running dnf with solv* files only 

kontura:
Tests for https://github.com/rpm-software-management/libdnf/pull/1076

I can drop the last two tests with `xfail` if you think their value is too low or this is not the place to keep such information.

https://bugzilla.redhat.com/show_bug.cgi?id=1865803


kontura:
Closing in favor of: https://github.com/rpm-software-management/dnf/pull/1703 and https://github.com/rpm-software-management/dnf/pull/1711

PR #906, opened 2020-11-09
Tests for the new groups-manager dnf plugin

m-blaha:
Requires: https://github.com/rpm-software-management/dnf-plugins-core/pull/412

### DIFF
--- a/dnf-behave-tests/common/file.py
+++ b/dnf-behave-tests/common/file.py
@@ -236,3 +236,11 @@ def then_file_sha256_checksums_are_following(context):
 def create_compressed_file_with(context, compression, filepath):
     target_path = prepend_installroot(context, filepath)
     create_compressed_file_with_contents(target_path, compression, context.text)
+
+
+@behave.step("I compress file \"{filepath}\" using \"{compression}\"")
+def create_compressed_file_with(context, compression, filepath):
+    file_path = prepend_installroot(context, filepath)
+    ensure_file_exists(file_path)
+    create_compressed_file_with_contents(
+        file_path, compression, read_file_contents(file_path))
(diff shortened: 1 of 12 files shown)


lukash (commented):
Binary file alert! :slightly_smiling_face: I would _really_ like to not have binary files in git. Any change in that is hard to review and in this case makes altering the contents of that comps file more difficult, as you need to re-gzip it. There's already a `create_compressed_file_with()` step in `createrepo_c` tests, so you can move it to `common` and use it in scenario setup to compress the file to the scenario's temporary location. I think it's the best solution.

Can you please add a few tests for basic error handling, like file not found, comps.xml syntax error... possibly others if any of the operations can fail on something (e.g. merging?).

m-blaha:
> Binary file alert! slightly_smiling_face I would _really_ like to not have binary files in git. Any change in that is hard to review and in this case makes altering the contents of that comps file more difficult, as you need to re-gzip it. There's already a `create_compressed_file_with()` step in `createrepo_c` tests, so you can move it to `common` and use it in scenario setup to compress the file to the scenario's temporary location. I think it's the best solution.

:) I was not aware or this step (we should think about creating some kind of a "directory" of existing steps). Makes sense and I will change it.

lukash:
> ... we should think about creating some kind of a "directory" of existing steps ...

You mean in a documentation sense to have an overview? Not sure it's worth the effort unless I'm missing an easy way to do it without making a duplicate list, which would imo just go out of sync. But I agree they're hard to find as they're scattered across files and there even may be some redundancy as a result of that too. But if you have a good idea? :slightly_smiling_face:

m-blaha:
@lukash I was thinking about using Doxygen (or another doc generating tool) and basically generate documentation for steps from docstrings. But it still would require some work - not all steps have those strings.

lukash:
> I was thinking about using Doxygen (or another doc generating tool) and basically generate documentation for steps from docstrings. But it still would require some work - not all steps have those strings.

That could work.. just requires effort to set up and would we then put it up on a webserver somewhere or would people generate it locally...

PR #908, opened 2020-11-10
Add test for a new behavior of modular filtering with source

j-mracek:
Source packages will not remove binary packages


kontura:
LGTM, thanks!

PR #912, opened 2020-11-16
Module switch tests

j-mracek:


dnf-behave-tests/features/module/switch.feature
```diff
@@ -0,0 +1,74 @@
...
+   When I execute dnf with args "module switch nodejs:8"
+   Then the exit code is 0
+    And Transaction contains
+        | Action                   | Package                                       |
+        | downgrade                | nodejs-1:8.11.4-1.module_2030+42747d40.x86_64 |
+        | downgrade                | npm-1:8.11.4-1.module_2030+42747d40.x86_64 |
+        | module-stream-switch     | nodejs:10 -> 8     |
+    And modules state is following
+        | Module    | State     | Stream    | Profiles       |
+        | nodejs    | enabled   | 8         |  minimal       |
```
lukash, line 74: I know it may be hard to create some scenarios, but could you add at least some tests for error handling when the switch fails? At least non-existent module/stream should be trivial?

j-mracek: ok

j-mracek: I create an additional test. @lukash Please could you take a look?

dnf-behave-tests/features/module/switch.feature
```diff
@@ -72,3 +72,33 @@ Scenario: I can switch a module stream, install packages from a new module and k
...
+        | Module    | State     | Stream    | Profiles       |
+        | nodejs    | enabled   | 10        |  minimal       |
+   When I execute dnf with args "module switch nodejs:notexists"
+   Then the exit code is 1
+    And Transaction is empty
+    And modules state is following
+        | Module    | State     | Stream    | Profiles       |
+        | nodejs    | enabled   | 10        |  minimal       |
+    And stderr contains "Error: Problems in request:"
+    And stderr contains "missing groups or modules: nodejs:notexists"
```
lukash, line 104: Please use the `stderr is` step unless it's just not possible due to variable output. It is much more reliable than this. I'd like to stress it is valuable to even go out of your way to create/modify existing steps just to avoid unreliable checks like these.

Also, the `module install nodejs:8/minimal` and `module switch nodejs:10` steps aren't necessary as a setup just to get the error?

Thanks for the test and I would squash it with the other commit if it wasn't your intention all along :wink:

j-mracek: I am sorry, they are necessary, because the test also check whether state of module nodejs was changed.

PR #914, opened 2020-11-23
Fix system-upgrade comps tests

lukash:
The new transaction store/replay code stores the comps actions into the transaction json, so they are replayed during the upgrade transaction.


pkratoch:
Why are the groups not included in transactions in some scenarios? I know the tests pass, but I am curious why this happens.

lukash:
> Why are the groups not included in transactions in some scenarios? I know the tests pass, but I am curious why this happens.

Good question, I was a bit hasty about this and didn't even notice. I'll investigate.

lukash:
@pkratoch so I'm reasonably certain it's because of code inconsistency in adding the groups/envs to the history transaction through a different code path if there are no packages. If there are packages, the code path taken also tests the `rpm.RPMTRANS_FLAG_TEST` (which is on for `system-upgrade download`) and doesn't write groups to the history if that's the case. The code is a spaghetti monster. I want to at least not make it worse, might even try to improve some things if possible. I don't think we need to block this PR on it, lets get the CI working and I'll prepare a fix for the transaction behavior.

PR #921, opened 2020-12-03
Test that reposync --downloadcomps works with not compressed comps

m-blaha:
Requires: https://github.com/rpm-software-management/dnf/pull/1702


kontura:
Test passes, thanks!

PR #922, opened 2020-12-08
Add test for module remove --all with argument that has no match

j-mracek:
It tests that command passes and does not generate a traceback.

### DIFF
--- a/dnf-behave-tests/features/module/remove.feature
+++ b/dnf-behave-tests/features/module/remove.feature
@@ -214,4 +214,8 @@ Scenario: module removed with --all and not existing module argument - no traceb
    When I execute dnf with args "module remove --all noexists"
    Then the exit code is 0
     And Transaction is empty
-    And stderr does not contain "Traceback"
+    And stderr is
+        """
+        Problems in request:
+        missing groups or modules: noexists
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/features/module/remove.feature
```diff
@@ -206,3 +206,12 @@ Scenario: Packages belonging to multiple modules are not removed with --all
...
     And stdout does not contain "belongs to multiple modules, skipping"
+
+@bz1904490
+Scenario: module removed with --all and not existing module argument - no traceback
+  Given I use repository "dnf-ci-fifthparty"
+    And I use repository "dnf-ci-fifthparty-modular"
+   When I execute dnf with args "module remove --all noexists"
+   Then the exit code is 0
+    And Transaction is empty
+    And stderr does not contain "Traceback"
```
kontura, line 217: For the new test I would prefer using a positive step, the expectation is more explicit.
```suggestion
    And stderr is
        """
        Problems in request:
        missing groups or modules: noexists
        """
```

kontura (changes requested):

kontura:
Test passes, thanks!

PR #923, opened 2020-12-09
Adjust tests of outputs according to the latest improvements in dnf

j-mracek:
DNF is going to provide more information when it is unable to enable
a stream.


lukash:
I'll just nitpick on the commit message, it is very non-descriptive. Something like this would seem better:
> Adjust test outputs for improved module enable error messages

(also, thanks for changing it to the `stderr is ...` test :slightly_smiling_face: )

j-mracek:
Fixed

m-blaha:
LGTM!

PR #925, opened 2020-12-15
<REPOSYNC> doesn't have to be the first line (fixes verbose mode matching)

kontura:
For these particular tests `<REPOSYNC>` matches the potential `Last
metadata expiration check: 0:00:01 ago on Thu Dec 10 08:38:47 2020`
the downloading of repository is matched by the regex
`dnf-ci-fedora-updates test repository .* MB/s | .*`

I concede this is not the most clear or readable approach and we could
create something like `<REPOSYNC_VERBOSE>` to match the whole beginning
part but that would hide what repositories are used which I think is
a useful part of the verbose output.

For: https://baseos-stg-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/dnf-libdnf-test/226/testReport/

### DIFF
--- a/dnf-behave-tests/common/output.py
+++ b/dnf-behave-tests/common/output.py
@@ -12,18 +12,20 @@ from common.lib.text import lines_match_to_regexps_line_by_line
 
 
 def handle_reposync(expected, found):
-    if expected[0] == "<REPOSYNC>":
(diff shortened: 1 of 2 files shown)


lukash:
Yeah, given the circumstances this is fine, thanks for fixing it.

PR #934, opened 2021-01-12
Use delta rpm createrepo_c tests only on rhel 8 (RhBug:1914828)

kontura:
https://bugzilla.redhat.com/show_bug.cgi?id=1914828


kontura:
Previously I have misunderstood the request, we want to keep the test for fedoras. Disable it just for rhel 9 and higher.

PR #938, opened 2021-01-21
[needs-restarting] fix -r in nspawn containers

inknos:
https://github.com/rpm-software-management/dnf-plugins-core/pull/422


m-blaha (approved):
Thanks!

PR #940, opened 2021-01-27
Test logfile permissions (RhBug:1910084)

pkratoch:
Tests for https://github.com/rpm-software-management/dnf/pull/1714

### DIFF
--- a/dnf-behave-tests/common/cmd.py
+++ b/dnf-behave-tests/common/cmd.py
@@ -4,8 +4,11 @@ from __future__ import absolute_import
 from __future__ import print_function
 
 import behave
+import glob
+import os
 
 from common.lib.cmd import assert_exitcode, run_in_context
+from common.lib.file import prepend_installroot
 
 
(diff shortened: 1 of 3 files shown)


kontura:
I have just one suggestion, it would make sense to me to test that rotating the logs also respects umask. I think this could be achieved by setting dnf conf `log_size` to some really small value so that it rotates right away for the first command, maybe `1k`?

Other than that LGTM. :slightly_smiling_face:

pkratoch:
Thanks for the review and the suggestion of new test!

There was actually another bug - the new dnf.rpm.log was sometimes created with mode 644 because rpm is setting umask during a transaction run.

I think that the best behavior is not to respect umask during the log rotation, but keep the same mode as the old log. This also means that user can change permissions of the base log file and it will be respected further on.

So, here is the dnf PR: https://github.com/rpm-software-management/dnf/pull/1736

And I added test that log rotation preserves the log file permissions.

kontura:
That is great, thanks!

Tests pass.

PR #941, opened 2021-01-28
libmodulemd-2.10.0 always outputs quoted stream names

kontura:
This is described in the release notes:
https://github.com/fedora-modularity/libmodulemd/releases/tag/libmodulemd-2.10.0


lukash:
Tested locally, makes the tests pass.

PR #945, opened 2021-02-03
[microdnf] Add test of the "protected_packages" option (in dnf.conf)

jrohel:
Requires: https://github.com/rpm-software-management/libdnf/pull/1104


j-mracek:
LGTM

PR #946, opened 2021-02-04
Speed rpm -qa up by passing --nodigest --nosignature options

dmach:
We need to list NEVRAs, digest or signature verification is not important.
Cuts rpm execution time by ~70%.

### DIFF
--- a/dnf-behave-tests/features/steps/lib/rpmdb.py
+++ b/dnf-behave-tests/features/steps/lib/rpmdb.py
@@ -21,6 +21,9 @@ def get_rpmdb_rpms(installroot="/"):
     """
     result = []
     ts = rpm.TransactionSet(installroot)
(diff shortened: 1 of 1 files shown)


m-blaha:
We should remove the dnf-docker-test directory. It is no longer used and could cause confusion.

dmach:
I made the original change in a wrong location, but it revealed a technical debt at least :)
I force-pushed a new patch, fixing it in the right location.

lukash:
I'm ready for the amazing speedup :slightly_smiling_face: let's see...

glum23:
Hi, just FYI, removal of dnf-docker-test directory will probably break our workflows, it is still present in our scripts and dockerfiles for older releases. It'd be nice to discuss changes like that with us next time. Thanks

PR #948, opened 2021-02-05
Remove dnf-docker-test directory (it is not used for testing any more)

m-blaha:


dnf-behave-tests/features/config.feature
```diff
@@ -63,6 +58,7 @@ Scenario: Reposdir option in dnf.conf file in installroot
   Given I configure dnf with
         | key      | value      |
         | reposdir | /testrepos |
+    And I execute "createrepo_c {context.scenario.repos_location}/dnf-ci-fedora"
```
lukash, line 61: Why did you add these? I does appear the repodata actually weren't generated... but the tests were passing?

Tbh. I don't like adding these createrepo_c calls, the idea was the behave steps do this for you. We do already have some createrepo_c (and mergerepo_c) calls in some scenarios, I just didn't see a simple way to abstract it in steps...

Maybe we could have a `I generate repository data for "NAME" with "ARGS"`... but let's not complicate this now. Just thinking out loud really. :slightly_smiling_face:

m-blaha: The test was passing only by pure chance. The thing is that the `baseurl` of the newly configured directory points to global `fixtures/repos/dnf-ci-fedora` directory. And the repository there was created by some of previous scenarios. If this scenario was the first one executed it would fail (without creating the repodata).
But I agree that having a step for this is better than calling createrepo_c directly. I can have a look and prepare such step.

PR #952, opened 2021-02-09
Shared actions and overlay for CI in Github Actions

lukash:


Dockerfile.f32
```diff
@@ -42,18 +42,18 @@
     dnf -y install fakeuname;
 
 
+# upgrade all packages to the latest available versions
+RUN set -x && \
+    dnf -y --refresh upgrade
+
+
```
lukash, line 49: @m-blaha I wanted to point out this commit to you, do you think it would have any unwanted side-effects? Was there a reason to not upgrade to nightlies versions of the stack?

I've introduced a new overlay for the CI which only builds components we maintain, so I need stuff from nightlies...

dmach:
I've briefly reviewed the PR and it looks good to me. Let's merge it and test how it works in production.

Dockerfile.f32
m-blaha: I think the `distro` build type will stop working - you will allways get the latest nightlies copr build. The jenkins job uses `jjb` build type which I'm not sure is needed. We could either switch it to `local` or (if we do not want to get debugging tools installed in jenkins container) change condition on following line to something like 
`if [ "$TYPE" == "distro"]; then `

lukash: Ok, I'll have a look at that. I never really understood the build types I guess, so I just left it as it is to be a "future Lukas problem" :slightly_smiling_face: 

I know the jjb is not needed and the test result xmls are unused. We'll see how the log outputs are usable for "test results" and if I need to figure out something better.

PR #958, opened 2021-02-12
Add tests for protection of reason for installonlypkgs

j-mracek:


LICENSE
```diff
@@ -262,7 +262,7 @@ in one of these ways:
     copy of the Corresponding Source for all the software in the
     product that is covered by this License, on a durable physical
     medium customarily used for software interchange, for a price no
-    more than your reasonable cost of physically performing this
```
m-blaha, line 265: :) probably not intended...

j-mracek: No idea how it happened.

PR #966, opened 2021-03-02
Ci host container

lukash:
Nightly generation of the host container to run the CI on.


.github/workflows/ci-host.yml
```diff
@@ -0,0 +1,31 @@
...
+          IMAGE="ghcr.io/rpm-software-management/dnf-ci-host"
+
+          echo ${{secrets.CONTAINER_PAT}} | docker login ghcr.io -u dnf-bot --password-stdin
+
+          docker build -t ${IMAGE}:${DATE} -t ${IMAGE}:latest -<<EOF
+          FROM fedora:latest
+          RUN dnf -y update
+          RUN dnf -y install dnf-plugins-core
+          RUN dnf -y copr enable rpmsoftwaremanagement/rpm-gitoverlay
+          RUN dnf -y install rpm-gitoverlay fuse-overlayfs parallel podman wget
```
dmach, line 27: This leaves metadata cache (/var/cache/dnf) on disk.
Many containers remove it after the container is created, on the other hand, it might be worth keeping it if you know that containers using the image will use it.
In such case, consider removing only the solv files (`rm -rf /var/cache/dnf/*.solv*`) to reduce the image size while keeping the xml files around in the cache so they don't have to be downloaded again.

lukash: I'm not sure, at this stage the cache isn't used and the point here is to save the repetition of the dnf calls, so I guess it's a good idea. I added a `dnf clean all` command at the end.

PR #968, opened 2021-03-04
Errata applicability

j-mracek:


kontura:
Can I ask why do we need two versions of the `perl` modules in each stream?

The way I understand it the first versions of `perl` streams  (`5.24-1.module_el8+5182+8241aed2` and `5.26-1.module_el8+6437+6543afr6`) don't come into play anywhere, it will always select the newer version no?

kontura:
Thanks for adding the helpful commit message.
LGTM.

PR #970, opened 2021-03-05
Test that installonly_limit=1 is not allowed

m-blaha:
Requires: https://github.com/rpm-software-management/libdnf/pull/1155
Requires: https://github.com/rpm-software-management/dnf/pull/1740


pkratoch:
Can you also add a test that when `installonly_limit` is set to 1, kernel upgrade doesn't fail as shown in the bug?

m-blaha:
@pkratoch good point. I've added it.

PR #971, opened 2021-03-09
Make tests pass on both Fedora and RHEL

m-blaha:
Dear reviewer, if you refuse to merge this PR I would understand. This tests patching and workarounding the wording differences between Fedora and RHEL is gradually getting unsustainable.
We should branch the RHEL tests.


lukash:
> We should branch the RHEL tests.

That's 8.4, right? Shouldn't we just do that anyway? Any reason to put it off?

PR #972, opened 2021-03-09
Test that reposync does not re-download packages

m-blaha:
Requires: https://github.com/rpm-software-management/librepo/pull/232

https://bugzilla.redhat.com/show_bug.cgi?id=1931904


kontura:
The test suite passed with the librepo change merged, LGTM, thanks!

@janblazek these tests are adding a new dependency for `attr`.

PR #974, opened 2021-03-16
Add tests for handling of obsoletes

j-mracek:
https://bugzilla.redhat.com/show_bug.cgi?id=1818118

### DIFF
--- a/dnf-behave-tests/features/obsoletes.feature
+++ b/dnf-behave-tests/features/obsoletes.feature
@@ -31,6 +31,40 @@ Scenario: Install of obsoleting package, even though higher version than obsolet
         | install       | PackageA-Obsoleter-0:1.0-1.x86_64         |
 
 
+Scenario: Do not install of obsoleting package using upgrade command, when obsoleted package not on the system
+   When I execute dnf with args "upgrade PackageA-Obsoleter"
+   Then the exit code is 1
+    And Transaction is empty
+
+
+@bz1818118
+Scenario: Install of obsoleting package using upgrade command, when obsoleted package on the system
+  Given I execute dnf with args "install PackageE-0:1.0-1.x86_64"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                                   |
+        | install       | PackageE-0:1.0-1.x86_64                   |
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/features/repo-priorities.feature
```diff
@@ -19,6 +19,9 @@ Scenario: Install an RPM from the highest-priority repository
     And Transaction is following
         | Action        | Package                                   |
         | install       | flac-0:1.3.3-2.fc29.x86_64                |
+   When I execute dnf with args "upgrade"
+   Then the exit code is 0
+    And Transaction is empty
```
kontura, line 24: I don't think we should be testing obsoletes in the `repo-priorities.feature` but If it is somehow related to repo priorities please make a separate scenario.

dnf-behave-tests/features/obsoletes.feature
```diff
@@ -31,6 +31,40 @@ Scenario: Install of obsoleting package, even though higher version than obsolet
...
 
 
+Scenario: Do not install of obsoleting package, when obsoleted package not on the system
+   When I execute dnf with args "upgrade PackageA-Obsoleter"
+   Then the exit code is 1
+    And Transaction is empty
+
+
+@bz1818118
+Scenario: Install of obsoleting package, when obsoleted package on the system
```
kontura, line 41: Could you add to all the scenario titles that the install is done using the `upgrade` command? 
I think it would make it more obvious what is being tested here.
Maybe something like:
```Gherkin
Scenario: Install of obsoleting package via the upgrade command, when obsoleted package on the system
```

kontura (changes requested):

dnf-behave-tests/features/repo-priorities.feature
j-mracek: I am sorry, I put the change into the same commit. The code only extend the original test of priority. It ensures that after installing the latest package from priority repo, it did not upgrade it to package from repo that has higher priority value. I realized in the code that it is really easy to break the behaviour. Shall I split it in another commit?

kontura (approved):

kontura:
Thanks for the update.
LGTM.

The tests pass locally.

PR #976, opened 2021-03-17
Test listing module with summary that contains empty line

pkratoch:
Test for https://github.com/rpm-software-management/dnf/pull/1739

The only weird thing is that I could reproduce the issue only with additional new line in the module summary, not any multiline summary.


lukash:
@pkratoch I think this needs a rebase too just to fix the CI.

PR #978, opened 2021-03-22
Add plugin to manage exception after transaction resolution

inknos:
https://bugzilla.redhat.com/show_bug.cgi?id=1815327

### DIFF
--- a/dnf-behave-tests/features/dnf-transaction-exception.feature
+++ b/dnf-behave-tests/features/dnf-transaction-exception.feature
@@ -1,0 +1,24 @@
+@no_installroot
+Feature: Test exception when dnf tries to do a transaction resolved but not valid anymore
+
+Background: setup repository
+  Given I use repository "miscellaneous"
+    And I use repository "utility-plugins"
+    And I configure dnf with
+        | key          | value        |
+        | pluginpath   | /dnf-plugins |
+    And I do not disable plugins
+
+
+@bz1815327
+@bz1887293
+@bz1909845
+Scenario: Dnf fails to update a plugin if it does not exist after resolving the transaction
+  Given I successfully execute dnf with args "install dummy"
+    And I successfully execute dnf with args "install dnf-plugin-remove-dummy"
+   When I execute dnf with args "remove dummy"
+   Then the exit code is 1
+   And stderr is
+   """
+   Error: An rpm exception occurred: package not installed
+   """
--- a/dnf-behave-tests/fixtures/specs/utility-plugins/dnf-plugin-remove-dummy.spec
+++ b/dnf-behave-tests/fixtures/specs/utility-plugins/dnf-plugin-remove-dummy.spec
@@ -1,0 +1,30 @@
+Name: dnf-plugin-remove-dummy
+Version: 1
+Release: 1
+Summary: Plugin that removes the pkg dummy
+
+License: GPLv3+
+Url: None
+
+%description
+This package adds a plugin that removes the package dummy
+after the transaction gets resolved
+
+%install
+mkdir -p %{buildroot}/dnf-plugins/
+cat >> %{buildroot}/dnf-plugins/dnf-plugin-remove-dummy.py <<EOF
+import dnf
+import os
+
+class RemoveDummy(dnf.Plugin):
+    name = "remove-dummy"
+
+    def resolved(self):
+        print("RESOLVED HOOK: removing dummy")
+        os.system("rpm -e dummy")
+EOF
+
+%files
+/dnf-plugins/%{name}.py
+
+%changelog


dnf-behave-tests/features/dnf-transaction-exception.feature
```diff
@@ -0,0 +1,15 @@
...
+    And I configure dnf with
+        | key          | value                                 |
+        | pluginpath   | /dnf-plugins |
+    And I do not disable plugins
+
+
+Scenario: Dnf fails to update a plugin if it does not exist after resolving the transaction
+  Given I successfully execute dnf with args "install wilee"
+   When I execute dnf with args "remove update wilee"
+   Then the exit code is 0
```
m-blaha, line 15: As I understand the problem you are trying to reproduce I would add to `dnf-transaction` repository another package let's say `llama-1.0.0`.
Than I would change `wilee` plugin to call `rpm -e llama` instead.
And the Scenario would be:
```
Given I successfully execute dnf with args "install llama"
And I successfully execute dnf with args "install wilee"
When I execute dnf with args "remove llama"
Then the exit code is 0
```
Current of dnf should raise an exception "_rpm.error: package not installed"
Does it make sense?

dnf-behave-tests/fixtures/specs/dnf-transaction/wilee-1.spec
```diff
@@ -0,0 +1,28 @@
...
+License: GPLv3+
+Url: None
+
+%description
+This package adds a plugin that removes the package that installed it
+
+%install
+mkdir -p %{buildroot}/dnf-plugins/
+cat >> %{buildroot}/dnf-plugins/wilee-dnf-plugin.py <<HERE
+import dnf
```
m-blaha, line 15: `import os` missing

dnf-behave-tests/features/dnf-transaction-exception.feature
```diff
@@ -0,0 +1,15 @@
...
+  Given I use repository "dnf-transaction"
+    And I configure dnf with
+        | key          | value                                 |
+        | pluginpath   | /dnf-plugins |
+    And I do not disable plugins
+
+
+Scenario: Dnf fails to update a plugin if it does not exist after resolving the transaction
+  Given I successfully execute dnf with args "install wilee"
+   When I execute dnf with args "remove update wilee"
```
lukash, line 14: So IIUC this `wilee` package installs a plugin and then the test actually works on the plugin package, which seems to be an unnecessary "inception" complication that may be a bit confusing? Better perhaps to have one package with the plugin and another package that is being installed/upgraded.

Also, the command is `remove update wilee`, AFAICS it's a mistake, this calls `remove` on two packages: `update` and `wilee`.

dnf-behave-tests/features/dnf-transaction-exception.feature
lukash, line 15: You should probably do more on the result checking front, i.e. making sure what you expect to have transpired really has transpired?

dnf-behave-tests/fixtures/specs/dnf-transaction/wilee-1.spec
```diff
@@ -0,0 +1,28 @@
...
+
+License: GPLv3+
+Url: None
+
+%description
+This package adds a plugin that removes the package that installed it
+
+%install
+mkdir -p %{buildroot}/dnf-plugins/
+cat >> %{buildroot}/dnf-plugins/wilee-dnf-plugin.py <<HERE
```
lukash, line 14: `EOF` is conventionally used as the marker instead of `HERE`.

dnf-behave-tests/fixtures/specs/dnf-transaction/wilee-1.spec
```diff
@@ -0,0 +1,28 @@
...
+import dnf
+
+class Wilee(dnf.Plugin):
+    name = "wilee"
+
+    def resolved(self):
+        print("RESOLVED")
+        os.system("rpm -i wilee")
```
lukash, line 22: This seems to be doing an install, not a remove :slightly_smiling_face:

lukash (commented):
Just some quick general feedback...

dnf-behave-tests/features/dnf-transaction-exception.feature
m-blaha: The test should check that "parsed dnf output transaction" contains `llama` package removal and the rpm transaction is empty (not sure we have a step for it)? Or what is the proposed fix in the dnf? I mean should it return 0 / nothing to do (because the package that was expected to be removed was already removed from outside), or should it return an error that system state during the transaction does not correspond to system state in transaction resolution time?

dnf-behave-tests/features/dnf-transaction-exception.feature
inknos: regarding `remove`, `update`. I was distracted my bad. anyway the exception should trigger both with `remove` and `update` command

dnf-behave-tests/features/dnf-transaction-exception.feature
inknos: it makes sense I am trying some changes now. Thank you

lukash: Well, good question I guess :slightly_smiling_face: I am not sure at all, but I would lean towards the error state. I'm not sure about the real-world use case semantic details (what is the practical situation of rpm being called to remove the package at the same time dnf is running), but it is something unexpected and we are assuming a simple scenario. If there's actually something a bit more complex resulting in this, the user should be notified of the unusual situation, rather than risking it being ignored.

m-blaha: That's also what I would prefer.
Just return nice error message instead of current ugly traceback.

dnf-behave-tests/fixtures/specs/dnf-transaction/wilee-1.spec
inknos: Fixed: it was a spelling issue :)

dnf-behave-tests/fixtures/specs/dnf-transaction/wilee-1.spec
```diff
@@ -0,0 +1,29 @@
...
+import dnf
+import os
+
+class Wilee(dnf.Plugin):
+    name = "wilee"
+
+    def resolved(self):
+        print("RESOLVED")
+        os.system("rpm -e llama")
+HERE
```
m-blaha, line 24: You need to replace `HERE` by `EOF` also here.

dnf-behave-tests/features/dnf-transaction-exception.feature
```diff
@@ -0,0 +1,16 @@
...
+
+Background: setup repository
+  Given I use repository "dnf-transaction"
+    And I configure dnf with
+        | key          | value                                 |
+        | pluginpath   | /dnf-plugins |
+    And I do not disable plugins
+
+
+Scenario: Dnf fails to update a plugin if it does not exist after resolving the transaction
```
m-blaha, line 12: For the context - please reference the bug number.

dnf-behave-tests/features/dnf-transaction-exception.feature
inknos: @lukash @m-blaha  thanks for the help, now the test is fine and is able to reproduce and test the error. the solution in my opinion is the same as Marek's. simply avoid the crash and return a dnf error. I am preparing the code now for the test and dnf as well

dnf-behave-tests/fixtures/specs/dnf-transaction/wilee-1.spec
```diff
@@ -0,0 +1,29 @@
+Name: wilee
```
lukash, line 1: Alright, one more note, can you actually name this package e.g. `dnf-plugin-remove-llama`, as it is not a random package, we can actually give it a meaningful description.

dnf-behave-tests/fixtures/specs/dnf-transaction/wilee-1.spec
```diff
@@ -0,0 +1,29 @@
...
+Version: 1.0
+Release: 1.fc29
+Summary: Package that adds a plugin that removes package
+
+License: GPLv3+
+Url: None
+
+%description
+This package adds a plugin that removes the package that installed it
```
lukash, line 10: And also, the description is no longer accurate :slightly_smiling_face:

lukash (commented):
Sorry, I found this as a pending review, forgot to post it :(

dnf-behave-tests/features/dnf-transaction-exception.feature
```diff
@@ -0,0 +1,23 @@
+@no_installroot
+Feature: Test exception when dnf tries to do a transaction resolved but not valid anymore
+
+Background: setup repository
+  Given I use repository "dnf-transaction-exception"
```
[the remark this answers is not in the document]
inknos: > I wonder if we want to make this a repository dedicated to these utility plugins...
Good idea

> What about naming the repository e.g. utility-plugins as a place for these helper plugins in case we have more of these in the future?
`utility-plugins` is surely better

> I found the watermelons grin in fixtures/specs/plugins-callbacks/*
`fixtures/specs/plugins-callbacks/*` is actually the folder from which I took the example for this code :)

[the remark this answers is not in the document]
inknos: The name for the repo is now, `dnf-transaction-exception -> utility-plugins`: do you think the name of the feature-file should be changed?
Maybe now I am overthinking the problem but...
Should the feature-file be moved in a subdir as well such as `features/utility-plugins/dnf-transaction-exception.feature`?

lukash:
@inknos took the liberty to rebase the PR (to fix the CI) and also added a FIXUP commit with some changes to spare you one more round of my nitpicks :slightly_smiling_face: 

I dropped the `llama` package, that shouldn't go into the `utility-plugins` repo as that's not a utility plugin, and I think it's not necessary to add a repo for a simple package, I used a package from the `miscellaneous` repo. Did some renaming to match and a few cosmetic touchups.

If you like it, please squash the FIXUP commit and I'm ready to merge this (the CI won't pass as there's a related dnf change, but I tested it locally).

inknos:
I squashed the commits, I think now it looks very consistent.
Thank you!!

lukash:
Thanks! Merging.

PR #980, opened 2021-03-23
Test repo gets expired when package has incorrect checksum

pkratoch:
https://bugzilla.redhat.com/show_bug.cgi?id=1941959

I am not yet sure if this test is good enough to test the bug. I remember the issue was not reproducible with local repositories and it should work with the `dnf-ci-gpg repo` used as http, but I can't test it without the bug fix.


lukash:
@pkratoch can you rebase please?

PR #984, opened 2021-04-14
More cleanups

lukash:
A few more cleanups in the test setup, this should be all of it.

The last two commits are up for discussion, I've gone the more radical route, if someone needs the removed functionality (Dockerfiles for multiple Fedoras, the TYPE argument in Dockerfiles), I can put it back.


dnf-behave-tests/common/lib/os_version.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+import operator
+
+from behave.tag_matcher import ActiveTagMatcher
```
m-blaha, line 8: Unused imports

dnf-behave-tests/common/lib/os_version.py
```diff
@@ -0,0 +1,32 @@
...
+                os_version = v.strip(' \t\n"')
+
+    # cut off any minor versions - in RHEL the version is e.g. 8.3, but we only want 8
+    os_version = os_version.split(".")[0]
+
+    # treat centos as RHEL in context of scenario tag matching
+    if os_id == "centos":
+        os_id = "rhel"
+
+    return os_id + "__" + os_version
```
m-blaha, line 32: This could be simplified using `python3-distro` package (available on both fedora and rhel):
```
import distro
def detect_os_version():
    os_id = distro.id()
    if os_id == "centos": os_id = "rhel"
    os_version = distro.major_version()
    return os_id + "__" + os_version
```
```

lukash: Much better, thanks :slightly_smiling_face: I did wonder if there's something, but thought it might not be on RHEL...

dnf-behave-tests/common/lib/os_version.py
lukash: Fixed.

PR #986, opened 2021-04-15
Rename dnf-behave-tests/features to dnf-behave-tests/dnf

lukash:
I think it's ok to merge and then QE can update their script, but let's wait for @janblazek to confirm.

After that, let's not hesitate to minimize the conflicts.

I think the CI will not pass, but here's my test: https://github.com/lukash/ci-dnf-stack/actions/runs/752025487


m-blaha (approved):
Thanks, the rename makes sense.

janblazek:
I confirm that this rename is not a problem for running the ci-dnf-stack on RHEL.

PR #988, opened 2021-04-28
Update the path of the -d/--devel argument for correct mount

kontura:
This was updated in Dockerfile: https://github.com/rpm-software-management/ci-dnf-stack/pull/984


lukash:
Oops, thanks!

PR #989, opened 2021-05-06
Add support for drop-in repos.d directory for the container

lukash:


janblazek:
I confirm that this works even for RHEL.

lukash:
I've added a commit making libfaketime an unconditional dep (it is not provided in https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/test-utils/ for CentOS and EPEL 8). Should be the last piece.

Btw. do we want to bikeshed whether to name the new directory `repos.d` or just `repos`? :slightly_smiling_face:

janblazek:
I don't really care about repos.d vs repos. repos.d at least somewhat indicates that it's content should be .repo files.

PR #991, opened 2021-05-11
Test groups without packagelist and merging environments

kontura:
Test that we can handle groups without `<packagelist>` element.

Also tests some changes from: https://github.com/rpm-software-management/dnf/pull/1764

### DIFF
--- a/dnf-behave-tests/dnf/comps-group.feature
+++ b/dnf-behave-tests/dnf/comps-group.feature
@@ -324,3 +324,94 @@ Scenario: Install an environment with empty name
         | env-install   | <name-unset>                      |
         | group-install | Test Group                        |
         | install-group | test-package-1.0-1.fc29.noarch    |
+
+
+@not.with_os=rhel__eq__8
+Scenario: List and info a group with missing packagelist
+  Given I use repository "comps-group-merging"
+   When I execute dnf with args "group list"
+   Then the exit code is 0
+    And stdout is
+       """
+       <REPOSYNC>
+       Available Environment Groups:
+          <name-unset>
+       Available Groups:
+          Test Group
+       """
+   When I execute dnf with args "group info test-group"
+   Then stdout is
+       """
+       <REPOSYNC>
+       Group: Test Group
+        Description: Test Group description updated.
+       """
+
+
+Scenario: Install a group with empty packagelist
+  Given I use repository "comps-group-merging"
+   When I execute dnf with args "group install test-group"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package       |
+        | group-install | Test Group    |
+
+
+@not.with_os=rhel__eq__8
+Scenario: Merge groups when one has empty packagelist
+  Given I use repository "comps-group"
+    And I use repository "comps-group-merging"
+   When I execute dnf with args "group install test-group"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                           |
+        | group-install | Test Group                        |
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/dnf/comps-group.feature
```diff
@@ -324,3 +324,65 @@ Scenario: Install an environment with empty name
...
+       """
+
+Scenario: Merge environment with missing names containg a group with missing name
+  Given I use repository "comps-group"
+    And I use repository "comps-group-merging"
+   When I execute dnf with args "group info no-name-env"
+   Then stdout is
+       """
+       <REPOSYNC>
+       Environment Group: None
```
pkratoch, line 384: In other places, we are using `<name-unset>` for groups/environments without name. Could it be used here as well to be consistent?

kontura: I have looked into this but to my understanding **not** all other places are using `<name-unset>`, for example the `info` command on a group without a name also returns `None` (here it could be seen by changing it to `When I execute dnf with args "group info no-name-group"`).

I was mostly trying to fix the traceback that occurs in the test without my related dnf patch.

And I have to say I am not very happy with it, testing in the outputs for `None` and replacing it with `<name-unset>` like we do in `dnf/cli/output.py` is kind of a hack and we should keep it to a minimum.

But you are right it would be more consistent if we printed `<name-unset>` everywhere.

(Also I have randomly stumbled into this and it might not be a real problem for anyone since every group/environment should have a name.)

kontura: Maybe @lukash has some advice since he was dealing with it before?

lukash: I don't know. The situation is rather far from optimal:
1. The group/env ID should be used consistently as the main identifier in the outputs
2. The name of groups and environments should be mandatory (from what I've heard we're not even supposed to print a warning when a name is missing)

>  And I have to say I am not very happy with it, testing in the outputs for None and replacing it with <name-unset> like we do in dnf/cli/output.py is kind of a hack and we should keep it to a minimum.

What's the hack here if I may ask? The name being `None` case needs to be handled, which means you need to test it and then do something. The bad thing here is the check is copy-pasted in multiple places, I've "only" put it in two places myself, as my excuse, just looking at the commit it appears it's a different Group object each time. That's another issue with the code. Looking at your patch, it seems you are working with yet another struct (in https://github.com/rpm-software-management/dnf/pull/1764/commits/84c76e674bb447b68c9bc2c6e3f96c3e4ee20af3), judging by the different name of the `name` attribute.

For the case above, not sure where's the code that prints the `None`, it should likely be handled in the same way... not sure if we can make the code look half decent in the process.

kontura: > What's the hack here if I may ask? The name being `None` case needs to be handled, which means you need to test it and then do something. The bad thing here is the check is copy-pasted in multiple places,

It felt very hackish when I tried adding it to other places to cover all the `None` outputs, but I guess you are right its more the fact that it is copied than the check it self.

Ok I will try to cover all the places (I should probably define some constant for `<name-unset>` or helper function or something).

pkratoch: Thanks for your work. I just realized that "@not.with_os=rhel__eq__8" is needed for the tests that fail without the dnf PR. Can you please add it? After that, I can merge it.

kontura: @pkratoch added. :+1:

PR #994, opened 2021-05-13
Implement steps for dnfdaemon

inknos:


Dockerfile
```diff
@@ -39,13 +39,19 @@ RUN set -x && \
     dnf -y builddep /opt/ci/dnf-behave-tests/requirements.spec && \
     pip3 install -r /opt/ci/dnf-behave-tests/requirements.txt
 
+RUN set -x && \
+    dnf -y install dbus-daemon python3-dbus polkit
```
lukash, line 43: Not here anymore, please. Add the deps to `dnf-behave-tests/requirements.spec` (see the step above).

dnf-behave-tests/container-test-wrapper
```diff
@@ -21,7 +21,7 @@ elif [ "$RESERVE" == "-R" ]; then
 fi
 
 TEST_EXIT=0
-behave $DEFINES "$@" || TEST_EXIT=$?
+( ./run-dbus;  behave $DEFINES "$@" ) || TEST_EXIT=$?
```
lukash, line 24: I think we should run the dbus and polkit processes conditionally (only if we're actually running daemon tests).

It would be nice to have some control over the processes, e.g. check if they failed to start, but I suspect with them forking to the background we can't do anything. Or use some process manager, I just very quickly found this: http://garywiz.github.io/chaperone/index.html

Thinking about it, maybe we could "manage" the processes ourselves from inside behave? That is, run them and make sure they're fine for the duration of the tests.

dnf-behave-tests/dnf/dnfdaemon/repoinfo.feature
```diff
@@ -0,0 +1,65 @@
...
+Config file         :
+URLs                :
+  Base URL          : file:///opt/ci/dnf-behave-tests/fixtures/repos/dnf-ci-thirdparty-updates
+  Mirrorlist        :
+  Metalink          :
+GPG                 :
+  Keys              :
+  Verify repodata   :
+  Verify packages   :
+"""
```
lukash, line 65: I think we'll ideally want to reuse the existing scenarios for testing the daemon too, though how we deal with the different outputs is beyond me... (we need to discuss this)

dnfdaemon-server/dbus/dnfdaemon-server.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=DNF daemon
+
+[Service]
+Type=dbus
+BusName=org.rpm.dnf.v0
+User=root
+ExecStart=/usr/bin/dnfdaemon-server
```
lukash, line 8: Hm, I might be missing something but it seems the files in the `dnfdaemon-server` directory aren't used? Is this a working PoC? (I haven't tried to run it :grin:)

dnf-behave-tests/dnf/dnfdaemon/repolist.feature
```diff
@@ -0,0 +1,15 @@
+@not.with_os=rhel__ge__7
+@no_installroot
+
+Feature: Test dnfdaemon-client repolist
+
+Background: Configure repositories and check if daemon is up
+  Given I use repository "dnf-ci-fedora"
+    And I use repository "dnf-ci-fedora-updates"
+    And dnfdaemon-server is up and running
```
m-blaha, line 9: I would prefer not run dnfdaemon-server in a step - just to be able to share the scenarios with microdnf which does not need it.
I was thinking about using some decorator like `@dnfdaemon` for scenarios suitable for daemon and run the server from environment.py:before_scenario() callback.

dnf-behave-tests/dnf/dnfdaemon/repoinfo.feature
m-blaha: Well the daemon and microdnf are currently sharing some of outputs and the goal is to share all of them.

dnf-behave-tests/dnf/environment.py
```diff
@@ -211,6 +215,37 @@ def get_microdnf_cmd(self, context):
 
         return result
 
+    def get_dnf5_cmd(self, context):
+        if context.dnf.dnfdaemon_mode:
+            result = ["dnfdaemon-client"]
```
m-blaha, line 220: This breaks `--command` option in dnfdaemon_mode. I'd prefer handle it consistently with microdnf5 and rely on `--command=dnfdaemon-client`.

dnf-behave-tests/dnf/environment.py
```diff
@@ -247,19 +282,28 @@ def before_scenario(context, scenario):
...
                              no_installroot='no_installroot' in scenario.effective_tags,
-                             dnf5_mode = dnf5_mode)
+                             dnf5_mode = dnf5_mode,
+                             dnfdaemon_mode = dnfdaemon_mode)
 
     write_config(context)
 
+    if dnfdaemon_mode == True:
+        p_dbus_daemon = subprocess.Popen(['/usr/bin/dbus-daemon', '--system'])
+        p_polkitd = subprocess.Popen(['/usr/lib/polkit-1/polkitd', '&'])
```
m-blaha, line 305: We should either terminate the processes after the scenario or detect that they are already running and do not try to launch them again. Otherwise I would expect strange things happening once more than one scenario is tagged with `@dnfdaemon`

m-blaha (changes requested):

dnf-behave-tests/dnf/repolist.feature
```diff
@@ -21,8 +22,9 @@ Scenario: Repolist without arguments
     And stdout does not contain "dnf-ci-thirdparty\s+dnf-ci-thirdparty"
 
 
+@dnfdaemon
 Scenario: Repolist with "enabled"
-   When I execute dnf with args "repolist enabled"
+   When I execute dnf with args "repolist --enabled"
```
m-blaha, line 27: I'd like to preserve also original scenario ("repolist enabled"). Although we do not want this syntax supported in dnf5/dnfdaemon, it's still supported in dnf4.
We can turn this scenario into `Scenario Outline` to test both possibilities (enabled / --enabled) and only tag `--enabled` variant with `@dnfdaemon`:

```
Scenario Outline: Repolist with "<argument>"
   When I execute dnf with args "repolist <argument>"
   ...
Examples: positional argument for yum compatibility
    | argument      |                     
    | enabled       |                     
@dnfdaemon
Examples: named argument
    | argument      |
    | --enabled     |
```

Or create a new scenario with `--enabled` which is probably better readable.

PR #997, opened 2021-05-20
Test that versionlock can lock packages with minorbump in release

m-blaha:
see https://bugzilla.redhat.com/show_bug.cgi?id=1961217

Requires: https://github.com/rpm-software-management/dnf-plugins-core/pull/430

### DIFF
--- a/dnf-behave-tests/fixtures/specs/miscellaneous/minorbump-0:1.0-1.1.spec
+++ b/dnf-behave-tests/fixtures/specs/miscellaneous/minorbump-0:1.0-1.1.spec
@@ -1,0 +1,16 @@
+Name:           minorbump
+Epoch:          0
+Version:        1.0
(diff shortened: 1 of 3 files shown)


dnf-behave-tests/fixtures/specs/miscellaneous/minorbump-0:1.0-1.1.spec
```diff
@@ -0,0 +1,16 @@
+Name:           minorbump
```
lukash, line 1: I think we need to sort out this `miscellaneous` situation :)

m-blaha: Yeah :). I basically did not want to create new repos when I need only one package (although a special one)

PR #999, opened 2021-05-21
Test for non UTF-8 files in the package

m-blaha:
Require: https://github.com/rpm-software-management/libdnf/pull/1237
RhBug: https://bugzilla.redhat.com/show_bug.cgi?id=1893176

### DIFF
--- a/dnf-behave-tests/dnf/cache.feature
+++ b/dnf-behave-tests/dnf/cache.feature
@@ -5,7 +5,7 @@ Feature: Tests for cache
 @destructive
 @no_installroot
 Scenario: Do not error out when fail to load/store expired_repos cache
-  Given I use repository "miscellaneous"
+  Given I use repository "simple-base"
     And I create file "/tmp/dnf/expired_repos.json" with
         """
         """
@@ -15,12 +15,12 @@ Scenario: Do not error out when fail to load/store expired_repos cache
    Then the exit code is 0
     And stdout is
         """
-        dummy-1:1.0-1.src
-        dummy-1:1.0-1.x86_64
-        weird-1:1.0-1.src
-        weird-1:1.0-1.x86_64
-        weird-1:2.0-1.src
-        weird-1:2.0-1.x86_64
+        dedalo-signed-0:1.0-1.fc29.src
+        dedalo-signed-0:1.0-1.fc29.x86_64
+        labirinto-0:1.0-1.fc29.src
+        labirinto-0:1.0-1.fc29.x86_64
+        vagare-0:1.0-1.fc29.src
+        vagare-0:1.0-1.fc29.x86_64
         """
     And stderr contains lines
         """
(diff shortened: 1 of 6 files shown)


dnf-behave-tests/fixtures/specs/miscellaneous/non_utf_filenames-1.0-1.spec
```diff
@@ -0,0 +1,27 @@
...
+BuildArch:      noarch
+
+%description
+Dummy.
+
+%install
+mkdir -p %{buildroot}/non-utf8-files/
+pushd %{buildroot}/non-utf8-files/
+  touch hex-80-�-
+  touch hex-a0-�-
```
lukash, line 20: Might just be me but how does this create filenames that are invalid utf-8? Maybe add a comment?

dnf-behave-tests/dnf/cache.feature
```diff
@@ -11,16 +11,12 @@ Scenario: Do not error out when fail to load/store expired_repos cache
...
+   When I execute dnf with args "repoquery --setopt=cachedir=/tmp/dnf/ dummy" as an unprivileged user
    Then the exit code is 0
     And stdout is
         """
         dummy-1:1.0-1.src
         dummy-1:1.0-1.x86_64
-        weird-1:1.0-1.src
-        weird-1:1.0-1.x86_64
-        weird-1:2.0-1.src
-        weird-1:2.0-1.x86_64
```
lukash, line 23: The original intent with the miscellaneous repo was to not add "random packages" into it, but to have a very simple repo that will not change. Granted it's misnamed and the name suggests the purpose for which you've used it.

I'm not against adding the rpm to it, but we should have a repo to which we don't add packages and can be used for the majority of tests that just needs "something"...

m-blaha: Oh, OK. I thought that miscellaneous was supposed for those kind of "special" packages (when you do not want to create a repo for them), and "simple-base" and "simple-updates" repos for those tests that just need any repo.

dnf-behave-tests/fixtures/specs/miscellaneous/non_utf_filenames-1.0-1.spec
m-blaha: It's not visible on the github preview. There is `0x80` byte in the first file (where github shows an euro sign) and `0xa0` byte on the second line.

lukash: Can you put it in a comment then? Even if the actual hex value is a part of the name just before the invalid bad byte, it's just not really visible in a text editor (depending on how it even displays the invalid bytes). Just clarify what's going on there...

dnf-behave-tests/dnf/cache.feature
lukash: Yeah that makes sense, I missed that there are simple-base and simple-updates now.

Wouldn't it make more sense to switch this scenario to use simple-base then? It doesn't have a special package in the miscellaneous repo.

m-blaha: I switch the test to simple-base repo.

dnf-behave-tests/fixtures/specs/miscellaneous/non_utf_filenames-1.0-1.spec
m-blaha: Done.

lukash:
Thanks!

PR #1000, opened 2021-05-21
Dnf5 support

lukash:
This adds support for running scenarios marked with the `@dnf5` tag in "dnf5 mode", which is stored in DNFContext and allows to conditionally handle differences between dnf4 and dnf5 in behave steps.

### DIFF
--- a/dnf-behave-tests/common/output.py
+++ b/dnf-behave-tests/common/output.py
@@ -11,20 +11,43 @@ from common.lib.diff import print_lines_diff
 from common.lib.text import lines_match_to_regexps_line_by_line
 
 
-def handle_reposync(expected, found):
+sync_line_dnf4 = re.compile(r".*[0-9.]+ +[kMG]?B/s \| +[0-9.]+ +[kMG]?B +[0-9]{2}:[0-9]{2}")
+bottom_line_dnf4 = re.compile(r"Last metadata expiration check: .*")
+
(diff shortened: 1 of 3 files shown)


dmach:
@lukash it would be good to replace <REPOSYNC> with something a bit less misleading (reposync is also a dnf command).
Maybe <REPODATA_SYNC> would be a better choice.

m-blaha:
Let's rather leave this to a separate PR. The change would be needed in lot of files and is kind of unrelated to this PR.

lukash:
@dmach I'm not opposed, but yeah, it's in a lot of places, separately might be better.

inknos:
I agree with your considerations and since the PR looks good to me, and I would need it to push more changes related to the daemon, I would merge it. However, what is the issue with the failing test?

About createrepo_c
```
  createrepo_c/zchunk.feature:76  create repository with zck metadata usign dictionaries
  createrepo_c/zchunk.feature:125  create repository with zck and dictionary metadata with bad package
```
The scenarios are not related. So, what should we do?

kontura:
@inknos we can ignore the createrepo_c tests, its a zchunk issue.

dmach:
FYI: This should help with easier parsing the transaction table, making it similar to DNF 4: https://github.com/rpm-software-management/libdnf/pull/1254

PR #1003, opened 2021-06-02
Do no hardcode user for ghcr.io use secret instead

kontura:
When the user was hardcoded to rpm-software-management everytime someone
had this workflow in master branch (of their ci-dnf-stack fork) this
workflow ran as scheduled and failed when pushing to
ghcr.io/rpm-software-management/dnf-ci-host.

Now if you define your own secret GHCR_USER the image gets pushed to
ghcr.io/${GHCR_USER}/dnf-ci-host and if no user is defined the workflow
does nothing.

This also means we need to have the user secret defined for
https://github.com/rpm-software-management/ci-dnf-stack

### DIFF
--- a/.github/workflows/ci-host.yml
+++ b/.github/workflows/ci-host.yml
@@ -11,10 +11,19 @@ jobs:
     steps:
       - name: Build the container
         run: |
-          DATE=`date "+%F"`
-          IMAGE="ghcr.io/rpm-software-management/dnf-ci-host"
+          # hack: Github replaces secrets with *** in the whole output.
+          # If there's a comment (#) at the end of the secret
+          # (e.g. "rpm-sofware-management #"), this will clean it up and
+          # since it is no longer the whole secret being printed, Github
+          # won't hide it anymore.
+          GHCR_USER=${{secrets.GHCR_USER}}
 
-          echo ${{secrets.GITHUB_TOKEN}} | docker login ghcr.io -u rpm-software-management --password-stdin
+          if [ -z "$GHCR_USER" ]; then exit 0; fi
+
+          DATE=`date "+%F"`
+          IMAGE="ghcr.io/${GHCR_USER}/dnf-ci-host"
+
+          echo ${{secrets.GITHUB_TOKEN}} | docker login ghcr.io -u ${GHCR_USER} --password-stdin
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-host.yml
```diff
@@ -11,12 +11,20 @@ jobs:
     steps:
       - name: Build the container
         run: |
-          DATE=`date "+%F"`
-          IMAGE="ghcr.io/rpm-software-management/dnf-ci-host"
+          # hack: Github replaces secrets with *** in the whole output (even in
+          # e.g. Copr URLs printed by rpm-gitoverlay). If there's a comment (#)
```
lukash, line 15: The mention of Copr URLs is not really relevant in this context, maybe leave it out? Seems like it could just add to confusion instead of being helpful.

.github/workflows/ci-host.yml
```diff
@@ -11,12 +11,20 @@ jobs:
...
+          # at the end of the secret (e.g.  "rpm-sofware-management #"), this
+          # will clean it up and since it is no longer the whole secret being
+          # printed, Github won't hide it anymore.
+          GHCR_USER=${{secrets.GHCR_USER}}
 
-          echo ${{secrets.GITHUB_TOKEN}} | docker login ghcr.io -u rpm-software-management --password-stdin
+          if [ -n "$GHCR_USER" ]; then
```
lukash, line 21: Why not just do an early `exit 0` instead of indenting the rest of the script into an `if`?

lukash (commented):
Thanks. I've added the secret to the repo. Just a couple of notes...

.github/workflows/ci-host.yml
kontura: Ok I have removed that part of the comment.

.github/workflows/ci-host.yml
kontura: I wasn't sure I can call `exit` in the workflow, but I see now its already used in the `setup-ci`. :+1: 

Changed it.

lukash:
Thanks. Merging, I haven't bothered testing, it looks like it should work :slightly_smiling_face: If it doesn't, we'll have a day old image used for the CI, so no big deal.

PR #1007, opened 2021-06-10
Parse DNF 5 transaction table

dmach:


dnf-behave-tests/dnf/steps/lib/dnf.py
```diff
@@ -96,6 +96,12 @@ def find_transaction_table_begin(lines):
 
         return i + 2
 
+    # DNF 5 requires a different match
+    trans_start_re = re.compile(r"Package +Arch +Version +Repository +Size")
+    for i in range(0, len(lines) - 1):
+        if trans_start_re.match(lines[i]):
+            return i + 1
+
```
lukash, line 104: Could you fork the code based on the `context.dnf.dnf5_mode` variable so that it's clearly separated on what is used for dnf 4 and what is used for dnf 5?

dnf-behave-tests/dnf/steps/lib/dnf.py
```diff
@@ -109,6 +115,16 @@ def find_transaction_table_end(lines):
...
             return i - 1
+
+    # DNF 5 requires a different match
+    for i in range(0, len(lines)):
+        if not lines[i].strip():
+            # empty line indicates the end of the transaction table
+            return i
+        if lines[i].startswith("Transaction Summary:"):
+            # "Transaction Summary" follows the transaction table
+            return i
```
lukash, line 126: So can both of the above mark the end of the table? It should only be one of them? We should be more strict when parsing the table and not match more than one semi-arbitrary string.

dmach: All right, let's make it an empty line then.
If `Transaction Summary:` immediately follows the transaction table, we should fix the code by adding the empty line.

dnf-behave-tests/dnf/steps/lib/dnf.py
dmach: Sure. Changed.

PR #1009, opened 2021-06-16
Add github action to refresh the *.pot file from source

m-blaha:
We are using ci-dnf-stack repo as a library of actions.

### DIFF
--- a/.github/actions/l10n-update-pot/action.yml
+++ b/.github/actions/l10n-update-pot/action.yml
@@ -1,0 +1,41 @@
+name: Update *.pot files
+
+# This action expects to have these repositories checked out:
+#   in directory `l10n` is the repository with weblate translations (e.g. dnf-l10n)
+#   in directory `source` is repository with sources (e.g. dnf)
+
+inputs:
+  branch:
+    required: true
+    description: The branch from source repository, which is a directory in the l10n repository in which translations for the branch are stored
+
+runs:
+  using: "composite"
+  steps:
+    - name: Generate *.pot files
+      shell: bash
+      run: |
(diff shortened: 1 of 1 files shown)


.github/actions/l10n-update-pot/action.yml
```diff
@@ -0,0 +1,41 @@
+name: Update *.pot files
+
+# This action expects to have these repositories checked out:
+#   in directory `l10n` is the repository with weblate translations (e.g. dnf-l10n)
+#   in directory `source` is repository with sources (e.g. dnf)
+
+inputs:
+  branch:
+    required: true
+    description: Subdirectory where to store translations
```
lukash, line 10: The argument is called `brach`, but the description talks about a subdirectory, so this might need a bit more explaining in the description? e.g.
> The branch from source repository, which is a directory in the l10n repository in which translations for the branch are stored

.github/actions/l10n-update-pot/action.yml
```diff
@@ -0,0 +1,41 @@
...
+
+        # copy new pot file to l10n repository
+        for POTFILE in ${POTFILES}; do
+          cp source/${POTFILE} l10n/${{inputs.branch}}/
+        done
+
+        # commit the changes
+        pushd l10n > /dev/null
+          git config --local user.email "packaging-team-maint@redhat.com"
+          git config --local user.name "Github action"
```
lukash, line 37: I wouldn't use this name specific to GHA, what about "Automatic Translations Update" or something similar?

m-blaha:
Thanks @lukash, I've used your suggestions.

PR #1012, opened 2021-06-25
Add repository and tests for "installonlypkgs" configuration option

jrohel:
PR also changes the `diff_rpm_lists` and `check_rpmdb_transaction` functions.

The previous version of `diff_rpm_lists` function used `RPM.is_installonly`method, which contained hardcoded list of "installonly" packages. This aproach was unusable for testing the "installonlypkgs" option, because the same package is used as both "installonly" and as "not-installony" (depending on the value of the "installonlypkgs" option tested).
The new version of `diff_rpm_lists` function still uses information about "installonly" packages, but "installonly" packages are automaticaly detected. Detection uses the assumption that multiple versions of a package with the same name and architecture on the system mean an "installonly" package.

There is a special case when "installonly" package is removed and another version is installed. In this case, it is not an "installonly" package for automatic detection, and the `diff_rpm_lists` function returns upgrade or downgrade. So, the check_rpmdb_transaction` function was extended to deal with this.

### DIFF
--- a/dnf-behave-tests/dnf/installonlypkgs.feature
+++ b/dnf-behave-tests/dnf/installonlypkgs.feature
@@ -1,0 +1,60 @@
+Feature: Test upgrading installonly packages - "installonlypkgs" configuration option
+
+
+Background: Install one version of each test package
+  Given I use repository "installonly"
+   When I execute dnf with args "install installonlyA-1.0 installonlyB-1.0 kernel-core-4.18.16"
+   Then the exit code is 0
+     And Transaction is following
+        | Action        | Package                               |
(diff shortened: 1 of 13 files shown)


m-blaha:
I know the kernel specs are just copy of existing ones, but still - the specs seem to be unnecessarily big and complicated. There is much smaller one in `specs/dnf-ci-priority-1/kernel-4.18.16-300.fc29.spec` file - would that serve your test case?

jrohel:
@m-blaha Thanks for the tip. I used smaller kernels from the "dnf-ci-priority-1" and "dnf-ci-priority-2".

m-blaha:
Thanks, LGTM.

PR #1014, opened 2021-07-07
Fix repo metadata download matching for dnf5

m-blaha:


lukash:
Hm, Does the usage of the "i" units (e.g. KiB instead of KB) depend on something? I think the original regex worked for me...

m-blaha:
I didn't find any condition in the code. It should use `progressbar/widgets/common.cpp:format_size` which in turn uses `libdnf::cli::utils::units::format_size` - and here are only "i" units defined.
It might work for you in case that only base B/s units were used.

lukash:
Okay, might be... thanks!

PR #1022, opened 2021-08-03
[dnf5] add dnfdaemon folder for specifi scenarios

inknos:


inknos:
@m-blaha right, so, not sure if everything will explode with this pr but the intent is to disable `@destructive` partially when the tests are run with `@dnf5daemon`.
The tests should be skipped when run wit `--no-destructive` but scenarios should be run at once when both `@destructive` and `@dnf5daemon` are present.

inknos:
I just realized that maybe we would like to do change the tags with the opposite logic.

I am thinking of a real use case scenario when we want to tag pre-existing scenarios with `@dnf5daemon` to test if `dnf5daemon-client` runs fine. Currently this would require marking each scenario with `@destructive`. Not good.

I will work on it again with the opposite logic. 🤔

The target is to have one tag, `@dnf5daemon`
Scenarios marked with it will do two things. they will always run one feature file at time to optimize the initial polkitd initialization time and will be skipped if `--non-destructive` is used. So `@dnf5daemon` will be a destructive label on its own.

PR #1025, opened 2021-08-06
Add options to provide aditional arguments to build and run containers

pkratoch:


container-test
```diff
@@ -329,6 +335,9 @@ class BehaveRunner(object):
 
         command = self.docker_bin + ['run', '--rm'] + self.param_tty + self.volumes
 
+        if self.command_line_args.container_args:
+            command += self.command_line_args.container_args.split(' ')
```
lukash, line 339: This will break on any quoted argument with spaces... Would it work if you don't split?

pkratoch: Right... And it wouldn't work without the split. So, I changed it to use append. It means the option has to be specified multiple times, but I suppose there can't be used any separator, because it could be also contained in the argument.

lukash: Yeah, you'd have to parse any quotation then... we don't really want that :slightly_smiling_face: How about just naming the argument "--contariner-arg" since it can now only contain one and you specify it multiple times? And adjust the help text too. If you're in a hurry feel free to merge yourself after the change.

pkratoch: Ok, renamed. I only left the destination variable in the plural, because it contains the resulting list. Thanks for your review.

PR #1027, opened 2021-08-16
createrepo_c: `--update` alone can remove modular metadata

kontura:


kontura:
Requires: https://github.com/rpm-software-management/createrepo_c/pull/280

PR #1033, opened 2021-08-31
Speed up running only tagged tests

m-blaha:
Currently each feature file is executed regardless it contains / not
contains any test with given tag.
With this change all feature files without tagged test are ignored.

time ./container-test -c dnfdaemon run -t dnfdaemon
Before:
real    3m43.881s
user    0m47.335s
sys     0m33.483s

After:
real    0m2.294s
user    0m0.276s
sys     0m0.213s

Also the output is now much cleaner and not polluted with unrelated
information about features with all scenarios skipped.

### DIFF
--- a/container-test
+++ b/container-test
@@ -206,8 +206,11 @@ class BehaveRunner(object):
 
     @property
     def tags(self):
-        tags = self.command_line_args.tags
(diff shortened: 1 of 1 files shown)


kontura:
Nice, thanks!

PR #1035, opened 2021-09-13
Dockerfile.ansible: cleanups and switch to Fedora 34

lukash:


Dockerfile.ansible
```diff
@@ -1,56 +1,32 @@
...
-# ---
-# $ podman run -it --rm ansible ./ansible
+# Example Usage:
+# $ podman build --build-arg TYPE=distro -t ansible -f Dockerfile.ansible
+# $ podman run --net none -it ansible ./ansible
 
+FROM fedora:34
+ARG TYPE=nightly
 
-# Build types
```
j-mracek, line 10: Why did you removed descriptions about build types? They look informative.

j-mracek: Only TYPE=nightly is not described there.

lukash: I've removed it because it was outdated. Maybe it deserves some more documentation, but it's just "nightly" or anything else atm. and maybe the best way to find out what it does is to look at the use... It's also common to both the Dockerfiles and it's somewhat briefly mentioned in the README.md.

> Only TYPE=nightly is not described there.

Not sure what this means exactly.

j-mracek: Do you want to enhance documentation as part of the PR or postpone it for future work? I don't see it as a blocker.

j-mracek:
LGTM

PR #1038, opened 2021-09-29
Add dedicated overlay for dnf5-ci testing

kontura:
Integration tests for dnf5 require only librepo and libdnf.
There is no need to build dnf, plugins or libcomps.

We could also possibly reuse dnf5-unstable overlay (which currently doesn't build librepo) but this seems like the more correct approach to me.


kontura:
For: https://github.com/rpm-software-management/libdnf/pull/1351

PR #1040, opened 2021-09-30
Additional tests for exclude_from_weak autodetection

pkratoch:


dnf-behave-tests/dnf/exclude_from_weak.feature
```diff
@@ -74,3 +74,53 @@ Scenario: Obsoletes are not disabled by exclude_from_weak
...
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                                   |
+        | upgrade       | PackageA-2.0-1.x86_64                     |
+
+
+@bz1699672
+@bz2005305
+Scenario: Upgrades installs supplementing package even if excluded from weak before
+  Given I use repository "exclude-from-weak"
```
j-mracek, line 98: Work for me - I am not sure whether the test and behavior of dnf is correct or expected.

pkratoch: I wrote these tests, to capture the behavior. It would be of course better if the supplementing package was not installed when excluded before, but I understand it's not possible to do, because the supplementing package might have appeared later on. But it's good to know about this limitation, hence the test.

j-mracek: I have theory -> `supplementing-pkg-2.0-1.x86_64` supplements `PackageB = 2.0` therefore it is installed when PackageB is upgraded to 2.0. But if `supplementing-pkg-2.0-1.x86_64` would supplement `PackageB` then it will be not installed. If it is true it is not nice behaviour (there was similar problem with recommends).

pkratoch: I split the test into two to see the difference and it's actually not the worst behavior. At least in one case dnf correctly excludes the supplementing package.

j-mracek:
Looks good to me

PR #1045, opened 2021-10-21
[versionlock] fix multi pkg lock (RhBug:2001039)

inknos:
fix: https://github.com/rpm-software-management/dnf-plugins-core/pull/447
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2013324
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2013332


m-blaha:
Hm, it looks like the tests are passing even on the unpatched version of versionlock.
I think the problem is in the initial state - to test the bug you have to have one of the `abcde` or `wget` packages locked before running `dnf versionlock abcde wget`. However you start with an empty `versionlock.list` file.
Also consider moving the tests to `dnf/plugins-core/versionlock-list-manipulation.feature` which seems to be more appropriate.

inknos:
Found the mistake. The tests were originally in `versionlock-exclude.feature` where you have a different locklist by default. I'll fix it now

PR #1051, opened 2021-11-04
copr: stderr output changed when the copr enable failed

schlupov:
Relates: https://github.com/rpm-software-management/dnf-plugins-core/pull/436

### DIFF
--- a/dnf-behave-tests/dnf/plugins-core/copr.feature
+++ b/dnf-behave-tests/dnf/plugins-core/copr.feature
@@ -77,7 +77,8 @@ Scenario: Test enabling a non-existent repo
 
       Please do not file bug reports about these packages in Fedora
       Bugzilla. In case of problems, contact the owner of this repository.
-      Error: Such repository does not exist.
+      Error: It wasn't possible to enable this project.
+      Project testuser/nonexistent does not exist.
       """
 
 
@@ -107,7 +108,7 @@ Given I create and substitute file "//etc/dnf/plugins/copr.conf" with
 
       Please do not file bug reports about these packages in Fedora
       Bugzilla. In case of problems, contact the owner of this repository.
-      Curl error (7): Couldn't connect to server for http://localhost:2/coprs/testuser/testproject/repo/fedora-30/dnf.repo?arch=x86_64 [Failed to connect to localhost port 2: Connection refused]
+      Error: Failed to connect to http://localhost:2/coprs/testuser/testproject/repo/fedora-30/dnf.repo?arch=x86_64: Connection refused
       """
 
 
@@ -122,6 +123,11 @@ Given I create and substitute file "//etc/dnf/plugins/copr.conf" with
       protocol = http
       port = {context.dnf.ports[copr]}
       """
+  And HTTP server GET /coprs/testuser/testproject/repo/fedora-31/dnf.repo?arch=x86_64 response headers are
+      """
+      Copr-Error-Data=eyJhdmFpbGFibGUgY2hyb290cyI6IFsiZmVkb3JhLTMwLXg4Nl82NCJdfQ==
+      """
+  And the server starts responding with HTTP status code 404
  When I execute dnf with args "copr enable testhub/testuser/testproject"
  Then the exit code is 1
   And stderr is
@@ -136,7 +142,13 @@ Given I create and substitute file "//etc/dnf/plugins/copr.conf" with
 
       Please do not file bug reports about these packages in Fedora
       Bugzilla. In case of problems, contact the owner of this repository.
-      Error: This repository does not have any builds yet so you cannot enable it now.
+      Error: It wasn't possible to enable this project.
+      Repository 'fedora-31-x86_64' does not exist in project 'testuser/testproject'.
+      Available repositories: 'fedora-30-x86_64'
+
+      If you want to enable a non-default repository, use the following command:
+        'dnf copr enable testuser/testproject <repository>'
+      But note that the installed repo file will likely need a manual modification.
       """
 
 
--- a/dnf-behave-tests/dnf/steps/fixtures/httpd.py
+++ b/dnf-behave-tests/dnf/steps/fixtures/httpd.py
@@ -38,7 +38,7 @@ class NoLogHttpHandler(SimpleHTTPRequestHandler):
         pass
 
 
-class LoggingHttpHandler(SimpleHTTPRequestHandler):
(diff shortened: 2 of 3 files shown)


dnf-behave-tests/dnf/plugins-core/copr.feature
```diff
@@ -136,7 +136,13 @@ Given I create and substitute file "//etc/dnf/plugins/copr.conf" with
 
       Please do not file bug reports about these packages in Fedora
       Bugzilla. In case of problems, contact the owner of this repository.
-      Error: This repository does not have any builds yet so you cannot enable it now.
```
schlupov, line 139: This has changed in Copr. The repository can be enabled even if there is no build yet, you just have to choose the right repository.

dnf-behave-tests/dnf/steps/fixtures/httpd.py
```diff
@@ -51,7 +51,14 @@ def do_GET(self):
             self.send_response(self.server._conf['status'])
             self.end_headers()
             return
-        super(LoggingHttpHandler, self).do_GET()
+        if self.server._conf.get("response_url"):
+            if self.server._conf["response_url"] == "/{context.dnf.tempdir}/coprs/testuser/testproject/repo/fedora-31" \
+                                                    "/dnf.repo":
```
lukash, line 56: The path is definitely wrong here. You've just created an arbitrary string which you used as an ID on both sides of the HTTP request, but this is, besiders arbitrary, an unnecessary complication. Use the real path of the request instead. There shouldn't be any scenario-specific constant in this place. Please put the headers which are supposed to be returned in the response into a dict and just retrieve them from the dict based on the real path of the GET request.

dnf-behave-tests/dnf/steps/fixtures/httpd.py
```diff
@@ -51,7 +51,14 @@ def do_GET(self):
...
+                self.send_response(404)
```
lukash, line 57: For that matter, this also isn't good. The particular response status code should also in general be stored in the same per-path dict. The global status that is being returned above (on line 51) is a simplification because we somehow never needed that to be different for two different URLs in a single scenario (the server is mostly used to serve files from a directory).

I think you should be fine with that single global status, so just put your code before that and then set the global status to be 404 in your step in the scenario, so that there's no specific 404 response in case you have extra headers configured, because those are unrelated.

dnf-behave-tests/dnf/steps/fixtures/httpd.py
```diff
@@ -51,7 +51,14 @@ def do_GET(self):
...
+                self.send_header("Copr-Error-Data", self.server._conf["headers"]["Copr-Error-Data"])
```
lukash, line 58: "Copr-Error-Data" is also a scenario-specific constant which shouldn't be here, if you do it the way I described, you should get rid of this here.

dnf-behave-tests/dnf/plugins-core/copr.feature
```diff
@@ -122,6 +123,10 @@ Given I create and substitute file "//etc/dnf/plugins/copr.conf" with
       protocol = http
       port = {context.dnf.ports[copr]}
       """
+  And HTTP server GET /{context.dnf.tempdir}/coprs/testuser/testproject/repo/fedora-31/dnf.repo response headers are
+      """
+      ('Copr-Error-Data', 'eyJhdmFpbGFibGUgY2hyb290cyI6IFsiZmVkb3JhLTMwLXg4Nl82NCJdfQ==')
```
lukash, line 128: This is kind-of a moot question in this context, but why are you base64-encoding text data? That makes little sense. I guess it's because you'd need to escape some characters in it? Poor excuse. This sucks, because this base64 string is not human readable, anyone reading it will not know what's inside without decoding. Any change to this string will again need to be encoded and hence it will not be clear in the PR what the change is, the reviewer again needs to go and decode.

lukash (commented):
Please separate the HTTP server changes into a separate commit (they should also be generic and not contain any bits specific to your particular scenario).

schlupov:
@lukash I updated the PR according to our call.

dnf-behave-tests/dnf/steps/fixtures/httpd.py
```diff
@@ -49,9 +49,12 @@ def do_GET(self):
         # process the request as usual.
         if 'status' in self.server._conf:
             self.send_response(self.server._conf['status'])
+            if "response_headers" in self.server._conf:
```
lukash, line 52: I'd recommend to add an empty `"response_headers"` dict to `_conf` when initializing it, i.e. around line 102 in `HttpServerContext.__init__()` (if I'm not mistaken).

This check should instead be:
```python
if self.path in self.server._conf["response_headers"]:
```

Because you can have some response headers configured but they don't need to be for this path and then you'd get an exception.

Also, please don't add your block into the `if 'status' in self.server._conf:` block. The response headers should work even if there's no special status set. (if it doesn't work because setting the headers doesn't cooperate with the `super(HttpHandler, self).do_GET()` being called at the end, let me know and we can have a look at that together)

dnf-behave-tests/dnf/steps/server.py
```diff
@@ -107,3 +107,10 @@ def step_http_log_contains(context):
     for e in expected:
         if e not in found:
             raise AssertionError('HTTP log does not contain "%s": %s' % (e, "\n" + "\n".join(found) + "\n"))
+
+
+@behave.step("HTTP server GET {url} response headers are")
+def step_http_server_get_response_headers(context, url):
+    headers = [tuple(context.text.split('=', 1))]
```
lukash, line 114: In general the `context.text` can be multiline with one header per line, can you please iterate the lines and add them to the dict one by one?

dnf-behave-tests/dnf/steps/server.py
```diff
@@ -107,3 +107,10 @@ def step_http_log_contains(context):
...
+    context.scenario.httpd.conf["status"] = 404
```
lukash, line 115: This line should not be here, it is again specific for your scenario, where you need to return 404 for your particular case. Remove this and instead add this step directly to your scenario:
```
And the server starts responding with HTTP status code 404
```
That should do it for you :slightly_smiling_face:

lukash (commented):
Thanks. Getting there, just a few minor tweaks!

lukash:
Looking good now, thanks!

PR #1061, opened 2022-01-20
Test for dnf group with disablerepo (RhBug:2064341)

inknos:
pr: https://github.com/rpm-software-management/dnf/pull/1819

bz: https://bugzilla.redhat.com/show_bug.cgi?id=2064341

### DIFF
--- a/dnf-behave-tests/dnf/comps-group-disabled-repo.feature
+++ b/dnf-behave-tests/dnf/comps-group-disabled-repo.feature
@@ -1,0 +1,38 @@
+Feature: Test group remove when repositories are disabled
+
+Background: Install group DNF-CI-Testgroup
+    Given I use repository "dnf-ci-thirdparty"
+      And I use repository "dnf-ci-thirdparty-2"
+      And I use repository "dnf-ci-fedora"
+     When I execute dnf with args "group install DNF-CI-Testgroup"
+     Then the exit code is 0
+      And Transaction is following
+          | Action        | Package                           |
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/comps-group-disabled-repo.feature
```diff
@@ -0,0 +1,43 @@
...
+
+Scenario: Remove group with disabled repository
+   When I execute dnf with args "group remove DNF-CI-Testgroup --disablerepo=\*dnf-ci-thirdparty"
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                           |
+        | remove        | filesystem-0:3.9-2.fc29.x86_64    |
+        | remove        | lame-0:3.100-4.fc29.x86_64        |
+        | remove-unused | setup-0:2.12.1-1.fc29.noarch      |
+        | remove-unused | lame-libs-0:3.100-4.fc29.x86_64   |
```
m-blaha, line 31: This is weird. May be I'm missing something, but:
1. `filesystem` and `lame` packages are part of both `DNF-CI-Testgroup` and `DNF-CI-Testgroup-2`
2. both groups are installed
3. when you try to remove one of the groups, the `filesystem` and `lame` packages should stay put on the system - because they are part of the other group

According to man dnf (group remove: Mark the group removed and remove those packages in the group  from  the system  which  do not belong to another installed group and were not installed explicitly by the user.) only the group should be removed, not the packages.

I didn't find any scenario testing this case (dnf group remove does not remove packages that belong to another installed group) in our suite.

PR #1062, opened 2022-01-25
Add a step for testing dnf4/5 stdout only in the appropriate mode

kontura:
It enables a couple of group tests for dnf5 using the new stdout matching.


lukash:
Looks neat, thanks!

PR #1065, opened 2022-02-03
Test solvfile cache is regenerated if solvfile version doesn't match

kontura:
Requires: https://github.com/rpm-software-management/libdnf/pull/1441

If we decide to use a different solvfile format the test can be easily modified.

### DIFF
--- a/dnf-behave-tests/dnf/steps/repo.py
+++ b/dnf-behave-tests/dnf/steps/repo.py
@@ -13,7 +13,13 @@ from common.lib.behave_ext import check_context_table
 from common.lib.checksum import sha256_checksum
 from common.lib.cmd import run_in_context
 from common.lib.diff import print_lines_diff
-from common.lib.file import copy_tree, create_file_with_contents, delete_file, ensure_directory_exists
+from common.lib.file import (
+    copy_tree,
+    create_file_with_contents,
+    delete_file,
+    ensure_directory_exists,
+    ensure_file_exists,
+)
 from fixtures import start_server_based_on_type, stop_server_type
 from fixtures.osrelease import osrelease_fixture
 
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/dnf/steps/repo.py
```diff
@@ -372,3 +378,22 @@ def given_system(context, system):
...
+    path = path.format(context=context)
+    ensure_file_exists(path)
+    fp = open(path, "r+b")
+    # Seek 40 bytes from start of file to the beginning userdata section
+    fp.seek(40, 0)
+    userdata_dnf_magic = fp.read(4).decode('utf-8')
+    assert userdata_dnf_magic == "dnfS"
+
+    userdata_dnf_version = fp.read(4).decode('utf-8')
+    assert userdata_dnf_version == "01.0"
```
kontura, line 395: This now relies on the libsolv solvfile format and dnf userdata section format, it will start failing when there is a change.

Unfortunately I cannot think of any better solution.

dnf-behave-tests/dnf/steps/repo.py
```diff
@@ -372,3 +378,22 @@ def given_system(context, system):
...
+    # Seek 40 bytes from start of file to the beginning userdata section
+    fp.seek(40, 0)
+    userdata_dnf_magic = fp.read(4)
+    assert userdata_dnf_magic == b'\x00dnf'
+
+    userdata_dnf_version = fp.read(4)
+    assert userdata_dnf_version == b'\x001.0'
+
+    fp.write(bytes(version, 'utf-8'))
+    fp.write(b'\x00\x00')
```
lukash, line 398: So here you rewrite the `libsolv_toolversion`, the original value of which is `"\0\0\0\0\01.2"` and IIUC after the rewrite the value is `"0.1\0\01.2"`? I guess this does the job but it'd be nicer to rewrite it in a bit cleaner way, how about making it look e.g. like `"\0\0\0bogus"` after the rewrite? The funky string will be much easier to differentiate in error messages, as opposed to ones and dots and zeros which can get mixed up easily.

kontura: Sure, I have changed it to `"\0\0\0bogus"`. :slightly_smiling_face:

dnf-behave-tests/dnf/steps/repo.py
```diff
@@ -372,3 +378,21 @@ def given_system(context, system):
...
+
+@behave.step("I invalidate solvfile version of \"{path}\"")
+def rewrite_solvfile_version(context, path):
+    path = path.format(context=context)
+    ensure_file_exists(path)
+    fp = open(path, "r+b")
+    # Seek 40 bytes from start of file to the beginning userdata section
+    fp.seek(40, 0)
+    userdata_dnf_magic = fp.read(4)
+    assert userdata_dnf_magic == b'\x00dnf'
```
lukash, line 391: These `assert`s are actually pretty confusing. Seeing them in test results I wondered what's going on and why isn't there a better message. Thing is one expects assert failures in test suites to be test result asserts, this one is asserting base test data before modifying them. Could you extend this with a message clearly saying this failed on mocking test data due to unexpected solv file content?

kontura: Right, I have added messages to both asserts, thanks! :+1:

PR #1066, opened 2022-02-03
Behavior of the rpmdb version is changed, each transaction causes change

jrohel:
* Each rpm transaction, including reinstalling the same package, will change the version of the rpm database.

* Format of the rpmdb version is changed, depends on librpm - The rpmdb version is now obtained from librpm. Format not defined. And can be changed in the future. E.g. now it's SHA1, but switch to SHA256 is ready. In fact, every non-empty string is valid.

It is related to the transition to the use of cookies from librpm.
https://github.com/rpm-software-management/dnf/pull/1811
https://github.com/rpm-software-management/libdnf/pull/1434


j-mracek:
LGTM

j-mracek:
LGTM!

PR #1067, opened 2022-02-03
Add help msg to container script

inknos:


container-test
```diff
@@ -139,6 +139,8 @@ class BehaveRunner(object):
 
     def __call__(self):
         command = getattr(self, 'command_%s' % self.command_line_args.command, None)
+        if command == None:
+            self.command_line_parser.print_help()
```
lukash, line 143: Not that it matters too much, but could you also add:
```python
error("No command specified.")
```

The `'None'` error message is understandable but not great.

inknos: sure.
the pr is just for laziness, I don't like to type `-h` to see the options 😄

PR #1068, opened 2022-02-04
Add step for generating repository that fails

pkratoch:
The gpg.feature requires repository with package with broken gpg
signature and createrepo_c started to return exit code 2 when
generating such repostiory.

### DIFF
--- a/dnf-behave-tests/common/lib/cmd.py
+++ b/dnf-behave-tests/common/lib/cmd.py
@@ -29,7 +29,7 @@ def run(cmd, shell=True, cwd=None):
     return proc.returncode, stdout, stderr
 
 
-def run_in_context(context, cmd, can_fail=False, **run_args):
+def run_in_context(context, cmd, can_fail=False, expected_exit_code=None, **run_args):
     if getattr(context, "faketime", None) is not None:
         cmd = context.faketime + cmd
 
@@ -48,6 +48,10 @@ def run_in_context(context, cmd, can_fail=False, **run_args):
 
     if not can_fail and context.cmd_exitcode != 0:
         raise AssertionError('Running command "%s" failed: %s' % (cmd, context.cmd_exitcode))
+    elif expected_exit_code is not None and int(expected_exit_code) != context.cmd_exitcode:
+        raise AssertionError(
+            'Running command "%s" had unexpected exit code: %s' % (cmd, context.cmd_exitcode)
+        )
 
 
 def assert_exitcode(context, exitcode):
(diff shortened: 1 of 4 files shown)


lukash:
What I'm a bit afraid of with this is that you don't verify in any way that createrepo_c fails with the error you expect it to fail. It can fail on anything else and behave will just eat the output and keep executing.

I'm not having great ideas, I'm not sure we could at least rely on the exit code 2 being somewhat unique to this kind of error and check that? That'd mean propagating the exit code into the step and further (`generate_repodata()` calls `run_in_context()` multiple times...).

But as long as the usage is quite unique maybe I'd be fine keeping the way you did it.

m-blaha:
Even when the createrepo_c returned exit code `2`, the repository is still created?

pkratoch:
> Even when the createrepo_c returned exit code `2`, the repository is still created?

Yes. Btw, there are tests for when createrepo_c exits with 2 that checks the repodata are still ok: https://github.com/rpm-software-management/ci-dnf-stack/commit/b88a6458bb222ab90a5e9ffb66ae209cfc81498b

> What I'm a bit afraid of with this is that you don't verify in any way that createrepo_c fails with the error you expect it to fail. It can fail on anything else and behave will just eat the output and keep executing.
> 
> I'm not having great ideas, I'm not sure we could at least rely on the exit code 2 being somewhat unique to this kind of error and check that? That'd mean propagating the exit code into the step and further (`generate_repodata()` calls `run_in_context()` multiple times...).
> 
> But as long as the usage is quite unique maybe I'd be fine keeping the way you did it.

My first idea was to put it into "then" step and check the exit code, but it fails on the assertion error and it would require bigger changes to do it this way. Propagating the expected exit code into `run_in_context()` and checking it there would work. There are multiple things that can cause exit code 2, so I didn't find it very useful, but now that I think about it, it could still be less frequent than exit code 1 and we'd at least differentiate those. I'll do that.

kontura:
LGTM

dnf-behave-tests/common/lib/cmd.py
```diff
@@ -48,6 +48,10 @@ def run_in_context(context, cmd, can_fail=False, **run_args):
 
     if not can_fail and context.cmd_exitcode != 0:
         raise AssertionError('Running command "%s" failed: %s' % (cmd, context.cmd_exitcode))
+    elif expected_exit_code is not None and int(expected_exit_code) != context.cmd_exitcode:
```
lukash, line 51: The conversion to `int` should happen on the input of the step, `run_in_context()` here should be accepting only numbers. Since I'm late to the party: https://github.com/rpm-software-management/ci-dnf-stack/pull/1069

PR #1073, opened 2022-02-28
Add test for mark command when no swdb on the system

j-mracek:


dnf-behave-tests/dnf/mark-install.feature
```diff
@@ -28,3 +28,16 @@ Scenario: Marking dependency as user-installed should not remove it automaticall
    And Transaction is following
         | Action        | Package                                   |
         | remove        | setup-0:2.12.1-1.fc29.noarch              |
+
+@bz2046581
+Scenario: Marking installed package by rpm to empty installroot as user-installed
```
lukash, line 33: The scenario name is confusing, did you mean:
> Marking a package installed by rpm in an empty installroot as user-installed

But "installed by rpm in an empty installroot" is just unnecessarily complicated and focusing on implementation details, this describes the important condition better:
> Marking a package not present in history DB as user-installed

j-mracek: Thanks, fixed

PR #1083, opened 2022-04-06
Enable dnf5 test

inknos:
This PR includes tests that *just work* with `@dnf5` tag.
It also includes functions to parse dnf4/5 stdout/err and exit codes. I also disables checking for transaction table stdout for dnf5

### DIFF
--- a/dnf-behave-tests/common/output.py
+++ b/dnf-behave-tests/common/output.py
@@ -111,6 +111,20 @@ def parse_dnf_version(text):
     assert False
 behave.register_type(dnf_version=parse_dnf_version)
 
+
+@behave.then("{dnf_version:dnf_version} exit code is")
+def then_dnf_exit_code_is(context, dnf_version):
+    """
+    Check for the test's exit code only if running in the
+    appropriate mode otherwise the step is skipped
+    """
+    dnf5_mode = hasattr(context, "dnf") and context.dnf.dnf5_mode
+    if dnf_version == "dnf5" and dnf5_mode:
+        then_the_exit_code_is(context)
+    if dnf_version == "dnf4" and not dnf5_mode:
+        then_the_exit_code_is(context)
+
+
 @behave.then("{dnf_version:dnf_version} stdout is")
 def then_dnf_stdout_is(context, dnf_version):
     """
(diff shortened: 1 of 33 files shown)


dnf-behave-tests/common/output.py
```diff
@@ -124,6 +172,30 @@ def then_dnf_stdout_is(context, dnf_version):
         then_stdout_is(context)
 
 
+@behave.then("dnf4 stdout matches line by line")
```
kontura, line 175: I am not sure if you prefer this approach but just adding a note that all the new steps could use the `{dnf_version:dnf_version}` pattern just like `@behave.then("{dnf_version:dnf_version} stdout is")`.

inknos: didn't know, I definitely like it more. thanks

dnf-behave-tests/common/output.py
```diff
@@ -54,6 +54,17 @@ def handle_reposync(expected, found, dnf5_mode):
     return expected, found
 
 
+@behave.then("{dnf_version:dnf_version} exit code is")
```
kontura, line 57: It seems the `{dnf_version:dnf_version}` pattern is being used before it is registered below (`behave.register_type(dnf_version=parse_dnf_version)`) so its not working.

dnf-behave-tests/common/output.py
```diff
@@ -124,6 +135,41 @@ def then_dnf_stdout_is(context, dnf_version):
...
+    dnf5_mode = hasattr(context, "dnf") and context.dnf.dnf5_mode
+    if dnf_version == "dnf5" and dnf5_mode:
+        then_stderr_is(context)
+    if dnf_version == "dnf4" and not dnf5_mode:
+        then_stderr_is(context)
+
+
+@behave.then("{dnf_version:dnf_version} stdout matches line by line")
+def then_dnf_stdout_matches_line_by_line(context, dnf_version):
+    """
```
kontura, line 153: Empty doc descriptions are added on a couple places.

kontura (changes requested):
Otherwise looks good.
All the marked tests are passing for me.

kontura (approved):

kontura:
Great, thank you!

PR #1084, opened 2022-04-11
repomanage: Add test when an rpm is both in new and old versions

kontura:
https://bugzilla.redhat.com/show_bug.cgi?id=2058676
https://bugzilla.redhat.com/show_bug.cgi?id=2034736

For: https://github.com/rpm-software-management/dnf-plugins-core/pull/452


kontura:
PR moved: https://github.com/rpm-software-management/dnf-plugins-core/pull/453

PR #1089, opened 2022-04-25
Add test for correct mapping of arch loongarch64 to basearch

j-mracek:
For commit: https://github.com/rpm-software-management/dnf/commit/96b08be69050cddbc4c3abdd227bd28969249db9


mbanas:
Looks good

PR #1093, opened 2022-05-04
Test for transaction-action plugin (RhBug:2023652)

j-mracek:
The test checks the absence of the traceback when the reason change is
handled by the transaction_action plugin. It uses reproducer provided by
the bug reporter. It is quite difficult to emulate the direct reason
change in DNF workflows therefore it uses installation of installonly
packages when one is removed. Alternatively the test can use group
install.


inknos:
Quck note. I personally don't like very long scenarios but I understand the difficulty to reproduce the behavior.
I would not know a better way to test it so LGTM 😃

PR #1100, opened 2022-05-12
Enable reposync scenario for RHEL 9

janblazek:
Scenario: Reposync does not stop downloading packages on the first error
should be already enabled in master and rhel-9.1.0


kontura:
The failing dnf4 CI is unrelated, fix: https://github.com/rpm-software-management/ci-dnf-stack/pull/1103

PR #1112, opened 2022-05-18
Fix stderr in builddep scenarios

inknos:
Depends on https://github.com/rpm-software-management/libdnf/pull/1521
Now the `stderr` is more consistent between dnf4 and dnf5

### DIFF
--- a/dnf-behave-tests/common/output.py
+++ b/dnf-behave-tests/common/output.py
@@ -106,77 +106,14 @@ def then_stdout_is(context):
 
 @parse.with_pattern(r"dnf4|dnf5")
 def parse_dnf_version(text):
-    if text in ("dnf4", "dnf5"):
-        return text
-    assert False
+    return text
 behave.register_type(dnf_version=parse_dnf_version)
 
 
-@behave.then("{dnf_version:dnf_version} exit code is")
-def then_dnf_exit_code_is(context, dnf_version):
-    """
-    Check for the test's exit code only if running in the
-    appropriate mode otherwise the step is skipped
-    """
-    dnf5_mode = hasattr(context, "dnf5_mode") and context.dnf5_mode
-    if dnf_version == "dnf5" and dnf5_mode:
-        then_the_exit_code_is(context)
-    if dnf_version == "dnf4" and not dnf5_mode:
-        then_the_exit_code_is(context)
-
-
-@behave.then("{dnf_version:dnf_version} stdout is")
-def then_dnf_stdout_is(context, dnf_version):
-    """
-    Check for exact match of the test's stdout only if running in the
-    appropriate mode otherwise the step is skipped.
-    """
-    dnf5_mode = hasattr(context, "dnf5_mode") and context.dnf5_mode
-    if dnf_version == "dnf5" and dnf5_mode:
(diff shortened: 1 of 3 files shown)


dnf-behave-tests/dnf/steps/cmd.py
```diff
@@ -301,6 +301,15 @@ def then_stdout_contains_lines(context):
         else:
             raise AssertionError("Stderr doesn't contain line: %s" % line)
 
+@behave.then("{dnf_version:dnf_version} stderr contains \"{text}\"")
+def then_dnf_stderr_contains(context, dnf_version, text):
+    dnf5_mode = hasattr(context, "dnf") and context.dnf5_mode
```
m-blaha, line 306: This should be `hasattr(context, "dnf5_mode")` since `dnf5_mode` is now stored directly in the context, not in context.dnf.

m-blaha: Unrelated to this PR - is there a reason why we have part of `then("stderr/stdout ...")` steps defined here in `dnf/steps/cmd.py` and part in `common/output.py`? May be we should gather them all in one place.
Maybe @lukash as he authored https://github.com/rpm-software-management/ci-dnf-stack/commit/dec924fae7215e5be61533346e45848630968ac7 ?

inknos: thanks. I see now why it's failing

Agree on the second comment, the steps are all around files. They could be grouped easily like `"dnf{4,5} std{out,err} contains"`

lukash: @m-blaha I think the reason is they aren't used in the createrepo_c tests and I dislike the "contains" variants, I still believe we should test the entire output, handling any variations with either regexes or more specific code.

So I tried to keep them local to the dnf tests, where ideally we'd be trimming down on their usage as well.

But regardless, there's also an argument to be made that in general steps that aren't used in createrepo_c tests should be kept local to dnf, just to keep the scope as small as possible...

inknos: I prefer the full output testing too. The reason is that too many functions can be too hard to maintain with the changes impacting dnf4/5.

What happened to me right now. The CI was mysteriously failing on some step that never caused issues. This and many steps are defined, like `"then stderr does not contain"` and `"then stderr contains"`, but they share the same function signature. I don't know how many other duplicates can be, I will take a look.

dnf-behave-tests/common/output.py
```diff
@@ -112,71 +112,12 @@ def parse_dnf_version(text):
...
-    dnf5_mode = hasattr(context, "dnf5_mode") and context.dnf5_mode
-    if dnf_version == "dnf5" and dnf5_mode:
-        then_stderr_matches_line_by_line(context)
-    if dnf_version == "dnf4" and not dnf5_mode:
-        then_stderr_matches_line_by_line(context)
+@parse.with_pattern(r"stdout|stderr")
+def parse_std_stream(text):
+    if text in ("stdout", "stderr"):
+        return text
+    assert False
```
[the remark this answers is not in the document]
inknos: should I change the other parser (line 107) as well?

PR #1113, opened 2022-05-25
Fix upgrade-cached.feature to run correctly in UBI container

m-blaha:
The test was allways passing in UBI container environment because dnf
was not running in german locales and the bug does not manifest itself
in english environment.
There are two causes of this:
1. glibc-langpack-de was missing in requirements.spec
2. rpm in UBI container is configured to install only en_US languages
   (see /etc/rpm/macros.image-language-conf)
In order to run dnf in german locale we need to provide minimal german
dnf.mo with required strings translated and add glibc-langpack-de to
requirements.spec.


dnf-behave-tests/dnf/upgrade-cached.feature
```diff
@@ -15,9 +17,14 @@ Scenario: Upgrade works correctly with non-english locale when packages were alr
   Given I use repository "dnf-ci-fedora-updates" as http
     # pre-download updates to the cache
     And I successfully execute dnf with args "upgrade --downloadonly"
+    # In UBI containers rpm does not install translation files (see /etc/rpm/macros.image-language-conf).
+    # We need to copy german dnf.mo to proper location in order to be able to run dnf in german locales.
+    And I copy file "{context.dnf.fixturesdir}/data/already-downloaded/dnf.mo" to "//usr/share/locale/de/LC_MESSAGES/dnf.mo"
```
kontura, line 22: I was confused by the `already-downloaded` directory name until I looked up what was the problem in the original bug.

Maybe some other name for directory would be clearer? Especially since the `dnf.mo` is not that easily readable and could also contain kind of anything. What about `dnf-translations`?

PR #1117, opened 2022-05-26
Active tag to enable skipping scenarios on dnf4

m-blaha:
New active tag `dnf` enables filtering scenarios according dnf5_mode
value.
@not.with_dnf=4  - do not run scenario with dnf4
@use.with_dnf=4  - run scenario only with dnf4
@not.with_dnf=5  - do not run scenario with dnf5
@use.with_dnf=5  - run scenario only with dnf5


dnf-behave-tests/dnf/environment.py
```diff
@@ -265,12 +269,15 @@ def before_scenario(context, scenario):
         if skip:
             scenario.skip(reason="Disabled by default by @xfail.")
 
-    if context.tag_matcher.should_exclude_with(scenario.effective_tags):
+    if context.os_tag_matcher.should_exclude_with(scenario.effective_tags):
         scenario.skip(reason="Disabled by OS version tag.")
 
+    if context.dnf_tag_matcher.should_exclude_with(scenario.effective_tags):
+        scenario.skip(reason="Disabled by active tag.")
```
lukash, line 276: I'd make the message more explicit by saying e.g. _"Disabled by dnf = 4/5 tag."_

lukash (commented):
Is this meant to work together with the `@dnf5` tag we already have?

On its own this would mean scenarios run on both dnf4 and dnf5. Since we want the default to be to only run on dnf4 and `@dnf5` now means to run on dnf5 as well as dnf4, don't we only need one more additional tag to say to run only on dnf5?

PR #1120, opened 2022-05-31
Fix scenario with relative path

m-blaha:
Scenario uses relative path but this cannot work without setting proper
working directory.


jrohel:
OK

PR #1118, opened 2022-05-26
buiddep: Test that BuildConflicts is respected

m-blaha:
Requires https://github.com/rpm-software-management/ci-dnf-stack/pull/1117

PR #1122, opened 2022-06-02
Store `dnf5_mode` and `dnfdaemon_mode` directly in behave context

kontura:
We need to know status of `dnf5_mode` in `before_all` for active tag
matching, this is before `DNFContext` is created in `before_scenario`.

We usually don't specify `dnf5_mode` manually, we use just `dnf5` tag.

Logic for inferring `dnf5_mode` from `dnf5` tag was present only in
`before_scenario` but was missing in `before_all`. Since we don't want
to duplicate it move it to `before_all` and store `dnf5_mode` directly
in behave `context` (`dnfdaemon_mode` as well).

I noticed PR: https://github.com/rpm-software-management/ci-dnf-stack/pull/1118 passed CI even though it shouldn't. It is because we don't specify `dnf5_mode` manually when running the CI and the new tests weren't run.
Another solution would be to run CI with define `dnf5_mode=1` but I think that would need to be added to `container-test` script.

### DIFF
--- a/dnf-behave-tests/common/output.py
+++ b/dnf-behave-tests/common/output.py
@@ -80,7 +80,7 @@ def then_stdout_is(context):
     if found == [""]:
         found = []
 
-    dnf5_mode = hasattr(context, "dnf") and context.dnf.dnf5_mode
+    dnf5_mode = hasattr(context, "dnf") and context.dnf5_mode
     clean_expected, clean_found = handle_reposync(expected, found, dnf5_mode)
 
     if clean_expected == clean_found:
@@ -118,7 +118,7 @@ def then_dnf_exit_code_is(context, dnf_version):
     Check for the test's exit code only if running in the
     appropriate mode otherwise the step is skipped
     """
-    dnf5_mode = hasattr(context, "dnf") and context.dnf.dnf5_mode
+    dnf5_mode = hasattr(context, "dnf") and context.dnf5_mode
     if dnf_version == "dnf5" and dnf5_mode:
         then_the_exit_code_is(context)
     if dnf_version == "dnf4" and not dnf5_mode:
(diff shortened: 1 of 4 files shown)


m-blaha (approved):
Makes sense. Thanks for fixing this.

dnf-behave-tests/common/output.py
```diff
@@ -80,7 +80,7 @@ def then_stdout_is(context):
     if found == [""]:
         found = []
 
-    dnf5_mode = hasattr(context, "dnf") and context.dnf.dnf5_mode
+    dnf5_mode = hasattr(context, "dnf") and context.dnf5_mode
```
lukash, line 83: But then the `hasattr(context, "dnf")` check is redundant (I wonder why it was needed...) and you can use `context.dnf5_mode` directly instead of this variable?

lukash: Oh. These are common steps, It must have been because of the createrepo_c tests, which don't have neither `context.dnf` not `context.dnf5_mode`, so I suspect you've broken createrepo_c tests with this?

m-blaha: oh, right. We probably need something like `hasattr(contect, "dnf5_mode") and context.dnf5_mode`.
Thanks for pointing this out.

lukash: Yeah, well I've just realized it shouldn't be broken, the `hasattr(context, "dnf")` should actually do the job, though it's not awfully nice or intuitive :slightly_smiling_face:

kontura: yes, the createrepo_c CI passed, but it is not nice.
I made a follow up PR: https://github.com/rpm-software-management/ci-dnf-stack/pull/1123 
Thanks!

PR #1124, opened 2022-06-08
dnf5: Enable test for solvfile versioning

kontura:
For: https://github.com/rpm-software-management/libdnf/pull/1535


kontura:
I have removed the check for the log we removed in the libdnf5 PR.

PR #1126, opened 2022-06-10
Test that failovermethod option does not print error

m-blaha:
Test for bug https://bugzilla.redhat.com/show_bug.cgi?id=2039906


jrohel:
What happens when `failovermethod=unsupported_value`? How about adding a test for this case?

m-blaha:
I was thinking about it. But given that the option was removed completely in upstream I decided to not add a test that works only on RHEL8 (this patch is downstream, specific only for RHEL8).
And in case of unsupported value, the `invalid configuration value` message is printed on terminal, which is fine I guess.

PR #1129, opened 2022-06-22
createrepo_c: test duplicates detection and warnings

kontura:
For: https://github.com/rpm-software-management/createrepo_c/pull/322


dnf-behave-tests/createrepo_c/duplicates.feature
```diff
@@ -0,0 +1,88 @@
...
+      Pool started (with 5 workers)
+      Pool finished with errors
+      """
+  And stderr contains only lines
+  """
+  Warning: Package 'package-0:0.2.1-1.fc29.x86_64' has duplicate metadata entries, only one should exist
+  Warning:     Sourced from location: 'subdir2/package-0.2.1-1.fc29.x86_64.rpm'
+  Warning:     Sourced from location: 'subdir/package-0.2.1-1.fc29.x86_64.rpm'
+  Warning:     Sourced from location: 'package-0.2.1-1.fc29.x86_64.rpm'
+  Warning:     Sourced from location: 'subdir3/package-0.2.1-1.fc29.x86_64.rpm'
```
dralley, line 33: The exact order of these items might be non-deterministic, I think it would depend on which files various threads in the threadpool manage to insert first.  We could sort the entries before printing them if that's a problem.

kontura: Yes it was non-deterministic, I was adding a step for that `stderr contains only lines` which doesn't check the order of lines but with your update we don't need that. That is better. :slightly_smiling_face: 
Thank you.

PR #1134, opened 2022-07-21
Additional tests for security filters with obsoletes

kontura:
Requires: https://github.com/rpm-software-management/dnf/pull/1835

### DIFF
--- a/dnf-behave-tests/dnf/check-update.feature
+++ b/dnf-behave-tests/dnf/check-update.feature
@@ -30,3 +30,18 @@ Given I use repository "dnf-ci-fedora-updates" with configuration
  When I execute dnf with args "upgrade"
  Then the exit code is 0
   And Transaction is empty
+
+
+@bz2101421
+Scenario: --security check-update doesn't show pkgs from resolved advisories (when obsoletes are involved)
+Given I use repository "check-update"
+  And I execute dnf with args "install A-1-1"
+ When I execute dnf with args "update --security"
+ Then the exit code is 0
+  And Transaction is empty
+ When I execute dnf with args "check-update --security"
+ Then the exit code is 0
+  And stdout is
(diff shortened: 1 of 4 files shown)


kontura:
I have removed the unrelated repoquery security-filters tests. I will add them in a separate dnf5 PR.

dnf-behave-tests/dnf/check-update.feature
```diff
@@ -30,3 +30,15 @@ Given I use repository "dnf-ci-fedora-updates" with configuration
...
+@bz2101421
+Scenario: --security check-update doesn't show pkgs from resolved advisories (when obsoletes are involved)
+Given I use repository "check-update"
+  And I execute dnf with args "install A-1-1"
+ When I execute dnf with args "update --security"
+ Then the exit code is 0
+ And Transaction is empty
+ When I execute dnf with args "check-update --security"
+ Then the exit code is 0
+ And stdout is empty
```
m-blaha, line 44: The test "almost" passes even on unpatched dnf. I mean, the output in the last step contains only "Last metadata expiration check: 0:00:01 ago on Po 8. srpna 2022, 15:26:55.". Is that alright?
As far as I understand the bugreport, it should contain some obsoleting packages, right?

m-blaha: Hmm, the output from GHA check shows some obsoleting packages. Maybe something is wrong with my local environment.

kontura: Oh but then maybe I should change it to:
```Gherkin
And stdout is
"""
<REPOSYNC>
"""

Because otherwise it could fail even without any packages in the output.

kontura: Updated.

PR #1136, opened 2022-07-25
Test for a non-file in /etc/dnf/vars

bajertom:
Hi. My first PR here. I had no idea how to name the test or if it would be better to add the code to some already existing one.
I'll use better, more descriptive commit message next time - my bad!

### DIFF
--- a/dnf-behave-tests/dnf/vars.feature
+++ b/dnf-behave-tests/dnf/vars.feature
@@ -87,3 +87,16 @@ Scenario: Variables with {} are substituted in repo id
       repo id\s+repo name
       dnf-ci-test-fedora\s+dnf-ci-test-fedora test repository
       """
+
+@bz2091636
+Scenario: Using dnf with non-files in /etc/dnf/vars
+  Given I create directory "/{context.dnf.installroot}/etc/dnf/vars/troublemaker"
+    And I use repository "dnf-ci-fedora"
+   When I execute dnf with args "repolist"
+   Then the exit code is 0
+    And stdout matches line by line
+      """
+      repo id\s+repo name
+      dnf-ci-fedora\s+dnf-ci-fedora test repository
+      """
+


dnf-behave-tests/dnf/vars-parsing.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Correct parsing of dnf/vars directory
+ 
+ 
+@bz2091636
+Scenario: Using dnf with non-files in /etc/dnf/vars
+  Given I create directory "/{context.dnf.installroot}/etc/dnf/vars"
+    And I create directory "/{context.dnf.installroot}/tmp"
+  Given I successfully execute "ln -s {context.dnf.installroot}/tmp {context.dnf.installroot}/etc/dnf/vars/symlink"
```
lukash, line 8: You don't need a `Given` here, you can use `And` for better consistency.

dnf-behave-tests/dnf/vars-parsing.feature
```diff
@@ -0,0 +1,12 @@
...
+    And I use repository "dnf-ci-fedora"
+   When I execute dnf with args "repolist"
+   Then the exit code is 0
+    And stdout contains "repo id"
```
lukash, line 12: This is a very weak check, what is the full expected output and could you check it in its entirety?

lukash (commented):
You can still change the commit message, just do a `git commit --amend` and force-push your branch.

You can also move the code to the `vars.feature` file so that you don't create this small file with one test. You can move the code and agai namend the commit and force-push to update the PR.

dnf-behave-tests/dnf/vars-parsing.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Correct parsing of dnf/vars directory
+ 
+ 
+@bz2091636
+Scenario: Using dnf with non-files in /etc/dnf/vars
+  Given I create directory "/{context.dnf.installroot}/etc/dnf/vars"
```
j-mracek, line 6: If the path is modified to create a directory in `/etc/dnf/vars` like /{context.dnf.installroot}/etc/dnf/vars/troublemaker" then the step with creating a symlink is not required (directory is also not a file object).

dnf-behave-tests/dnf/vars-parsing.feature
j-mracek, line 12: It can also check `stderr` that it does not contains a traceback.

```
And stderr does not contain "Traceback"
```

lukash: tbh. that's not a good check, we could have that check in every scenario. Checking the exit code should be enough. (I've always been wondering if checking `stderr` is empty in we don't have an explicit `stderr` check would be feasible. That'd be much more robust, but it's another discussion.)

j-mracek: I believe that the main part for the bug is test of the return code. The test of outputs is nice to have - it provides expected output after loading wars and even small peace is enough. And test that `stderr` does not contain `traceback` is also nice.

j-mracek: But it works - it is not sensitive to plugins, and wording of error message.

j-mracek: @lukash I think that your proposal is good (to test every scenario stderr) but it requires another implementation that you proposed because there are several dnf commands that has not empty stderr. I also think that such a change shouldn't be included in this PR.

lukash: I hope it's clear from my first comment I wasn't suggesting it for this PR...

j-mracek: Sorry for the confusion

dnf-behave-tests/dnf/vars-parsing.feature
j-mracek: It looks like that you forget to push changes

dnf-behave-tests/dnf/vars-parsing.feature
j-mracek: Well I think that the test see a different output. There should be only `dnf-ci-fedora` repository because repolist is run together with installroot option. To test it just change expected RC to 1 and run it with patched RPM. The bug resulted in crash that prevent any output and it created a traceback, therefore the test doe not need to test exact match with output. In `repolist.feture` I've found following line `And stdout contains "dnf-ci-fedora\s+dnf-ci-fedora test repository\s+enabled"`. May be it can be used also for the test. What do you think?
PS: I guess that you did not push the latest changes.

PR #1138, opened 2022-08-09
[dnf5 overlay] move to new codebase

inknos:


kontura:
I think this is running into a problem with spec name. It expects that the repo dnf5 has a spec called `dnf5.spec`.
I made a PR that should help: https://github.com/rpm-software-management/dnf5/pull/3

overlays/dnf5-ci/overlay.yml
```diff
@@ -10,7 +10,7 @@ components:
 
   - name: libdnf
```
kontura, line 11: There is a couple more problems, the `name` also needs to be set to `dnf5` (for both overlays).
And `rpm-gitoverlay` trips up because there is no `git tag` in the new repo. I am looking into fixing `rpm gitoverlay`.

inknos: I had the code fixed locally. I pushed an update

kontura:
There are two failing tests, that are not related to this and are even expected to fail.
Thanks!

PR #1141, opened 2022-08-10
Test for passing plugin parameters with dashes in names

jan-kolarik:
Requires https://github.com/rpm-software-management/dnf/pull/1839.

### DIFF
--- a/dnf-behave-tests/dnf/plugins-core/enable-plugins-dashes.feature
+++ b/dnf-behave-tests/dnf/plugins-core/enable-plugins-dashes.feature
@@ -1,0 +1,36 @@
+Feature: Tests for enabling plugins containing dashes in names
+
+Background: Enable plugins
+  Given I do not disable plugins
+    And I use repository "dnf-ci-fedora"
+   When I execute dnf with args "repoquery empty --verbose"
+   Then the exit code is 0
+    And stdout contains "Loaded plugins: .*config-manager"
+
+@bz1980712
+Scenario: Existing plugin containing dashes in the name could be disabled
+   When I execute dnf with args "repoquery empty --disableplugin=config-manager --verbose"
+   Then the exit code is 0
+    And stdout does not contain "Loaded plugins: .*config-manager"
+    And stderr does not contain "No matches found for the following disable plugin patterns: config-manager"
+
+@bz1980712
+Scenario: Existing plugin containing dashes in the name could be enabled
+   When I execute dnf with args "repoquery empty --disableplugin='*' --enableplugin=config-manager --verbose"
+   Then the exit code is 0
+    And stdout contains "Loaded plugins: .*config-manager"
+    And stderr does not contain "No matches found for the following enable plugin patterns: config-manager"
+
+@bz1980712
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/plugins-core/enable-plugins-dashes.feature
```diff
@@ -0,0 +1,35 @@
...
+@bz1980712
+Scenario: Existing plugin containing dashes in the name could be disabled
+   When I execute dnf with args "repoquery empty --disableplugin=config-manager --verbose"
+   Then the exit code is 0
+    And stdout does not contain "config-manager"
+    And stderr does not contain "No matches found for the following disable plugin patterns: config-manager"
+
+@bz1980712
+Scenario: Existing plugin containing dashes in the name could be enabled
+   When I execute dnf with args "--disableplugin=config-manager"
```
m-blaha, line 16: `--disableplugin` only affects the current run of dnf, disablement of plugin is not persistent. That said, this line doesn't do anything useful and does not affect the following `repoquery --enableplugin...`
To simulate enabling a disabled plugin, you coud use `--disableplugin='*' --enableplugin='config-manager`

jan-kolarik: Thank you Marek, how could I've missed that :slightly_smiling_face:

dnf-behave-tests/dnf/plugins-core/enable-plugins-dashes.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Tests for enabling plugins containing dashes in names
+
+Background: Enable plugins
+  Given I do not disable plugins
+    And I use repository "dnf-ci-fedora"
```
m-blaha, line 5: Here you should make sure, that the config-manager plugin is really available and enabled. Either by checking `dnf -v` output or by checking the exit code of `dnf config-manager --help` command.

dnf-behave-tests/dnf/plugins-core/enable-plugins-dashes.feature
```diff
@@ -0,0 +1,33 @@
...
+
+@bz1980712
+Scenario: Existing plugin containing dashes in the name could be disabled
+   When I execute dnf with args "repoquery empty --disableplugin=config-manager --verbose"
+   Then the exit code is 0
+    And stdout does not contain "config-manager"
```
m-blaha, line 11: I would prefer more accurate check here. The exact match would require defined set of available plugins, but at least something like `And stdout does not contain "Loaded plugins: .*config-manager"` would be better. Or maybe `stdout matches line by line` step might be useful (at least for positive cases).

jan-kolarik: Definitely, this is better and I didn't check that `contain` step is using pattern matching.

dnf-behave-tests/dnf/plugins-core/enable-plugins-dashes.feature
jan-kolarik: I agree, thanks.

PR #1142, opened 2022-08-11
builddep: Test for using macro defines with source rpms

jan-kolarik:
Requires https://github.com/rpm-software-management/dnf-plugins-core/pull/461.


kontura:
Could you also check if the [dnf5 builddep plugin](https://github.com/rpm-software-management/dnf5/blob/main/dnf5-plugins/builddep_plugin/builddep.cpp) has the same problem? And if it has add the warning there as well?

We could then enable this test also for dnf5 with the `@dnf5` tag.

jan-kolarik:
> Could you also check if the [dnf5 builddep plugin](https://github.com/rpm-software-management/dnf5/blob/main/dnf5-plugins/builddep_plugin/builddep.cpp) has the same problem? And if it has add the warning there as well?
> 
> We could then enable this test also for dnf5 with the `@dnf5` tag.

Yes I checked it and it is happening also with dnf5. I will deliver the fix in separate PR and then I will add the dnf5 tag here and close this comment thread.

kontura:
Is this waiting on something or can we move it from the draft stage and merge it?

jan-kolarik:
@kontura Oh sorry, I just forgot to switch the state.

PR #1143, opened 2022-08-11
Test to fix upgrade from noarch file

inknos:
https://github.com/rpm-software-management/dnf/pull/1840

### DIFF
--- a/dnf-behave-tests/dnf/upgrade-from-noarch.feature
+++ b/dnf-behave-tests/dnf/upgrade-from-noarch.feature
@@ -1,0 +1,17 @@
+Feature: Upgrade package from noarch to arch
+
+Background: Install package to upgrade
+   Given I use repository "upgrade-from-noarch"
+    When I execute dnf with args "install it-1.0"
+    Then the exit code is 0
+     And DNF Transaction is following
+         | Action        | Package         |
+         | install       | it-1.0-1.x86_64 |
+
+Scenario: Upgrade to rpm from file which is not noarch
+    When I execute dnf with args "upgrade {context.dnf.fixturesdir}/repos/upgrade-from-noarch/noarch/it-2.0-1.noarch.rpm"
+    Then the exit code is 0
+     And DNF Transaction is following
+         | Action        | Package         |
+         | upgrade       | it-2.0-1.noarch |
+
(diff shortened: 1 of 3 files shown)


dnf-behave-tests/dnf/upgrade-from-noarch.feature
```diff
@@ -0,0 +1,18 @@
...
+Background: Install package to upgrade
+   Given I use repository "upgrade-from-noarch"
+    When I execute dnf with args "install it-1.0"
+    Then the exit code is 0
+     And DNF Transaction is following
+         | Action        | Package         |
+         | install       | it-1.0-1.x86_64 |
+
+Scenario: Upgrade to rpm from file which is not noarch
+   Given I execute dnf with args "install it.noarch --downloadonly --downloaddir={context.dnf.installroot}/it"
```
m-blaha, line 12: "Given I **successfully** execute..."?
Or better, you do not need to download the package, you can directly use fixtures path in the following `I execute dnf with args` step: `{context.dnf.fixturesdir}/repos/upgrade-from-noarch/noarch/it-2.0-1.noarch.rpm`

dnf-behave-tests/dnf/upgrade-from-noarch.feature
```diff
@@ -0,0 +1,18 @@
...
+    When I execute dnf with args "install {context.dnf.installroot}/it/it-2.0-1.noarch.rpm"
```
m-blaha, line 13: The test passes even on unpatched dnf. I guess you should use the `upgrade` command here, `install` command is not affected by the bug.

m-blaha (changes requested):

dnf-behave-tests/dnf/upgrade-from-noarch.feature
inknos: thanks, I fixed it

dnf-behave-tests/dnf/upgrade-from-noarch.feature
inknos: I am testing again, good catch

inknos: @m-blaha should be good to go now

PR #1144, opened 2022-08-17
Move the testing container to the latest Fedora

m-blaha:
Requires https://github.com/rpm-software-management/dnf5/pull/11

### DIFF
--- a/.github/workflows/ci-host.yml
+++ b/.github/workflows/ci-host.yml
@@ -29,11 +29,7 @@ jobs:
           echo ${{secrets.GITHUB_TOKEN}} | docker login ghcr.io -u ${GHCR_USER} --password-stdin
 
           docker build -t ${IMAGE}:${DATE} -t ${IMAGE}:latest -<<EOF
-          # TODO(lukash) temporarily lock fedora to 35 due to bugged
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-host.yml
```diff
@@ -29,11 +29,7 @@ jobs:
           echo ${{secrets.GITHUB_TOKEN}} | docker login ghcr.io -u ${GHCR_USER} --password-stdin
 
           docker build -t ${IMAGE}:${DATE} -t ${IMAGE}:latest -<<EOF
-          # TODO(lukash) temporarily lock fedora to 35 due to bugged
-          # clang-format (clang-tools-extra-0:14.0.0-1.fc36):
-          # https://github.com/llvm/llvm-project/issues/55260
-          # change back to fedora:latest when fixed
-          FROM fedora:35
+          FROM fedora:latest
```
lukash, line 32: I was wondering whether it's better to have `fedora:latest` or a concrete version, because there tend to be issues when moving to the next version and with latest it's kind of arbitrary, unexpected and usually inconvenient if it breaks.

m-blaha: Good point. Thinking about it, having an explicit Fedora version would be clearer. I'll change it.

PR #1148, opened 2022-08-22
Test for showing detailed error message when TMPDIR does not exist

jan-kolarik:
Requires https://github.com/rpm-software-management/librepo/pull/262.

### DIFF
--- a/dnf-behave-tests/dnf/removed-tmpdir.feature
+++ b/dnf-behave-tests/dnf/removed-tmpdir.feature
@@ -1,0 +1,13 @@
+Feature: Better error message when TMPDIR is missing
+
+@bz2019993
+Scenario: show detailed error message when $TMPDIR does not exist
+Given I use repository "dnf-ci-fedora" with configuration
+      | key             | value                                       |
+      | baseurl         |                                             |
+      | mirrorlist      | {context.dnf.installroot}/tmp/mirrorlist    |
+      | gpgcheck        | 0
+  And I set environment variable "TMPDIR" to "/tmp/dummy.ABC123"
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/removed-tmpdir.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Better error message when TMPDIR is missing
+
+@bz2019993
+Scenario: show detailed error message when $TMPDIR does not exist
+Given I use repository "dnf-ci-fedora" with configuration
+      | key             | value                                       |
+      | baseurl         |                                             |
+      | mirrorlist      | {context.dnf.installroot}/tmp/mirrorlist    |
+      | gpgcheck        | 0
+  And I set dnf command to "TMPDIR=/tmp/dummy.ABC123 dnf"
```
jrohel, line 10: Redefines the dnf command to set an environment variable.
I think we have support for the definition of environmental variables in CI.
`And I set environment variable "TMPDIR" to "/tmp/dummy.ABC123"`

jan-kolarik: Yes I've already tried that, but this command is probably not exporting the variable to child processes, so I didn't work.

jan-kolarik: So I double-checked your proposal and it is really working. I don't understand what I was doing wrong before :slightly_smiling_face:

jan-kolarik: Thanks.

PR #1150, opened 2022-08-31
Tests for system-upgrade plugin migration to core

jan-kolarik:
Requires:
- https://github.com/rpm-software-management/dnf-plugins-extras/pull/210
- https://github.com/rpm-software-management/dnf-plugins-core/pull/462


j-mracek:
Please could you also add a decorator for each scenario containing bug where the movement was requested. `@bz2054235`

j-mracek:
Note for the release - we need to request distribution to add the package to the compose.

inknos:
all good here, thank you

PR #1153, opened 2022-09-07
A test for not double-encoding RPM URLs passed on CLI

lukash:


jan-kolarik:
No problem here, merging. Thanks :slightly_smiling_face:

PR #1154, opened 2022-09-07
Advisory: Add test for upgrades with pkg of various arches

kontura:
For: https://github.com/rpm-software-management/libdnf/pull/1576

https://bugzilla.redhat.com/show_bug.cgi?id=2124483

@bajertom, @glum23 please take a look. 
I have added a bunch of tests for various situations when different arches and obsoletes are present.
Mostly for the `upgrade` command but there is a couple for `list` as well.

https://issues.redhat.com/browse/RHELX-71
https://issues.redhat.com/browse/RHELX-72

### DIFF
--- a/dnf-behave-tests/dnf/security-upgrade.feature
+++ b/dnf-behave-tests/dnf/security-upgrade.feature
@@ -269,3 +269,20 @@ Given I use repository "security-upgrade"
   When I execute dnf with args "upgrade --security --setopt=obsoletes=false"
  Then the exit code is 0
   And Transaction is empty
+
+
+# Tests that we correctly filter packages and include both arches in transaction
+# (both have an advisory) if we didn't libsolv would fail to resolve the transaction.
+Scenario: --security upgrade with packages of multiple arches installed
+Given I use repository "security-upgrade-multilib"
+  And I successfully execute dnf with args "install B-1-1.x86_64 B-1-1.i686"
+  Then Transaction is following
+      | Action        | Package        |
+      | install       | B-0:1-1.x86_64 |
+      | install       | B-0:1-1.i686   |
+ When I execute dnf with args "upgrade --security"
+ Then the exit code is 0
+  And Transaction is following
+      | Action        | Package        |
+      | upgrade       | B-0:2-2.x86_64 |
+      | upgrade       | B-0:2-2.i686   |
--- a/dnf-behave-tests/dnf/security-upgrade.feature
+++ b/dnf-behave-tests/dnf/security-upgrade.feature
@@ -93,3 +93,67 @@ Given I use repository "security-upgrade"
       | Action        | Package        |
       | install       | F-0:2-2.x86_64 |
       | obsoleted     | E-0:1-1.x86_64 |
+
+
+@bz2124483
+Scenario: --security upgrade of a package that changes arch from noarch to x86_64
+Given I use repository "security-upgrade"
+  And I execute dnf with args "install change-arch-noarch-1-1.noarch"
+ When I execute dnf with args "upgrade --security"
+ Then the exit code is 0
+  And RPMDB Transaction is following
+      | Action        | Package                  |
+      | install       | change-arch-noarch-0:2-2.x86_64 |
+      | remove        | change-arch-noarch-0:1-1.noarch |
+  And DNF Transaction is following
+      | Action        | Package                  |
+      | upgrade       | change-arch-noarch-0:2-2.x86_64 |
+
+
+@bz2124483
+Scenario: --security upgrade of a package that changes arch from x86_64 to noarch
+Given I use repository "security-upgrade"
+  And I execute dnf with args "install change-arch-noarch-reversed-1-1.x86_64"
+ When I execute dnf with args "upgrade --security"
+ Then the exit code is 0
+  And RPMDB Transaction is following
+      | Action        | Package                           |
+      | install       | change-arch-noarch-reversed-0:2-2.noarch |
+      | remove        | change-arch-noarch-reversed-0:1-1.x86_64 |
+  And DNF Transaction is following
+      | Action        | Package                           |
+      | upgrade       | change-arch-noarch-reversed-0:2-2.noarch |
+
+
+@bz2124483
+Scenario: --security upgrade of a package that changes arch from i686 to x86_64 is not allowed
+Given I use repository "security-upgrade"
+  And I successfully execute dnf with args "install change-arch-1-1.i686"
+  # Make sure change-arch-2-2.x86_64 is available since we are testing we don't upgrade to it.
+  # It also has to have an available advisory. (We cannot verify that here because the updateinfo command is bugged when dealing with arch changes)
+  And I successfully execute dnf with args "repoquery change-arch-2-2.x86_64"
+  Then stdout is
+  """
+  <REPOSYNC>
+  change-arch-0:2-2.x86_64
+  """
+ When I execute dnf with args "upgrade --security"
+ Then the exit code is 0
+  And Transaction is empty
+
+
+@bz2124483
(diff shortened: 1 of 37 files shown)


glum23:
Scenario: --security upgrade with advisories with pkgs of different arches (noarch variant)
fails due to the step 
And I successfully execute dnf with args "install json-c-1-1 bind-libs-lite-1-1"
Error: Unable to find a match: json-c-1-1 bind-libs-lite-1-1

In fact, the related repo is empty for some reason at that time:
      Repo-id            : security-upgrade-noarch
      Repo-name          : security-upgrade-noarch test repository
      Repo-status        : enabled
      Repo-revision      : 1550000000
      Repo-updated       : Fri Sep  9 10:41:46 2022
      Repo-pkgs          : 0
      Repo-available-pkgs: 0
      Repo-size          : 0
      Repo-baseurl       : file:///opt/ci/dnf-behave-tests/fixtures/repos/security-upgrade-noarch
      Repo-expire        : 172800 second(s) (last: Fri Sep  9 10:41:47 2022)
      Repo-filename      : /tmp/dnf_ci_installroot_ex_37o8j/etc/yum.repos.d/security-upgrade-noarch.repo
      Total packages: 0

glum23:
Further observations:
Scenario: --security upgrade of a i686 package that is obsoleted by a x86_64 pkg is not allowed
  missing --security in  the step When I execute dnf with args "upgrade"
Scenario: --security upgrade with advisories with pkgs of different arches (noarch variant)
  typo in the comment: Techincally

kontura:
> Scenario: --security upgrade with advisories with pkgs of different arches (noarch variant) fails due to the step And I successfully execute dnf with args "install json-c-1-1 bind-libs-lite-1-1" Error: Unable to find a match: json-c-1-1 bind-libs-lite-1-1
> 
> In fact, the related repo is empty for some reason at that time: Repo-id : security-upgrade-noarch Repo-name : security-upgrade-noarch test repository Repo-status : enabled Repo-revision : 1550000000 Repo-updated : Fri Sep 9 10:41:46 2022 Repo-pkgs : 0 Repo-available-pkgs: 0 Repo-size : 0 Repo-baseurl : file:///opt/ci/dnf-behave-tests/fixtures/repos/security-upgrade-noarch Repo-expire : 172800 second(s) (last: Fri Sep 9 10:41:47 2022) Repo-filename : /tmp/dnf_ci_installroot_ex_37o8j/etc/yum.repos.d/security-upgrade-noarch.repo Total packages: 0
> 
> The same problem with Scenario: --security upgrade of a package that changes arch from i686 to x86_64 is not allowed No match for argument: change-arch-1-1.i686 and Scenario: --security upgrade of a package that changes arch from x86_64 to i686 is not allowed No match for argument: change-arch-reversed-1-1.x86_64 and Scenario: --security upgrade of a noarch package that is obsoleted by a x86_64 pkg No match for argument: obsoleted-change-arch-noarch-1-1.noarch and Scenario: --security upgrade of a x86_64 package that is obsoleted by a noarch pkg No match for argument: obsoleted-change-arch-noarch-reversed-1-1.x86_64 etc.

I think this was caused by not running the `fixtures/specs/build.sh` script to create test packages (or by not running in a container which does it automatically during `./container-test build`).



> Further observations: Scenario: --security upgrade of a i686 package that is obsoleted by a x86_64 pkg is not allowed missing --security in the step When I execute dnf with args "upgrade" Scenario: --security upgrade with advisories with pkgs of different arches (noarch variant) typo in the comment: Techincally

Thank you for spotting these! They should be fixed now.
I was missing the `--security` switch because I was trying out how dnf behaves in a regular update and it doesn't upgrade packages across architecture (when none of the are `noarch`) even with specific obsolete. Based on that I think even security upgrades shouldn't do it.

Also I have split the tests into multiple commits and added a bunch of new one.

glum23:
I've rerun and reviewed the test again (with packages available in security-upgrade-* repos) and I'd say that
"Scenario: --security upgrade of a i686 package that is obsoleted by a x86_64 pkg is not allowed"
has not the proper configuration of packages. When it is changed so that  "upgrade" is called instead of
"upgrade --security", the package obsoleted-change-arch.i686 is not upgraded (and it should be, otherwise
--security is useless). The obsoleter package has to provide obsoleted-change-arch.i686.

The same problem appears for the scenario
"Scenario: --security upgrade of a x86_64 package that is obsoleted by a i686 pkg is not allowed",
however, here the package obsoleted-change-arch-reversed is upgraded when --security is removed
because there's another upgrade package (x86_64) available.

kontura:
> I've rerun and reviewed the test again (with packages available in security-upgrade-* repos) and I'd say that "Scenario: --security upgrade of a i686 package that is obsoleted by a x86_64 pkg is not allowed" has not the proper configuration of packages. When it is changed so that "upgrade" is called instead of "upgrade --security", the package obsoleted-change-arch.i686 is not upgraded (and it should be, otherwise --security is useless). The obsoleter package has to provide obsoleted-change-arch.i686.
> 
> The same problem appears for the scenario "Scenario: --security upgrade of a x86_64 package that is obsoleted by a i686 pkg is not allowed", however, here the package obsoleted-change-arch-reversed is upgraded when --security is removed because there's another upgrade package (x86_64) available.


I see what you mean but I am afraid the provide won't help. I have looked into it and libsolv has the [same default](https://github.com/openSUSE/libsolv/blob/master/src/policy.c#L1409-L1410) rule where it doesn't allow changing of architecture other than when noarch is involved (even with obsoletes). 
There is a flag `SOLVER_FLAG_ALLOW_ARCHCHANGE` for it but libdnf doesn't use it.

I could add `obsoleter-change-arch-1-1.i686` to the first test so it would match with the second "Scenario: --security upgrade of a x86_64 package that is obsoleted by a i686 pkg is not allowed" and it would upgrade to it without `--security` but not with it because there would not be an advisory for it.
It is a different situation but I cannot think of any better approach. What do you think?

glum23:
OK, I think it's a good suggestion. It will be evident that 'upgrade --security' differs from 'upgrade' and at the same time an improper arch is not used. Thanks

kontura:
- I have added the new `obsoleter-change-arch-1-1.i686` rpm as discussed (by renaming `obsoleter-change-arch-1-1.spec` to `obsoleter-change-arch-1-1.i686.spec` which creates both x86_64 and i686 versions).

- I have also added a bunch of checks and comments to make it clear what test data are required if it wasn't visible from the test itself.

dnf-behave-tests/dnf/security-upgrade.feature
```diff
@@ -13,6 +13,36 @@ Given I use repository "security-upgrade"
...
+@bz2124483
+# This scenario is the same as the one above just with noarch packages present in the
+# repo and advisory instead of i686 this causes different behavior - it fails.
+# Technically we could get bz2088149 again with just a different package set.
+# However while this is possible it shouldn't happen because noarch packages are special - they work
+# on all architectures and we shouldn't see noarch builds together with arch specific builds as it
+# is setup in this scenario (in repo security-upgrade-noarch).
+Scenario: --security upgrade with advisories with pkgs of different arches (noarch variant)
+Given I use repository "security-upgrade-noarch"
+  And I successfully execute dnf with args "install json-c-1-1 bind-libs-lite-1-1"
```
j-mracek, line 25: Please could you add then step with check of the transaction table? It will be more transparent what (architecture) is installed on the system

kontura: Sure, added.
I have also added a link to a PR with more details about the tests situation.

j-mracek: Perfect

j-mracek (approved):
LGTM

PR #1155, opened 2022-09-08
Tests for rollback of group upgrade transaction rollback

jan-kolarik:
Requires https://github.com/rpm-software-management/dnf/pull/1852.

### DIFF
--- a/dnf-behave-tests/dnf/history-rollback-env-upgrade.feature
+++ b/dnf-behave-tests/dnf/history-rollback-env-upgrade.feature
@@ -16,6 +16,14 @@ Background:
     And History is following
         | Id     | Command                                              | Action        | Altered   |
         | 1      | group install DNF-CI-RollbackTestEnv                 | Install       | 7         |
+   When I execute dnf with args "group list --installed DNF-CI-RollbackTestEnv"
+   Then the exit code is 0
+    And stdout is
+    """
+    <REPOSYNC>
+    Installed Environment Groups:
+       DNF-CI-RollbackTestEnv
+    """
   Given I use repository "dnf-ci-group-rollback-2"
     And I successfully execute dnf with args "group upgrade DNF-CI-RollbackTestEnv"
    Then Transaction is following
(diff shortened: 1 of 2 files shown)


pkratoch:
The change allows not only the undo of the history upgrade rollback, but also the redo (so, the sequence of actions is: `dnf group upgrade`, `dnf history undo last`, `dnf history redo last`). It's probably not a common use case, but can you add tests also for that?

jan-kolarik:
@pkratoch Thanks for the feedback. I was thinking about the sequence of actions you proposed - `upgrade`, `undo last`, `redo last` - and I don't understand what the last operation should mean? Intuitively when I do `undo A` I go to the time point before the A and when I want to undo that I will do the A operation again. That's fine and this is working now. But when you want to redo `undo A` it means do nothing, right? :slightly_smiling_face: I've also check the similar sequence: `install package`, `undo last`, `redo last` for a single package with the current DNF version and it doesn't work either.

pkratoch:
> @pkratoch Thanks for the feedback. I was thinking about the sequence of actions you proposed - `upgrade`, `undo last`, `redo last` - and I don't understand what the last operation should mean? Intuitively when I do `undo A` I go to the time point before the A and when I want to undo that I will do the A operation again. That's fine and this is working now. But when you want to redo `undo A` it means "do nothing", right? slightly_smiling_face

Yes, it's not usable in this context, but I imagine you could have scenario where you would want to replicate list of transactions on another system or you'd want to complete saved transaction or something like that (I remember an RFE for this, but I don't know if and how it was implemented at the end). In either case, you don't want to fail with error that the action is unknown (as it was before your patch).

> I've also check the similar sequence: `install package`, `undo last`, `redo last` for a single package with the current DNF version and it doesn't work either.

Hm, I think it works, there is just nothing to do... or did I misunderstand what you meant?
```
# dnf install nano
# dnf history undo last
# dnf history redo last
Last metadata expiration check: 0:00:18 ago on Tue Sep 13 13:16:19 2022.
Dependencies resolved.
Nothing to do.
Complete!
Warning, the following problems occurred while running a transaction:
  Package nevra "nano-6.0-2.fc36.x86_64" not installed for action "Removed".
# echo $?
0
```

jan-kolarik:
@pkratoch Thank you, I understand now :slightly_smiling_face: I will prepare the additional test for that.

> Hm, I think it works, there is just nothing to do... or did I misunderstand what you meant?

You're right, I've just seen the `Warning` there and overlooked that the whole operation exited successfully.

PR #1157, opened 2022-09-13
Test that user-installed package is not removed as unused dep

m-blaha:
for dnf5 requires https://github.com/rpm-software-management/dnf5/pull/35


kontura:
Tested locally with up to date dnf5 and it works.
Thanks for the test!

PR #1158, opened 2022-09-15
Tests for dnf5's history list command

inknos:
https://github.com/rpm-software-management/dnf5/pull/26

Test coverage is partial.

- added step dnf4/5 history list is (there are still some problems on history range when is reverse)
- fixed step dnf4/5 exit code is
- added <HELP> keyword to parse help stdout (only implemented for dnf5 now)

### DIFF
--- a/dnf-behave-tests/common/output.py
+++ b/dnf-behave-tests/common/output.py
@@ -36,6 +36,20 @@ def strip_reposync_dnf5(found_lines, line_number):
     if line_number < len(found_lines) and found_lines[line_number].strip() == "Repositories loaded.":
         found_lines.pop(line_number)
 
+def strip_help_dnf4(found_lines, line_number):
+    pass
+
+def strip_help_dnf5(found_lines, line_number):
+    end_of_help_dnf5 = "--disableplugin=PLUGIN_NAME,...        Alias for '--disable-plugin'"
+    if line_number < len(found_lines) and (found_lines[line_number].strip() == "Missing command" or found_lines[line_number].strip() == "Usage:" ):
+        found_lines.pop(line_number)
+
+    while line_number < len(found_lines) and found_lines[line_number].strip() != end_of_help_dnf5:
+        found_lines.pop(line_number)
+
+    if line_number < len(found_lines) and found_lines[line_number].strip() == end_of_help_dnf5:
+        found_lines.pop(line_number)
+
 
 def handle_reposync(expected, found, dnf5_mode):
     line_number = 0
@@ -45,6 +59,15 @@ def handle_reposync(expected, found, dnf5_mode):
                 strip_reposync_dnf5(found, line_number)
             else:
                 strip_reposync_dnf4(found, line_number)
+
+            expected.pop(line_number)
+            break
+
+        if line == "<HELP>":
+            if dnf5_mode:
+                strip_help_dnf5(found, line_number)
+            else:
+                strip_help_dnf4(found, line_number)
 
             expected.pop(line_number)
             break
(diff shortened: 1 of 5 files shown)


dnf-behave-tests/dnf/steps/lib/dnf.py
```diff
@@ -310,6 +310,35 @@ def parse_history_list(lines):
...
+        command = command.replace(action, "")
+        command = command.replace(altered, "")
+        result.append({
+            "_line": line,
+            "id": idd.strip(),
+            "command": command.strip(),
+            "date": date.strip(),
+            "action": action.strip(),
+            "altered": altered.strip()
+        })
```
lukash, line 338: This approach of parsing the history list is extremely brittle and error-prone. It should still be possible to parse the lines with a single regex, take a look at https://github.com/lukash/ci-dnf-stack/commit/1873b8a7cf88194a67bf1728328bb97810532b87 as an inspiration (it's an implementation of the history info step, it's waiting on the `history last` PR to be posted).

dnf-behave-tests/dnf/steps/history.py
```diff
@@ -73,10 +73,27 @@ def history_equal(history, table):
 
 
 @behave.then('stdout is history list')
-def step_impl(context):
+def then_stdout_is_history_list(context):
     assert_history_list(context, context.cmd_stdout)
 
 
+@behave.then("{dnf_version:dnf_version} stdout is history list")
```
lukash, line 80: Is `{dnf_version:dnf_version}` really necessary? Isn't `{dnf_version}` enough?

dnf-behave-tests/common/output.py
```diff
@@ -36,6 +36,20 @@ def strip_reposync_dnf5(found_lines, line_number):
     if line_number < len(found_lines) and found_lines[line_number].strip() == "Repositories loaded.":
         found_lines.pop(line_number)
 
+def strip_help_dnf4(found_lines, line_number):
+    pass
+
+def strip_help_dnf5(found_lines, line_number):
+    end_of_help_dnf5 = "--disableplugin=PLUGIN_NAME,...        Alias for '--disable-plugin'"
```
lukash, line 43: I understand your intention and I don't know what to do about this exactly, but this is too specific and will break too often...

dnf-behave-tests/common/output.py
```diff
@@ -36,6 +36,20 @@ def strip_reposync_dnf5(found_lines, line_number):
...
+        found_lines.pop(line_number)
+
+    while line_number < len(found_lines) and found_lines[line_number].strip() != end_of_help_dnf5:
+        found_lines.pop(line_number)
+
+    if line_number < len(found_lines) and found_lines[line_number].strip() == end_of_help_dnf5:
+        found_lines.pop(line_number)
+
 
 def handle_reposync(expected, found, dnf5_mode):
```
lukash, line 54: If you want to also handle the `<HELP>` here, some renaming needs to be done to keep the naming of methods accurate.

dnf-behave-tests/common/output.py
```diff
@@ -66,6 +89,38 @@ def then_stdout_is_empty(context):
...
+@behave.then("stdout is help")
+def then_stdout_is_help(context):
+    expected = context.text.format(context=context).rstrip().split('\n')
+    found = context.cmd_stdout.rstrip().split('\n')
+
+    if found == [""]:
+        found = []
+
+    dnf5_mode = hasattr(context, "dnf5_mode") and context.dnf5_mode
+    clean_expected, clean_found = handle_help(expected, found, dnf5_mode)
```
lukash, line 101: The `handle_help()` method doesn't exist, and the `stdout is help` step (or its dnf4/dnf5 variants) is never used...

dnf-behave-tests/dnf/history-list-dnf5.feature
```diff
@@ -0,0 +1,147 @@
+# This tests are somehow redundat for dnf4. the features
+# are tested for history command. Example:
+# history list last-1 has the same behavior as history last-1
+# in dnf4
+#
+# This feature file can be merged with history-list.feature
+# in the future if the behavior of history command will change
+# assuming list by default
+@dnf5
+Feature: history list list
```
lukash, line 10: I don't like the duplication of tests in this file. The reason here is dnf4 supports omitting the subcommand and defaults to `list`, and there's a bunch of tests like `dnf history last`. I think it would be much better to rewrite the tests to include the subcommand: `dnf history list last` and then, only for dnf4 specifically test that the shorter version works. You should only need a couple of tests for that, and the rest of the tests should be shared if possible.

The way it is now, the dnf5 tests are split between the old file and this new one, and it's hard to make sense of it and see what is tested where and for which version, etc. It adds chaos which will be hard and time-consuming to maintain.

dnf-behave-tests/dnf/history-list-dnf5.feature
```diff
@@ -0,0 +1,147 @@
...
+  And I successfully execute dnf with args "remove abcde"
+  And I successfully execute dnf with args "install nodejs"
+
+
+Scenario: history list last-1
+ When I execute dnf with args "history list last-1"
+ Then the exit code is 0
+  And stdout is history list
+      | Id | Command | Action  | Altered |
+      | 2  |         | Removed | 3       |
```
lukash, line 25: Case in point, this test is already in the `history-list.feature` file. I wouldn't like to go over the two files and try to match which test is there, what's missing and whats actually redundant...

dnf-behave-tests/dnf/steps/history.py
m-blaha: I thing it is good. What's confusing here is that the second `dnf_version` is a type with parser registered here: https://github.com/rpm-software-management/ci-dnf-stack/blob/5295644b1b6652765aefbe58341a8f6dba9ea2d3/dnf-behave-tests/common/output.py#L110
Having this type enforces that only `dnf4` and `dnf5` values of this parameter are accepted.

PR #1159, opened 2022-09-19
Ensure `CQRlib` is reinstalled from `dnf-ci-fedora-updates-testing`

kontura:
Since the original repo was still available it could be used for the reinstall -> we need to drop it to ensure the newly added repo is used.

This way the test should not longer be nondeterministic.

This started failing for: https://github.com/rpm-software-management/dnf5/pull/41


jan-kolarik:
Great, tested locally passing. Thanks for the fix :slightly_smiling_face:

PR #1160, opened 2022-09-20
Enable passing tests with "--allowerasing" for dnf5

jrohel:


dnf-behave-tests/dnf/protect-running-kernel.feature
```diff
@@ -70,7 +70,7 @@ Scenario: Running kernel is not protected against obsoleting with config protect
 
 
 # @dnf5
-# TODO(nsella) Unknown argument "--allowerasing" for command "install"
+# TODO(jrohel) different stdout
```
lukash, line 73: You can use two steps, `And dnf4 stdout is` and `And dnf5 stdout is` to fix this right away.

jrohel: Thank you for the tip. Fixed.

PR #1161, opened 2022-09-21
[dnf5] A step for checking the history transaction records

lukash:
Adds transaction history tests for all scenarios that test the system state, meaning both system state and the transaction history are tested to be correct and consistent for these scenarios.

Requires: https://github.com/rpm-software-management/dnf5/pull/53

### DIFF
--- a/dnf-behave-tests/dnf/downgrade.feature
+++ b/dnf-behave-tests/dnf/downgrade.feature
@@ -55,6 +55,14 @@ Scenario: Downgrade RPM that requires downgrade of dependency
         | glibc-all-langpacks-2.28-9.fc29.x86_64 | Dependency | dnf-ci-fedora |
         | glibc-common-2.28-9.fc29.x86_64        | Dependency | dnf-ci-fedora |
         | setup-2.12.1-1.fc29.noarch             | Dependency | dnf-ci-fedora |
+    And dnf5 transaction items for transaction "last" are
+        | action    | package                                   | reason       | repository    |
+        | Downgrade | glibc-0:2.28-9.fc29.x86_64                | User         | dnf-ci-fedora |
+        | Downgrade | glibc-common-0:2.28-9.fc29.x86_64         | Dependency   | dnf-ci-fedora |
+        | Downgrade | glibc-all-langpacks-0:2.28-9.fc29.x86_64  | Dependency   | dnf-ci-fedora |
+        | Replaced  | glibc-0:2.28-26.fc29.x86_64               | User         | @System       |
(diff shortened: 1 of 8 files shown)


dnf-behave-tests/dnf/remove-pkgspec.feature
```diff
@@ -29,6 +29,14 @@ Scenario Outline: Remove an RPM by <pkgspec-type>
...
     And package state is
         | package | reason | from_repo |
+    And transaction items for transaction "last" are
+        | action | package                                  | reason | repository |
+        | Remove | basesystem-0:11-6.fc29.noarch            | Clean  | @System    |
+        | Remove | filesystem-0:3.9-2.fc29.x86_64           | Clean  | @System    |
+        | Remove | glibc-0:2.28-9.fc29.x86_64               | User   | @System    |
+        | Remove | glibc-all-langpacks-0:2.28-9.fc29.x86_64 | Clean  | @System    |
+        | Remove | glibc-common-0:2.28-9.fc29.x86_64        | Clean  | @System    |
+        | Remove | setup-0:2.12.1-1.fc29.noarch             | Clean  | @System    |
```
j-mracek, line 39: It is also a problem of dnf4 - it uses as repository for remove action always `@System`. Be honest it is not informative value at all.

lukash: Maybe this could be changed to `from_repo` for outbound actions...? not the goal of this PR though.

dnf-behave-tests/dnf/downgrade.feature
```diff
@@ -55,6 +55,14 @@ Scenario: Downgrade RPM that requires downgrade of dependency
         | glibc-all-langpacks-2.28-9.fc29.x86_64 | Dependency | dnf-ci-fedora |
         | glibc-common-2.28-9.fc29.x86_64        | Dependency | dnf-ci-fedora |
         | setup-2.12.1-1.fc29.noarch             | Dependency | dnf-ci-fedora |
+    And transaction items for transaction "last" are
```
j-mracek, line 58: Because the check is only for DNF5 it would be nice to add that information even to the step syntax. And because the check uses history output it could be beneficial to somehow mention it.
What about `history transaction items for DNF5 transaction "last" are`? But there are multiple ways how it can be formulated.

lukash: I've prepended "dnf5 " to the name of the step, we already have "dnf5 stdout/stderr is"...

PR #1163, opened 2022-09-26
createrepo_c: using a dnf command: disable plugins and use installroot

kontura:
This prevents fails when the host system has specific dnf configuration (such as using subscription manager) and it prints additional messeages to stdout.


mbanas (commented):
LGTM, restarted the suite with this patch and it passed.

jan-kolarik:
LGTM too, thanks for the fix.

PR #1164, opened 2022-09-29
createrepo_c: using a dnf command: disable plugins and use installroot

kontura:
This prevents fails when the host system has specific dnf configuration (such as using subscription manager) and it prints additional messeages to stdout.


kontura:
This is a backport to 9.2 branch and the createrepo_c tests failed in an expected way with upstream createrepo_c.

PR #1166, opened 2022-10-10
Run upgrade before distro-sync in case there is a new dependency

kontura:
If dnf-nighly has a new version that has a new dependency the distory-sync will fail because it is limited to the dnf-nightly repo.

Lets keep the distro-sync command even though it will mostly do nothing but there could teoretically be a situation where the fedora repo has newer packages than nighly and since we are running in a nightly mode we want to make sure we have nighly packages.


jan-kolarik:
LGTM

PR #1167, opened 2022-10-11
Add tests for download

inknos:
- test option resolve
- test option alldeps

### DIFF
--- a/dnf-behave-tests/dnf/dnf5/download.feature
+++ b/dnf-behave-tests/dnf/dnf5/download.feature
@@ -1,0 +1,63 @@
+Feature: Download commands with options --resolve --alldeps
+
+Background:
+  Given I use repository "dnf-ci-fedora"
+
+@not.with_dnf=4
+@dnf5
+Scenario: Download a single rpm
+  Given I set working directory to "{context.dnf.tempdir}"
+   When I execute dnf with args "download abcde"
+   Then the exit code is 0
+    And file "/{context.dnf.tempdir}/abcde-2.9.2-1.fc29.src.rpm" exists
+    And file "/{context.dnf.tempdir}/abcde-2.9.2-1.fc29.noarch.rpm" exists
+
+
+@not.with_dnf=4
+@dnf5
+Scenario: Download with resolve option when no package is installed
+  Given I set working directory to "{context.dnf.tempdir}"
+   When I execute dnf with args "download abcde --resolve"
+   Then the exit code is 0
+    And file "/{context.dnf.tempdir}/abcde-2.9.2-1.fc29.src.rpm" exists
+    And file "/{context.dnf.tempdir}/abcde-2.9.2-1.fc29.noarch.rpm" exists
+    And file "/{context.dnf.tempdir}/flac-1.3.2-8.fc29.x86_64.rpm" exists
+    And file "/{context.dnf.tempdir}/wget-1.19.5-5.fc29.x86_64.rpm" exists
+
+
+@not.with_dnf=4
+@dnf5
+Scenario: Download with resolve option when a dependency is installed
+  Given I set working directory to "{context.dnf.tempdir}"
+    And I successfully execute dnf with args "install wget"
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/dnf5/download.feature
```diff
@@ -0,0 +1,60 @@
...
+   When I execute dnf with args "download abcde --resolve --alldeps"
+   Then the exit code is 0
+    And file "/{context.dnf.tempdir}/abcde-2.9.2-1.fc29.src.rpm" exists
+    And file "/{context.dnf.tempdir}/abcde-2.9.2-1.fc29.noarch.rpm" exists
+    And file "/{context.dnf.tempdir}/flac-1.3.2-8.fc29.x86_64.rpm" exists
+    And file "/{context.dnf.tempdir}/wget-1.19.5-5.fc29.x86_64.rpm" exists
+
+@not.with_dnf=4
+@dnf5
+Scenario: Download with alldeps options
```
jan-kolarik, line 53: This is the same test as the first one, probably just forgotten code block?

jan-kolarik: And also I thought the ``--alldeps`` is not intended to be used without the ``--resolve``.

dnf-behave-tests/dnf/dnf5/download.feature
```diff
@@ -0,0 +1,60 @@
...
+@not.with_dnf=4
+@dnf5
+Scenario: Download with resolve option when a dependency is installed
+  Given I set working directory to "{context.dnf.tempdir}"
+    And I successfully execute dnf with args "install wget"
+   When I execute dnf with args "download abcde --resolve"
+   Then the exit code is 0
+    And file "/{context.dnf.tempdir}/abcde-2.9.2-1.fc29.src.rpm" exists
+    And file "/{context.dnf.tempdir}/abcde-2.9.2-1.fc29.noarch.rpm" exists
+    And file "/{context.dnf.tempdir}/flac-1.3.2-8.fc29.x86_64.rpm" exists
```
jan-kolarik, line 37: I was just thinking if not to add the explicit check that ``wget`` package does not exist, like:

``And file "/{context.dnf.tempdir}/wget-1.19.5-5.fc29.x86_64.rpm" does not exist``

jan-kolarik (changes requested):

dnf-behave-tests/dnf/dnf5/download.feature
inknos: > This is the same test as the first one, probably just forgotten code block?

my bad, I forgot to add `--alldeps`

> And also I thought the --alldeps is not intended to be used without the --resolve.

It should throw an error as we decided. It wasn't a defined behaviour at the time of writing those tests. I will change the scenario according to our decision.

dnf-behave-tests/dnf/dnf5/download.feature
inknos: good idea, I would add it

dnf-behave-tests/dnf/dnf5/download.feature
inknos: This is needed, for the tests too
https://github.com/rpm-software-management/dnf5/pull/185

dnf-behave-tests/dnf/dnf5/download.feature
```diff
@@ -0,0 +1,59 @@
...
+    And file "/{context.dnf.tempdir}/wget-1.19.5-5.fc29.x86_64.rpm" exists
+
+@not.with_dnf=4
+@dnf5
+Scenario: Download with alldeps options
+  Given I set working directory to "{context.dnf.tempdir}"
+   When I execute dnf with args "download abcde --alldeps"
+   Then the exit code is 2
```
jan-kolarik, line 58: I would add also the check for the `stderr` here ...

jan-kolarik:
Thanks!

PR #1172, opened 2022-11-01
offline-upgrade: Add security filters tests

jan-kolarik:
Add security filters tests for the `offline-upgrade` command.

Requires PR: https://github.com/rpm-software-management/dnf-plugins-core/pull/465

### DIFF
--- a/dnf-behave-tests/dnf/plugins-core/offline-upgrade-security-filters.feature
+++ b/dnf-behave-tests/dnf/plugins-core/offline-upgrade-security-filters.feature
@@ -1,0 +1,341 @@
+Feature: Test the security filters for offline-upgrade commands
+
+
+Background:
+  Given I enable plugin "system-upgrade"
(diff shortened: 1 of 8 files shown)


dnf-behave-tests/dnf/plugins-core/offline-upgrade-security-filters.feature
```diff
@@ -0,0 +1,340 @@
...
+      ================================================================================
+       Package            Architecture   Version        Repository               Size
+      ================================================================================
+      Upgrading:
+       advisory_B         x86_64         1.0-4          dnf-ci-security         6.0 k
+      
+      Transaction Summary
+      ================================================================================
+      Upgrade  1 Package
+      """
```
pkratoch, line 339: Can you add a package with a lower-severity security update to also test that only "Critical" security updates are applied?

PR #1175, opened 2022-11-09
Add dnf5 only tests for rich dependencies

jan-kolarik:
Tests for handling rich dependencies in DNF5.

Related PR: https://github.com/rpm-software-management/dnf5/pull/125

### DIFF
--- a/dnf-behave-tests/fixtures/specs/dnf-ci-rich/porridge-1.0-1.spec
+++ b/dnf-behave-tests/fixtures/specs/dnf-ci-rich/porridge-1.0-1.spec
@@ -8,7 +8,7 @@ URL:            None
 
 Summary:        The made up package for rich dependencies testing.
 
-Requires:       (milk if (float or oat))
+Requires:       (milk if (flour or oat))
 
 %description
 Rich dependencies testing package.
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/fixtures/specs/dnf-ci-rich/porridge-1.0-1.spec
```diff
@@ -8,7 +8,7 @@ URL:            None
 
 Summary:        The made up package for rich dependencies testing.
 
-Requires:       (milk if (float or oat))
+Requires:       (milk if (flour or oat))
```
j-mracek, line 11: I guess that the change is not required by new scenarios. But it looks like a typo in require, but I am not sure. Otherwise it looks good.

jan-kolarik: Yes, it is not required, but the mistake irritated me :smiley: Most probably a typo, because in tests comments it appears correctly using the `flour`.

j-mracek: May be moving it into another commit would be better. I am curious why the change did not break any other test.

PR #1185, opened 2022-11-22
libdnf5 python API: add steps and first two tests

kontura:
It adds two steps:

One general step to execute arbitrary python code, this should be
used if we want to test setting up base, config or repositories API.
It could also potentially be used for dnf4 api tests.

One more specific step for libdnf5 that already has setup base, config
and repositories in a default way with installroot that a lot of tests
would need to do.

Originally I wanted to create the new directory for tests (`api`) outside
 the `dnf` directory (as agreed) but given that the api tests will need
basically the same infrastructure as the dnf tests it makes sense to
keep it together. Otherwise we would have to duplicate quite a lot of
the setup and move basically all dnf steps in the `common` steps
directory.

### DIFF
--- a/dnf-behave-tests/dnf/api/package-query.feature
+++ b/dnf-behave-tests/dnf/api/package-query.feature
@@ -1,0 +1,34 @@
+@dnf5
+@not.with_dnf=4
+Feature: api: query packages
+
+
+Scenario: Construct query and filter labirinto package
+Given I use repository "simple-base"
+ When I execute python libdnf5 api script with setup
+      """
+      query = libdnf5.rpm.PackageQuery(base)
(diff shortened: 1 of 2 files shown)


glum23 (commented):
The test dnf/api/package-query.feature passed on Fedora37, however
* it is necessary to install dnf5, libdnf5, and python3-libdnf5 manually (the packages have to be added to requirements when a separate branch for dnf5 will be created)
* the test is disabled by default for systems with dnf4 (it has to be explicitly required to be run)
The test failed on rawhide probably due to an error in behave, it has to be further investigated and fixed.
From my POW the PR can be merged.

PR #1186, opened 2022-11-24
Enable group command related tests for dnf5

m-blaha:
Requires: https://github.com/rpm-software-management/dnf5/pull/170

### DIFF
--- a/dnf-behave-tests/dnf/comps-group.feature
+++ b/dnf-behave-tests/dnf/comps-group.feature
@@ -31,12 +31,14 @@ Scenario: Install and remove group using group name
         | group-remove  | DNF-CI-Testgroup                  |
 
 
-# @dnf5
-# TODO(nsella) Unknown argument "install" for command "group"
+# The rest of scenarios use a group id to specify the group.
+# This way scenarios can be used for both dnf4 and dnf5 testing.
+
+@dnf5
 Scenario: Install and remove group
   Given I use repository "dnf-ci-thirdparty"
     And I use repository "dnf-ci-fedora"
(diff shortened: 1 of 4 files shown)


dnf-behave-tests/dnf/comps-group.feature
```diff
@@ -452,7 +456,7 @@ Scenario: Install a group with empty packagelist
 
 
 # @dnf5
-# TODO(nsella) Unknown argument "install" for command "group"
+# TODO(nsella) Merged group produces different package set
```
pkratoch, line 459: In dnf5, only the first solvable is used to get the package list ( https://github.com/rpm-software-management/dnf5/blob/main/libdnf/comps/group/group.cpp#L124 ). For a while, I wondered if it was a design decision, but then I found it in this unmerged PR that attempted to document design decisions: https://github.com/rpm-software-management/libdnf/pull/1350/files#diff-f313c151f0caffdb2b4fe4a02adf2bde652498e2cb1d29296d153e67b7c29e77R35
So, I guess it was done only as a first iteration and should be changed. I can submit an issue, if you agree.

m-blaha: Oh, I completely missed this PR. We should keep such design decisions documented somewhere - it contains useful information and in unmerged PR it is completely hidden...
As far as the issue is concerned, I agree, please go ahead and submit it.

pkratoch: https://github.com/rpm-software-management/dnf5/issues/183

PR #1189, opened 2022-12-02
dnf5 API testing transactions

glum23:
added a helper function into dnf/steps/cmd.py for testing dnf5 API transactions
added two tests for testing transactions
passed on F37 (when dnf5 and python3-libdnf5 are installed and @not.with_dnf=4 is disabled)

### DIFF
--- a/dnf-behave-tests/dnf/api/transaction-remove-install.feature
+++ b/dnf-behave-tests/dnf/api/transaction-remove-install.feature
@@ -1,0 +1,24 @@
+@dnf5
+@not.with_dnf=4
+Feature: transaction: remove a package with dependency, then install the dependency again
+
+
+Scenario: Construct query, remove vagare, then install labirinto
+Given I use repository "simple-base"
+Given I successfully execute dnf with args "install vagare"
+# vagare depends on labirinto
+ When I execute python libdnf5 api script with setup
+      """
+      goal = libdnf5.base.Goal(base)
+      goal.add_rpm_remove("vagare")
+      goal.add_rpm_install("labirinto")
+      execute_transaction(goal, "remove and install")
+      """
+ Then the exit code is 0
+  And stdout is
+      """
+      vagare-1.0-1.fc29.x86_64 : Remove
+      """
+  And RPMDB Transaction is following
+      | Action        | Package                       |
+      | remove        | vagare-0:1.0-1.fc29.x86_64    |
--- a/dnf-behave-tests/dnf/api/transaction-simple-install.feature
+++ b/dnf-behave-tests/dnf/api/transaction-simple-install.feature
@@ -1,0 +1,21 @@
+@dnf5
+@not.with_dnf=4
+Feature: transaction: install a package without dependencies
+
+
+Scenario: Construct query and install labirinto package
+Given I use repository "simple-base"
+ When I execute python libdnf5 api script with setup
+      """
+      goal = libdnf5.base.Goal(base)
+      goal.add_rpm_install("labirinto")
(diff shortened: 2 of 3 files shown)


dnf-behave-tests/dnf/api/transaction-simple-install.feature
```diff
@@ -0,0 +1,21 @@
...
+      goal = libdnf5.base.Goal(base)
+      goal.add_rpm_install("labirinto")
+      execute_transaction(goal, "install a package without dependencies")
+      """
+   Then the exit code is 0
+    And stdout is
+ """
+ labirinto-1.0-1.fc29.x86_64 : Install
+ """
+ When I execute rpm with args "-q labirinto"
```
kontura, line 19: We have a step that checks changes that rpm sees.

We cannot use `Transaction is following` directly because that also parses output of dnf but in both scenarios we can:
```Gherkin
And RPMDB Transaction is following
    | install       | labirinto-0:1.0-1.fc29.x86_64 |
```
The step is checking differences in rpm db.

glum23: OK, I'll try whether it works. I had sth like this (definitely not exactly "RPMDB Transaction is following") in the code and it didn't work for me.

dnf-behave-tests/dnf/api/transaction-remove-install.feature
```diff
@@ -0,0 +1,27 @@
...
+   Then the exit code is 0
+    And stdout is
+ """
+ vagare-1.0-1.fc29.x86_64 : Remove
+ """
+ When I execute rpm with args "-q vagare"
+   Then the exit code is 1
+   And stdout contains "package vagare is not installed"
+ When I execute rpm with args "-q labirinto"
+   Then the exit code is 0
```
kontura, line 26: Could you align the steps more uniformly?
Usually we have all the names of steps (after the `When`/`Given`... keywords) staring at the same column:
```Gherkin
Given I successfully execute dnf with args "install vagare"
# vagare depends on labirinto
 When I execute python libdnf5 api script with setup
      """
      goal = libdnf5.base.Goal(base)
      goal.add_rpm_remove("vagare")
      goal.add_rpm_install("labirinto")
      execute_transaction(goal, "remove and install")
      """
 Then the exit code is 0
  And stdout is
      """
      vagare-1.0-1.fc29.x86_64 : Remove
      """
```
(You can also check other existing features.)

dnf-behave-tests/dnf/steps/cmd.py
```diff
@@ -440,12 +440,28 @@ def execute_python_libdnf5_api_script_with_setup(context):
     appended to already setup base with installroot and loaded
     available repositories.
     """
+
```
kontura, line 443: I would prefer not to add a new line here.

kontura (changes requested):

dnf-behave-tests/dnf/api/transaction-remove-install.feature
glum23: Oops, sorry, I've edited the files on a testing machine with weirdly set tabs and vim and I haven't checked them after scp.

dnf-behave-tests/dnf/api/transaction-simple-install.feature
glum23: should be fixed now

kontura (commented):
One last thing, can you squash the 3rd fixing commit into the previous ones?
I think it is better to have a cleaner history (its easier to do backports and to understand).

Otherwise it looks good to me. :+1:

PR #1190, opened 2022-12-05
offline-upgrade: Add security filters tests (backport to 8.8.0)

jan-kolarik:
Backport https://github.com/rpm-software-management/ci-dnf-stack/pull/1172 to `8.8.0` branch.

After locally testing using the RHEL8 VM, some additional changes were needed. I've added a commit from the https://github.com/rpm-software-management/ci-dnf-stack/pull/1191.

PR #1191, opened 2022-12-06
offline-upgrade: Fix security filters tests to use transaction tables

jan-kolarik:
Comparing against the exact package sizes is not a good idea as it can vary on different systems. Therefore checking the transaction table instead.

This was initiated after testing the existing tests locally not passing on the RHEL 8.8.0 branch.


jan-kolarik:
Merging as already reviewed and tested by QE for `8.8.0` here https://github.com/rpm-software-management/ci-dnf-stack/pull/1190.

PR #1192, opened 2022-12-06
offline-upgrade: Add security filters tests (backport to 9.2.0)

jan-kolarik:
Backport https://github.com/rpm-software-management/ci-dnf-stack/pull/1172 and https://github.com/rpm-software-management/ci-dnf-stack/pull/1191 to 9.2.0 branch.


glum23 (commented):
It seems to be OK for me, however the related issue hasn't been fixed in 9.2 yet so the testsuite will fail due to this test. It either should be fixed in 9.2 immediately or the tag @not.with_os=rhel__eq__9 has to be added (and removed just after the fix).

martinobanana:
It's now safe to backport to 9.2 since the feature is present.

PR #1197, opened 2022-12-14
Separate dnf4/dnf5 CI container dependencies

kontura:
This would also have to be added to dnf5: https://github.com/rpm-software-management/dnf5/pull/195

Another approach could be just do the branching first (separate branch for dnf4 and dnf5) and then hard-code it. It would be simpler approach.


kontura:
Closing in favor of:
https://github.com/rpm-software-management/ci-dnf-stack/pull/1205
https://github.com/rpm-software-management/ci-dnf-stack/pull/1204

PR #1201, opened 2023-01-05
Check package returns correct reason in Python API

jan-kolarik:
This test is intended to check if the reason value of package from query is correctly wrapped by the SWIG.

Requires https://github.com/rpm-software-management/dnf5/pull/219
Tests rpm-software-management/dnf5#216


jan-kolarik:
Hmm, now I am thinking about it and the better place should be right in the `dnf5` component tests, so the problem is discovered more earlier.

PR #1205, opened 2023-01-11
Remove requirements for dnf5

kontura:
This branch is for dnf4 tests, there is no need to install dnf5 deps.


kontura:
The failing `dnf5daemon` tests are expected. In fact the dnf5 CI runs are being disabled by this PR but that will take effect only once the PR is merged.

jan-kolarik:
Looks OK, just one question - so we want to keep the other existing `dnf5` things, like tags and other `dnf5` specific directives?

kontura:
I am not sure, I tried to do the most minimal changes to make it work without dnf5 dependencies.
It might be nice to have the branches as similar as possible in case we want to backport some tests.

But yes there is `dnf5` things that are not relevant for the branch anymore.

PR #1210, opened 2023-01-24
Test DNF-Automatic reboot option

evan-goode:
Requires https://github.com/rpm-software-management/dnf/pull/1879

### DIFF
--- a/dnf-behave-tests/dnf/dnf-automatic/reboot.feature
+++ b/dnf-behave-tests/dnf/dnf-automatic/reboot.feature
@@ -1,0 +1,114 @@
+@no_installroot
+Feature: dnf-automatic reboots
+
+
+Background:
+Given I delete file "/etc/yum.repos.d/*.repo" with globs
+  And I create file "/etc/dnf/dnf.conf" with
+    """
+    [main]
+    plugins=0
+    """
+  And I use repository "simple-base"
+  And I successfully execute dnf with args "install labirinto"
+  And I use repository "simple-updates"
+
+Scenario: dnf-automatic does not reboot when reboot = never
+  Given I create file "/etc/dnf/automatic.conf" with
+    """
+    [commands]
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/dnf-automatic/reboot.feature
```diff
@@ -0,0 +1,101 @@
...
+    plugins=0
+    """
+
+
+Scenario: dnf-automatic does not reboot when reboot = never
+  Given I create file "/etc/dnf/automatic.conf" with
+    """
+    [commands]
+    reboot = never
+    reboot_command = "echo 'I LOVE REBOOTING'"
```
martinpitt, line 19: This is a nice way to test the various scenarios. If you also want to test the default command, either cancel the scheduled reboot (`shutdown -c`) or create a temp file with a shell script similar to this echo and bind-mount it over `/usr/sbin/shutdown`.

dnf-behave-tests/dnf/dnf-automatic/reboot.feature
```diff
@@ -0,0 +1,101 @@
...
+
+    [emitters]
+    emit_via = stdio
+    """
+    And I use repository "simple-base"
+    And I successfully execute dnf with args "install labirinto"
+    And I use repository "simple-updates"
```
inknos, line 26: What about adding this to `Background` and just use the same repo for every scenario? I believe it would make the code more readable and each scenario much shorter. For instance, if you use the repos `dnf-ci-fedora{,updates}` then you can play with packages that do not require reboots as well like `abcde`.

evan-goode: How's that, I still kept the test with the kernel upgrade the same but the rest are shorter.

inknos:
LGTM

PR #1212, opened 2023-01-30
Add BZ reference

inknos:
In the previous [review](https://github.com/rpm-software-management/ci-dnf-stack/pull/1210), I forgot to mention that the Bugzilla tag was missing.

PR #1213, opened 2023-01-31
Fix the syntax error

m-blaha:
Background does not support tags, we need to tag scenarios.


kontura:
This is for: https://github.com/rpm-software-management/ci-dnf-stack/pull/1212

PR #1214, opened 2023-01-31
Add python API test for RepoQuery

glum23:
add python API test scenarios for RepoQuery

### DIFF
--- a/dnf-behave-tests/dnf/api/repo-query.feature
+++ b/dnf-behave-tests/dnf/api/repo-query.feature
@@ -1,0 +1,91 @@
+@dnf5
+Feature: api: query repos
+
+
+Background:
+Given I use repository "simple-base"
+  And I use repository "dnf-ci-fedora"
+  And I use repository "dnf-ci-fedora-updates"
+
+
+Scenario: Construct query and filter dnf-ci-fedora repo
+ When I execute python libdnf5 api script with setup
+      """
+      query = libdnf5.repo.RepoQuery(base)
+      query.filter_id("dnf-ci-fedora")
+      for repo in query:
+          print(repo.get_id())
+      """
+ Then the exit code is 0
+  And stdout is
+      """
+      dnf-ci-fedora
+      """
+
+
+Scenario: Construct query and filter fails due to bad argument type
+ When I execute python libdnf5 api script with setup
+      """
+      query = libdnf5.repo.RepoQuery(base)
+      query.filter_id(123)
+      for repo in query:
+          print(repo.get_id())
+      """
+ Then the exit code is 1
+  And stdout is empty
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/api/repo-query.feature
```diff
@@ -0,0 +1,95 @@
...
+Given I use repository "simple-base"
+  And I use repository "dnf-ci-fedora"
+  And I use repository "dnf-ci-fedora-updates"
+ When I execute python libdnf5 api script with setup
+      """
+      from libdnf5.common import QueryCmp_ICONTAINS
+      query = libdnf5.repo.RepoQuery(base)
+      query.filter_id("CI", QueryCmp_ICONTAINS)
+      for repo in query:
+          print(repo.get_id())
```
kontura, line 47: The tests are failing here because the output repo names are in a different order.
I think it is because the `libdnf5.repo.RepoQuery` uses a `set` as an internal data storage which does not guarantee order.
We could fix this for example by ordering the output in the script, something like:
```python
reposlist = sorted(list(query))
for repo in reposlist:
    print(repo.get_id())
```

Another option could be to use a different step: 
```gherkin
    And stdout contains lines               
      """
      simple-base
      dnf-ci-fedora
      """
```
Which can handle out of order lines, but it is a weaker check, it will still pass even if the output contains additional lines.

dnf-behave-tests/dnf/api/repo-query.feature
```diff
@@ -0,0 +1,95 @@
...
+      """
+      simple-base
+      dnf-ci-fedora-updates
+      """
+
+
+Scenario: Construct query and filter using QueryCmp_NOT_IGLOB
+Given I use repository "simple-base"
+  And I use repository "dnf-ci-fedora"
+  And I use repository "dnf-ci-fedora-updates"
```
kontura, line 80: Since this section is (can be) used by all of the `Scenarios` you could extract it into a `Background` at the top of the file to de-duplicate it.

Something like in https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/installonly.feature but with all three repos.

kontura (changes requested):

dnf-behave-tests/dnf/api/repo-query.feature
glum23: OK, sorry, the test passed for me even without sorting. I've sorted the list of ids instead of sorting the query objects.

dnf-behave-tests/dnf/api/repo-query.feature
glum23: OK, done. Thanks

kontura:
Great, thanks!

kontura (approved):

PR #1215, opened 2023-01-31
Test config-manager --dump main

pkratoch:
Requires: https://github.com/rpm-software-management/dnf-plugins-core/pull/480

### DIFF
--- a/dnf-behave-tests/dnf/plugins-core/config-manager.feature
+++ b/dnf-behave-tests/dnf/plugins-core/config-manager.feature
@@ -337,3 +337,12 @@ Scenario: config-manager --save preserves comments and empty lines
 
         # trailing comment
         """
+
+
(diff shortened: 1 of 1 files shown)


jan-kolarik:
This should be probably moved to the `dnf-4-stack` branch ...

pkratoch:
Thanks, I forgot we branched. I moved it there :)

jan-kolarik:
Hmm I don't know if it is because of the "move" or because it's the first commit after changing the GitHub workflow for `dnf-4-stack` branch, but it is only triggering the `dnf5` tests instead of just `dnf` one...

pkratoch:
I will try to open the PR again.

pkratoch:
New PR: https://github.com/rpm-software-management/ci-dnf-stack/pull/1225

PR #1216, opened 2023-02-03
Add tests for transaction.test()

evan-goode:
Requires https://github.com/rpm-software-management/dnf5/pull/251

@glum23 I added a `test_transaction` function in `dnf-behave-tests/dnf/steps/cmd.py` instead of adding on to `execute_transaction`, since `transaction.run()` already does the same checks that `transaction.test()` does.

### DIFF
--- a/dnf-behave-tests/dnf/api/transaction-test.feature
+++ b/dnf-behave-tests/dnf/api/transaction-test.feature
@@ -1,0 +1,30 @@
+@dnf5
+Feature: transaction: dry-run a transaction
+
+
+@bz2109660
+Scenario: Test labirinto install transaction when it should succeed
+Given I use repository "simple-base"
+ When I execute python libdnf5 api script with setup
+      """
+      goal = libdnf5.base.Goal(base)
+      goal.add_rpm_install("labirinto")
+      assert test_transaction(goal) == libdnf5.base.Transaction.TransactionRunResult_SUCCESS
+      """
+ Then the exit code is 0
+  And stderr is empty
+
+@bz2109660
+Scenario: Test labirinto install transaction when it should fail
+Given I use repository "simple-base"
+ When I execute python libdnf5 api script with setup
+      """
+      goal = libdnf5.base.Goal(base)
+      goal.add_rpm_install("labirinto")
+      execute_transaction(goal, "install a package")
+
+      # We have already run the transaction, so test_transaction should return a failing status
+      assert test_transaction(goal) == libdnf5.base.Transaction.TransactionRunResult_ERROR_RPM_RUN
+      """
+ Then the exit code is 0
+  And stderr is empty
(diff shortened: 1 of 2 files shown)


glum23:
@evan-goode thanks, LGTM, just @not.with_dnf=4 should not be included into purely dnf5 tests (I know it is present in some other tests, however, it was "inherited" there from the original master branch). Also, I'm not sure whether the empty line 471 in [dnf-behave-tests/dnf/steps/cmd.py](https://github.com/rpm-software-management/ci-dnf-stack/pull/1216/files#diff-bf849980d27ef02f2a2a955b0100345f08bd6b721f4f2eb332c38d6b33003e7e) is in accordance with ci-dnf-stack's coding style, @kontura would know. I'd say it'd be useful to run the test before merging this PR so I need https://github.com/rpm-software-management/dnf5/pull/251 to be merged first.

kontura:
I believe we don't have a set coding style regarding the empty line before `return`. Looking at other python functions in `ci-dnf-stack` `steps` some have it some don't. I think it is fine either way.

evan-goode:
Took out the blank line and the @not.with_dnf=4.

evan-goode:
#251 has been merged so this PR should be ready to go.

dnf-behave-tests/dnf/api/transaction-test.feature
```diff
@@ -0,0 +1,30 @@
...
+      """
+      goal = libdnf5.base.Goal(base)
+      goal.add_rpm_install("labirinto")
+      execute_transaction(goal, "install a package")
+
+      # We have already run the transaction, so test_transaction should return a failing status
+      assert test_transaction(goal) == libdnf5.base.Transaction.TransactionRunResult_ERROR_RPM_RUN
+      """
+ Then the exit code is 0
+  And stderr does not contain "Traceback"
```
kontura, line 30: We would like to have the tests as strict as possible therefore I think it would be better to use
```gherkin
And stderr is empty
```
The stderr really should be empty, no problems are expected here right?
(for both tests)

kontura (changes requested):

dnf-behave-tests/dnf/api/transaction-test.feature
glum23: I've just run the test on f38 with dnf5-5.0.6-1.fc39, the first scenario passed as it is and also with "stderr is empty". The second scenario failed due to `AttributeError: 'Transaction' object has no attribute 'set_description'`, the build probably doesn't contain related PR.

evan-goode: > The stderr really should be empty, no problems are expected here right?
Yes, thanks, done!

evan-goode: > I've just run the test on f38 with dnf5-5.0.6-1.fc39, the first scenario passed as it is and also with "stderr is empty". The second scenario failed due to `AttributeError: 'Transaction' object has no attribute 'set_description'`, the build probably doesn't contain related PR. When dnf5-5.0.6-0.20230220010121.5.0.5+54.g9a7b19b1.fc39 is used instead, both scenarios work OK, also with "stderr is empty".

Yes, see https://github.com/rpm-software-management/ci-dnf-stack/pull/1221

kontura:
Thank you.

PR #1218, opened 2023-02-08
createrepo_c: adjust to a new default of not generating dbs

kontura:
Requires: https://github.com/rpm-software-management/createrepo_c/pull/347


dralley (approved):
Looks like there's a chicken and egg problem

PR #1220, opened 2023-02-14
Add `CAP_SYS_CHROOT` capability when running DNF CI

kontura:
Since podman-4.4.0 `CAP_SYS_CHROOT` is required in order to use installroot (chroot) inside the containers.

https://bugzilla.redhat.com/show_bug.cgi?id=2167735


jrohel:
It looks OK. Fixes chroot permission issue.

PR #1222, opened 2023-02-16
Enable "list" command tests for dnf5

m-blaha:
There are small differences in dnf5 output that needed to be handled.

Requires https://github.com/rpm-software-management/dnf5/pull/298


kontura:
Thanks!

PR #1224, opened 2023-02-20
Update environment.py

zengwei2000:
fix typo


kontura:
We use [`behave`](https://behave.readthedocs.io/en/stable/) framework for the testing.
I don't think we should inflect the word. It is more clear to me in the original form.
I am against this change.

jan-kolarik:
Yes, I agree with @kontura, we should not change the `behave` occurrence in this case.

PR #1225, opened 2023-02-20
Test config-manager --dump main

pkratoch:
Requires: https://github.com/rpm-software-management/dnf-plugins-core/pull/480


jan-kolarik:
Thanks!

PR #1226, opened 2023-02-21
Switch CI to Fedora 38, it is the primary target for dnf5

kontura:


m-blaha:
Hm, there is a strange failure on 38 - pyftpdlib gets installed (with some deprecation warnings, though), but python still complains that `ModuleNotFoundError: No module named 'pyftpdlib'`

kontura:
I am looking into it.

kontura:
@m-blaha it seems adding `python3-pyftpdlib` dependency fixed it.
I found that in: https://bugzilla.redhat.com/show_bug.cgi?id=2160708 -> https://pagure.io/releng/issue/11242

But to be honest I don't really understand why it is needed since we install it through PIP.

There are also other issues which I am investigating but I think those are unrelated to the `pyftpdlib`.

m-blaha:
Hm, maybe behave somehow manipulates with `sys.path` and then python is not able to find modules in `/usr/local/lib/python3.11/site-packages/` where pip puts them. (The distribution rpm puts pyftpdlib to `/usr/lib/python3.11/site-packages/`).
But it's just my guess, I haven't investigated it thoroughly...

m-blaha:
And it really does.
I've created this simple `environment.py` file:
```
import sys
import pdb; pdb.set_trace()
import pyftpdlib
```

and run behave in that directory:

```
[root@992ed59c8251 test]# behave
> /test/environment.py(3)<module>()
-> import pyftpdlib
(Pdb) sys.path
['/test', '/usr/lib64/python311.zip', '/usr/lib64/python3.11', '/usr/lib64/python3.11/lib-dynload', '/usr/lib64/python3.11/site-packages', '/usr/lib/python3.11/site-packages']
(Pdb)
```

So no `/usr/local/lib/python3.11/site-packages` directory present.
If the program continued:

```
(Pdb) continue
Exception ModuleNotFoundError: No module named 'pyftpdlib'
  File "/usr/bin/behave", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3.11/site-packages/behave/__main__.py", line 183, in main
    return run_behave(config)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/behave/__main__.py", line 127, in run_behave
    failed = runner.run()
             ^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/behave/runner.py", line 804, in run
    return self.run_with_paths()
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/behave/runner.py", line 808, in run_with_paths
    self.load_hooks()
  File "/usr/lib/python3.11/site-packages/behave/runner.py", line 784, in load_hooks
    exec_file(hooks_path, self.hooks)
  File "/usr/lib/python3.11/site-packages/behave/runner_util.py", line 386, in exec_file
    exec(code, globals_, locals_)
  File "environment.py", line 3, in <module>
    import pyftpdlib
ModuleNotFoundError: No module named 'pyftpdlib'
```

m-blaha:
If I ran the very same test in F37 container, the `/usr/local/lib/python3.11/site-packages` directory is present in `sys.path` and everything passes. There has to be some change in python or behave between 37 and 38 causing this.

m-blaha:
I guess we can merge this change (as testing on f38 is really essential). But I'm afraid the issue can strike back in the future when testing on distributions that do not ship python3-pyftpdlib. @kontura or do you want to investigate it more?

kontura:
I would like to find the root cause.

But regardless I need to fix the CI first it is failing on issues unrelated to `python3-pyftpdlib`.

kontura:
The `pyftpdlib` problem was resolved by: https://bugzilla.redhat.com/show_bug.cgi?id=2179979
There is also a problem with new swig in f38, it should be resolved by: https://github.com/rpm-software-management/dnf5/pull/426

kontura:
@m-blaha This should be finally ready.

m-blaha:
Thank you, I'll take a look.

PR #1227, opened 2023-02-22
Enable install-non-existent.feature for dnf5

m-blaha:
Also one scenario needed to be changed because of change in the --skip-broken switch meaning. What the scenario actually tests is --skip-unavailable option.

Requires https://github.com/rpm-software-management/dnf5/pull/316


j-mracek:
LGTM

jan-kolarik:
Looks good, thanks!

PR #1233, opened 2023-03-14
Test for storing the comps metadata cache along with other metadata

jan-kolarik:
Requires https://github.com/rpm-software-management/dnf5/pull/346.


m-blaha (approved):
Than you.

PR #1234, opened 2023-03-14
Add test for installation according binary name

j-mracek:
This is a new functionality in DNF5 where it installs provider of binary using binary name without binary path


j-mracek:
@m-blaha May I ask you for a review of the test?

PR #1236, opened 2023-03-15
Adjust obsoletes test to new dnf5 behavior.

m-blaha:
Now install takes only obsoleters of the latest version into account.

Requires: https://github.com/rpm-software-management/dnf5/pull/354


j-mracek:
LGTM

PR #1237, opened 2023-03-16
Adjust tests problems output to stderr

m-blaha:
Now are resolve logs printed to stderr, need to adjust several steps. Also removed dnf4/dnf5 steps since this branch is supposed to be run only with dnf5.

Requires: https://github.com/rpm-software-management/dnf5/pull/365


jan-kolarik:
Thx!

PR #1239, opened 2023-03-20
Tests for keepcache option functionality

jan-kolarik:
Requires https://github.com/rpm-software-management/dnf5/pull/459.


j-mracek:
Perfect, I like that tests also have a negative control.

PR #1240, opened 2023-03-21
Enable tests for search command

jan-kolarik:
Modify existing tests related to `search` command to work with DNF5.

### DIFF
--- a/dnf-behave-tests/dnf/search-sort.feature
+++ b/dnf-behave-tests/dnf/search-sort.feature
@@ -1,9 +1,10 @@
-# @dnf5
-# TODO(nsella) implement command search
+@dnf5
 Feature: Sort search command output
+
 
 Background:
   Given I use repository "search-sort"
+
 
 @bz1811802
 Scenario: sort alphanumerically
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/dnf/search-sort.feature
```diff
@@ -9,7 +8,7 @@ Background:
 Scenario: sort alphanumerically
   When I execute dnf with args "search name"
   Then the exit code is 0
-   And stdout is
+   And dnf4 stdout is
```
m-blaha, line 11: Do we need this dnf4/dnf5 distinction in the main branch? I've already started to remove it from tests that I newly enable for dnf5. I thing dnf4 is not supposed to be tested by this branch.

dnf-behave-tests/dnf/search.feature
```diff
@@ -35,13 +35,17 @@ Scenario: with installed and availiable newest package doesn't duplicate results
...
+   <REPOSYNC>
+   Matched fields: name (exact), summary
+    setup.noarch: A set of system configuration and setup files
+    setup.src: A set of system configuration and setup files
    """
 
 
-Scenario: with installed and availiable newest package and --showduplicates duplicates results
+# TODO(jkolarik): Do we want this in DNF5? Current package query behavior seems different 
+# and here are two same NEVRAs...
```
m-blaha, line 46: The redundant NEVRA is for the installed packages (which technically is a different package, although the user cannot see the difference from the output).
I guess current dnf5 behavior is OK until we add a possibility for the user to have also repoid in the output (e.g. something like --queryformat, or some kind of structured output like json).

m-blaha (approved):

PR #1246, opened 2023-04-05
Enable tests for check-upgrade command

evan-goode:
The check-update subcommand is (probably?) being renamed to check-upgrade in DNF5.

Requires https://github.com/rpm-software-management/dnf5/pull/353.


dnf-behave-tests/dnf/security-security.feature
```diff
@@ -11,9 +11,8 @@ Background: Use repository with advisories
         | install       | security_B-0:1.0-1.x86_64 |
 
 
-# @dnf5
-# TODO(nsella) Unknown argument "check-update" for command "microdnf"
-Scenario: Security check-update when there are no such updates
+@dnf5
```
jan-kolarik, line 14: If I see it correctly, now we can set a global `@dnf5` tag for this whole feature file.

evan-goode: Done, thanks.

PR #1249, opened 2023-04-14
Enable tests with autoremove command for dnf5

m-blaha:
Requires: https://github.com/rpm-software-management/dnf5/pull/456


jan-kolarik (approved):
Thanks!

PR #1252, opened 2023-04-22
[dnf5] Download command: download latest version

evan-goode:
Test the download subcommand when more than one version of a package is available in the repositories.

Requires https://github.com/rpm-software-management/dnf5/pull/483

### DIFF
--- a/dnf-behave-tests/dnf/dnf5/download.feature
+++ b/dnf-behave-tests/dnf/dnf5/download.feature
@@ -58,3 +58,14 @@ Scenario: Download with alldeps options
    Then stderr contains "Option \"--alldeps\" should be used with \"--resolve\""
     And the exit code is 2
 
+@not.with_dnf=4
+@dnf5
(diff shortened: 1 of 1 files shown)


evan-goode:
Tests regressions of https://github.com/rpm-software-management/dnf5/issues/472 and [bz2187981](https://bugzilla.redhat.com/show_bug.cgi?id=2187981).

dnf-behave-tests/dnf/dnf5/download.feature
```diff
@@ -58,3 +58,14 @@ Scenario: Download with alldeps options
    Then stderr contains "Option \"--alldeps\" should be used with \"--resolve\""
     And the exit code is 2
 
+@not.with_dnf=4
```
jan-kolarik, line 61: Minor: Could you please add `@dnf5` tag for this whole feature and drop the `@not.with_dnf=4` tag as we are already branched?

PR #1258, opened 2023-04-24
Test undoing install transaction when another arch is installed (dnf-4-stack)

kontura:
For https://bugzilla.redhat.com/show_bug.cgi?id=2172292 
The fix is in libsolv.

### DIFF
--- a/dnf-behave-tests/dnf/install-remove-multilib.feature
+++ b/dnf-behave-tests/dnf/install-remove-multilib.feature
@@ -73,3 +73,14 @@ Scenario: Installing an older version of x86_64 pulls in a newer version of i686
       | install       | foo-1.0-1.x86_64               |
       | install-dep   | python3-foo-1.0-1.x86_64       |
       | install-dep   | python3-foo-clibs-1.0-1.x86_64 |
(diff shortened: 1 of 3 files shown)


m-blaha (approved):

dnf-behave-tests/dnf/install-remove-multilib.feature
```diff
@@ -73,3 +73,14 @@ Scenario: Installing an older version of x86_64 pulls in a newer version of i686
       | install       | foo-1.0-1.x86_64               |
       | install-dep   | python3-foo-1.0-1.x86_64       |
       | install-dep   | python3-foo-clibs-1.0-1.x86_64 |
+
+
+@bz2172292
+Scenario: undo install transaction when another arch is installed
+Given I successfully execute dnf with args "install perduto.i686"
+  And I successfully execute dnf with args "install perduto.x86_64"
+ When I execute dnf with args "history undo last"
```
j-mracek, line 82: Only note: What about to add a comment - the test requires the presence of available package for example and the operation is not supposed to do something?

PR #1265, opened 2023-05-02
Remove dnf5 search --showduplicates test covering the installed&available package

martinobanana:
Let's remove the test for search --showduplicates covering the installed&available package scenario.

I believe the behavior of dnf5 is more correct that it doesn't show the installed package when using --showduplicates. The list command of course shows both which cover the specific user scenario.

@jan-kolarik You've had TODO: there so please discuss


jan-kolarik:
Yep I agree. Unless we would modify the `search` behavior to include some additional information like the from repo_id, I don't see any benefit using the previous `dnf` behavior.

PR #1267, opened 2023-05-03
Switch CI to Fedora 38

jan-kolarik:
There are some changes in generating of SWIG bindings in Fedora 38 and we want to cover those in the CI.

Discovered in the https://github.com/rpm-software-management/dnf-plugins-core/pull/485. This PR is required to make the tests passing.


kontura:
Thanks!

PR #1275, opened 2023-05-09
Tests for transaction test file conflicts reporting

jan-kolarik:
Requires https://github.com/rpm-software-management/dnf5/pull/517.

### DIFF
--- a/dnf-behave-tests/dnf/install-file-conflicts.feature
+++ b/dnf-behave-tests/dnf/install-file-conflicts.feature
@@ -1,0 +1,16 @@
+@dnf5
+Feature: Tests for file conflicts reporting
+
+Background:
+  Given I use repository "dnf-ci-install-conflicts"
+
+
+Scenario: An error is reported when a package with a file conflict is tried to be installed
+  Given I successfully execute dnf with args "install package-one-1.0-1.x86_64"
+   Then Transaction is following
+        | Action                    | Package                      |
+        | install                   | package-one-0:1.0-1.x86_64   |
+   When I execute dnf with args "install package-two-1.0-1.x86_64"
+   Then the exit code is 0
+   Then stdout contains "Rpm transaction failed."
+    And stdout contains "file /usr/lib/package/conflicting-file from install of package-two-0:1.0-1.x86_64 conflicts with file from package package-one-0:1.0-1.x86_64"
--- a/dnf-behave-tests/fixtures/specs/dnf-ci-install-conflicts/package-one-1.0-1.spec
+++ b/dnf-behave-tests/fixtures/specs/dnf-ci-install-conflicts/package-one-1.0-1.spec
@@ -1,0 +1,22 @@
+Name:           package-one
+Epoch:          0
+Version:        1.0
+Release:        1
+
+License:        Public Domain
+URL:            None
+
+Summary:        The made up package to fail on a file conflict
+
+%description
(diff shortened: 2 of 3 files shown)


jan-kolarik:
I've got two concerns here:
- I was thinking if non-zero error code should be returned in this case
- Maybe we could also add a separate transaction run result like `ERROR_RPM_TEST_RUN`

jan-kolarik:
The exit code issue is addressed [here](https://github.com/rpm-software-management/dnf5/issues/525). I think we could merge this as it is now and the tests will be fixed later with the logic fixes in the mentioned issue.

kontura:
I would first prefer to have all the tests pass even though the createrepo_c tests are not related.
It should be fixed by: https://github.com/rpm-software-management/ci-dnf-stack/pull/1276

dnf-behave-tests/dnf/install-file-conflicts.feature
```diff
@@ -0,0 +1,17 @@
...
+  Given I use repository "dnf-ci-install-conflicts"
+
+
+Scenario: An error is reported when a package with file conflicts is tried to be installed
+   When I execute dnf with args "install FileConflict-1.0-1.x86_64"
+   Then the exit code is 0
+    And Transaction is following
+        | Action                    | Package                      |
+        | install                   | FileConflict-0:1.0-1.x86_64  |
+   When I execute dnf with args "install FileConflict-2.0-1.x86_64"
```
kontura, line 14: I find this quite confusing. It is not actually an `install` its doing an `upgrade` of pkg `FileConflict` from version 1 to 2.

I am also not sure why the conflict even happens? Is it because the `a_dir` changes from a file to a directory? Why is that a problem though? The packages are not installed at the same time.

For the test it would be much clearer to me if we had a `pkg-a` and `pkg-b` both containing the same file.

jan-kolarik: Yes, it could be simplified. I basically copied that from a very similar use case in the `module/install-errors.feature`. But I can create a simpler test case.

dnf-behave-tests/dnf/install-file-conflicts.feature
```diff
@@ -0,0 +1,17 @@
+@dnf5
+Feature: Tests for file conflicts reporting
+
+Background:
+  Given I use repository "dnf-ci-install-conflicts"
+
+
+Scenario: An error is reported when a package with file conflicts is tried to be installed
+   When I execute dnf with args "install FileConflict-1.0-1.x86_64"
```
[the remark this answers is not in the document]
jan-kolarik: Yep, I should start using the `I successfully execute` variant. I didn't know about that some time ago and I got used to this workflow...

dnf-behave-tests/dnf/install-file-conflicts.feature
kontura: I would prefer simplification.

Actually this is an RPM issue: https://docs.fedoraproject.org/en-US/packaging-guidelines/Directory_Replacement/  it would be better not to rely on it.

jan-kolarik: Oh, OK. Thanks for the context, then it definitely deserves rewriting.

kontura (approved):
Thanks!

PR #1277, opened 2023-05-16
New test for leaves plugin #rhbz2134638

martinobanana:
Just a simple test to verify that the leaves (and show-leaves) plugin works as expected. dnf5 behaves differently so this is just for dnf4.

Works fine on F37:
1 feature passed, 0 failed, 0 skipped
1 scenario passed, 0 failed, 0 skipped
23 steps passed, 0 failed, 0 skipped, 0 undefined
Took 0m1.354s

### DIFF
--- a/dnf-behave-tests/createrepo_c/mergerepo_c-modular.feature
+++ b/dnf-behave-tests/createrepo_c/mergerepo_c-modular.feature
@@ -97,8 +97,8 @@ Scenario: merged repository contains streams from both source repositories
       data:
         module: test-module
         profiles:
-          "test-profile1": [default]
-          "test-profile2": [default]
+          test-profile1: [default]
+          test-profile2: [default]
       ...
       ---
       document: modulemd
@@ -113,7 +113,7 @@ Scenario: merged repository contains streams from both source repositories
           Made up module
         license:
           module:
-          - "MIT"
+          - MIT
         profiles:
           test-profile1:
             rpms:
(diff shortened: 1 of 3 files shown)


kontura:
The failing createrepo_c tests are unrelated, the fix is here: https://github.com/rpm-software-management/ci-dnf-stack/pull/1279

dnf-behave-tests/dnf/plugins-core/leaves.feature
```diff
@@ -0,0 +1,44 @@
+Feature: Show packages not needed by any package
```
kontura, line 1: We usually have two empty lines after the `Feature` line, could you add those to be consistent?

dnf-behave-tests/dnf/plugins-core/leaves.feature
```diff
@@ -0,0 +1,44 @@
...
+    And Transaction is following
+        | Action        | Package                           |
+        | install       | abcde-0:2.9.2-1.fc29.noarch       |
+        | install-dep   | wget-0:1.19.5-5.fc29.x86_64       |
+        | install-weak  | flac-1.3.2-8.fc29.x86_64          |
+   Then stdout section "New leaves:" contains "abcde.noarch"
+   Then stdout section "New leaves:" contains "flac.x86_64"
+   When I execute dnf with args "leaves"
+   Then the exit code is 0
+   Then stdout matches line by line
```
kontura, line 17: We prefer to be as strict as possible so could you replace all the
```gherkin
Then stdout matches line by line
```
by
```gherkin
Then stdout is
```

I believe the output should be stable here.

PR #1278, opened 2023-05-16
Disable --allow-downgrade test for daemon

kontura:
This option is currently not supported by dnf5daemon-client.


jan-kolarik:
Thanks!

PR #1280, opened 2023-05-16
reinstall: added scenarios for --skip-unavailable

glum23:
Added scenarios for reinstall when at least one of the arguments is either not installed or does not exist, also for --skip-unavailable


j-mracek (approved):
LGTM

PR #1281, opened 2023-05-17
New tests for leaves and show-leaves plugins

martinobanana:
Just a simple test to verify that the leaves (and show-leaves) plugin works as expected. dnf5 behaves differently so this is just for dnf4.

2nd try after doing something wrong in https://github.com/rpm-software-management/ci-dnf-stack/pull/1277


kontura:
Thanks!

PR #1282, opened 2023-05-17
New tests for leaves and show-leaves plugins

martinobanana:
Just a simple test to verify that the leaves (and show-leaves) plugin works as expected.

Just a backport from dnf-4-stack https://github.com/rpm-software-management/ci-dnf-stack/pull/1281

This PR is waiting for the inclusion of pkgs to compose (otherwise the ci-dnf-stack will fail on building the image), please do not merge until it's ready for review.


glum23:
LGTM

PR #1283, opened 2023-05-17
dnf-nighly: keep libsolv fedora patches

kontura:
Since we take libsolv specfile from fedora we also need to keep patches if there are any. Otherwise the build will fail because the spec requires them.

There was a recent libsolv fedora release which contains several patches and it is currenly causing our nightly builds to fail.

Here is one failing action: https://github.com/rpm-software-management/ci-dnf-stack/actions/runs/4998121904/jobs/8957172363

### DIFF
--- a/overlays/dnf-nightly/overlay.yml
+++ b/overlays/dnf-nightly/overlay.yml
@@ -13,7 +13,7 @@ components:
       src: github:openSUSE/libsolv.git
     distgit:
       src: fedorapkgs:libsolv.git
-      patches: drop
+      patches: keep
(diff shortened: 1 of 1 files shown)


jan-kolarik (approved):
Nice, thanks for the fix!

jan-kolarik:
The same fix is probably also needed for the dnf-4 branch?

kontura:
> The same fix is probably also needed for the dnf-4 branch?

I would say that is not necessary, the github workflow runs only on the default branch (main) and I think we don't want to do the same build multiple times.

PR #1284, opened 2023-05-17
upgrade/downgrade/distro-sync: added scenarios for lists of packages, --skip-unavailable

glum23:
Added scenarios for lists of packages, some of them not installed, some of them not available, both with and without --skip-unavailable.

Depends on https://github.com/rpm-software-management/dnf5/pull/532


glum23:
Other commands (downgrade, distrosync) need scenarios like these.

kontura (approved):
Thank you!

PR #1286, opened 2023-05-22
Remove microdnf from install obsolete

j-mracek:
In Fedora 38, microdnf is obsoleted by dnf5 therefore we cannot install both - microdnf and dnf5.


kontura:
The problem was already resolved by: https://github.com/rpm-software-management/librepo/pull/278
Therefore I am closing this.

PR #1288, opened 2023-05-22
Update solver errors according to new formatting in dnf

inknos:
The new format contains information about repoid.


dnf-behave-tests/dnf/broken-dependencies-report.feature
```diff
@@ -0,0 +1,81 @@
+@dnf5
```
jrohel, line 1: Remove this file from the PR. It adds new tests for feature not present in RHEL 8.9.

inknos: done

PR #1291, opened 2023-05-23
Fix podman runs: temporarily hardcode crun downgrade

kontura:
There appears to be a problem with the newest crun. Reported issue: https://github.com/containers/crun/issues/1226

This should make the CI work again.


jan-kolarik:
Thanks for looking into this!

PR #1079, opened 2022-03-28
Add test for "swap" command in old microdnf

jrohel:
Requires:
https://github.com/rpm-software-management/microdnf/pull/121
https://github.com/rpm-software-management/libdnf/pull/1478

PR #1294, opened 2023-05-25
Add test for "swap" command in old microdnf

jrohel:
Requires:
https://github.com/rpm-software-management/microdnf/pull/121
https://github.com/rpm-software-management/libdnf/pull/1478


pkratoch:
In dnf, it's apparently also possible to swap groups and to swap packages using rpm file path (at least there are tests for that: https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/swap.feature). Is it possible also for microdnf?

Another use case that comes to my mind is swapping two alternate providers of a provide without removing a package that depends on this provide. It would be nice to test it as well, but we don't even have such a test for dnf.

jrohel:
@pkratoch 
These are more than a year old tests for old microdnf. They are testing a more than a year old PR that was merged a few days ago. I made a new PR because the original https://github.com/rpm-software-management/ci-dnf-stack/pull/1079 is against the "main" branch. And the "main" branch no longer tests the old microdnf.

The old microdnf does not support working with groups and provides. And I mean in general. Not just the "swap" command. But neither do the other commands - install, remove, ...

kontura:
@jrohel Can you please rebase the branch? There is a chance it will fix the CI.

jrohel:
@kontura Ok. I rebased the PR.

kontura:
Ok github is showing me dnf5 tests, I have no idea why, but it also run correct dnf4 action https://github.com/rpm-software-management/ci-dnf-stack/actions/runs/5079812697

I expect that will fail when trying to run the container, I am working on a fix for that.

PR #1304, opened 2023-06-01
Enable another set of groups/environments tests for dnf5

m-blaha:
Requires: https://github.com/rpm-software-management/dnf5/pull/585


kontura (approved):
Thank you!

PR #1307, opened 2023-06-05
Test coverage for --downloadonly option for install/upgrade commands

bajertom:
Test coverage for --downloadonly option for install/reinstall/upgrade/downgrade/group install commands

### DIFF
--- a/dnf-behave-tests/dnf/downloadonly.feature
+++ b/dnf-behave-tests/dnf/downloadonly.feature
@@ -1,3 +1,4 @@
+@dnf5
 Feature: Only download rpms with --downloadonly and store them in cache
 
 Scenario: Install/reinstall/upgrade work correctly with --downloadonly option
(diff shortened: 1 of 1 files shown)


glum23:
@bajertom, as you can see in the logs, tests end up with traceback after a call of the step "Transaction is empty". It would be better to exclude these steps. Another question is whether the implementation of the step is OK, what do you think @jan-kolarik?  The transaction is not supposed to be empty, however,  this code (called instead of "Transaction is empty") will successfully pass (no action and no package listed):  
~~~
   And RPMDB Transaction is following
~~~

jan-kolarik:
Even the `And RPMDB Transaction is empty` is passing in this case. I think the difference here is the DNF transaction which is present even if `--downloadonly` and the RPMDB transaction which happens only when doing changes in the system.

jan-kolarik (approved):
LGTM, thanks!

PR #1311, opened 2023-06-08
createrepo_c tools: switch default compression to zstd 

kontura:
This PR also contains: https://github.com/rpm-software-management/ci-dnf-stack/pull/1218 which should be merged first.
I just wanted to avoid too many conflicts as both PRs are changing a lot of the same lines.

For: https://github.com/rpm-software-management/createrepo_c/pull/371


kontura:
I think the dnf5 tests are using createrepo_c from nightly so it will not have the zstd default change -> it will fail in the `clean-cachefiles.feature`

I have manually triggered the nightly build. Once it rebuilds createrepo_c it should work.

jan-kolarik:
Similar thing should be probably also introduced for the dnf4 branch.

kontura:
> Similar thing should be probably also introduced for the dnf4 branch.

I have made: https://github.com/rpm-software-management/ci-dnf-stack/pull/1319

PR #1314, opened 2023-06-13
gpg tests - adjust to new outputs, enable more tests

m-blaha:
- GPG verification tests were made more strict by checking the full output.
- Some gpg tests were enabled for dnf5daemon.
- Also some adujstment to the new, different wording was needed.

Requires: https://github.com/rpm-software-management/dnf5/pull/607


jan-kolarik:
Thanks!

PR #1315, opened 2023-06-13
API: add more scenarios for PackageQuery

glum23:
dnf5 python API: extend current scenarios and add more scenarios for PackageQuery

### DIFF
--- a/dnf-behave-tests/dnf/api/package-query.feature
+++ b/dnf-behave-tests/dnf/api/package-query.feature
@@ -3,12 +3,21 @@
 Feature: api: query packages
 
 
-Scenario: Construct query and filter labirinto package
+Scenario: Construct queries and filter packages by name
 Given I use repository "simple-base"
  When I execute python libdnf5 api script with setup
       """
+      from libdnf5.common import QueryCmp_CONTAINS, QueryCmp_GLOB
       query = libdnf5.rpm.PackageQuery(base)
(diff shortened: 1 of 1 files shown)


kontura:
> @amatej, filtering works cryptically from my POW. When a package is installed (and available at the same time), two occurrences are usually found (the same nevra). E.g. when labirinto is installed in the scenario "Construct queries and filter packages by description", it is duplicated in the output. OK, let's say it is expected

Indeed it is expected. It is because in https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/steps/cmd.py#LL492C1-L492C46 we pass `true` which sets loading of the system repo (that contains installed packages)

> However, it does not work for the watermelon package in "Construct queries and filter packages by files". When it is not installed, nothing is filtered despite the fact that there are 3 different versions in the repo plugins-callbacks and all of them contain the filtered glob. Is that behavior expected? Or should I file an issue?

This is also expected because you are filtering by file and this file is not in primary it is only in filelists however dnf5 doesn't load filelists by default. It can be configured by adding:
```python
config.get_optional_metadata_types_option().set(libdnf5.conf.METADATA_TYPE_FILELISTS)
```
(It has to be set before repositories are loaded.)
Note: Some files are also in primary, typically its files in `/etc/...` or if their path contains `bin/`.

kontura (approved):

PR #1317, opened 2023-06-14
Add dnf5-unstable copr repo to dnf4 CI copr builds

kontura:
There are unreleased changes in dnf5-unstable that are not in f38. The old build is breaking microdnf builds because libdnf5 in f38 still provides libdnf (not libdnf5).


jan-kolarik (approved):
Thanks!

PR #1322, opened 2023-06-20
Tests for changelog --since

jan-kolarik:
Requires https://github.com/rpm-software-management/dnf5/pull/629.

### DIFF
--- a/dnf-behave-tests/dnf/plugins-core/changelog.feature
+++ b/dnf-behave-tests/dnf/plugins-core/changelog.feature
@@ -1,0 +1,32 @@
+@dnf5
+Feature: dnf changelog command
+
+
+Background: Enable changelog plugin
+  Given I enable plugin "changelog"
+    And I use repository "dnf-ci-changelog"
+
+
+Scenario: Listing changelogs since given date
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/fixtures/specs/dnf-ci-changelog/pkg-with-changelogs-1.0-1.spec
```diff
@@ -0,0 +1,20 @@
...
+A testing package.
+
+%files
+
+%changelog
+* Mon May 01 2023 DNF5 Team <pkgs@dnf.team> - 1.0-2
+- New downstream release 1.0-2
+
+* Fri Jan 20 2023 DNF5 Team <pkgs@dnf.team> - 1.0-1
+- New downstream release 1.0-1
```
pkratoch, line 20: Can you either add an older changelog entry, or test with a newer date, so that the output differs from outputting the whole changelog?

jan-kolarik: Of course, that's a good idea.

PR #1332, opened 2023-06-27
Remove scenarios for unsupported autoremove-n command

glum23:
Scenarios for commands unsupported by dnf5 should be removed in this branch


inknos:
lgtm

PR #1346, opened 2023-07-25
fileprovides: create fileprovides with already present solv files

kontura:
For: https://github.com/rpm-software-management/dnf5/issues/520
Requires: https://github.com/rpm-software-management/dnf5/pull/777

### DIFF
--- a/dnf-behave-tests/fixtures/specs/fileprovides/agua-1.0-1.fc29.spec
+++ b/dnf-behave-tests/fixtures/specs/fileprovides/agua-1.0-1.fc29.spec
@@ -1,0 +1,19 @@
+Name: agua
+Version: 1.0
+Release: 1.fc29
+Summary: Made up package
+
+License: GPLv3+
+Url: None
+
+%description
+agua description
+
+%install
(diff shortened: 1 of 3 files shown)


dnf-behave-tests/fixtures/specs/fileprovides/agua-1.0-1.fc29.spec
```diff
@@ -0,0 +1,19 @@
+Name: agua
+Version: 1.0
+Release: 1.fc29
+Summary: Made up package
+
+License: GPLv3+
+Url: None
+
+%description
+planta description
```
evan-goode, line 10: Nit: should be "agua description"

evan-goode (changes requested):

dnf-behave-tests/fixtures/specs/fileprovides/agua-1.0-1.fc29.spec
kontura: Fixed. :+1:

evan-goode (approved):

PR #1347, opened 2023-07-26
Modify and enable tests for repo substitutions

j-mracek:
Requires - patch


dnf-behave-tests/dnf/vars.feature
```diff
@@ -45,7 +43,7 @@ Scenario: Variables arch supports basearch `loongarch64` {}
       """
 
 
-# @dnf5
+@dnf5
 # TODO(nsella) different stdout
```
evan-goode, line 47: This TODO can be removed, right?

evan-goode (changes requested):

dnf-behave-tests/dnf/vars.feature
j-mracek: Yes, I am going to improve it

j-mracek: Fixed

evan-goode (approved):

PR #1352, opened 2023-08-01
Remove stderr test step from install-xml-base.feature

m-blaha:
The warning on stderr is not important for the test result and there is
a difference between dnf5 and dnf5daemon-client behavior - dnfdaemon
does not print the warning on stderr.
Without this step we can keep the same scenarios for both dnf5 and
dnf5daemon-client.


evan-goode (approved):
Thanks, perfect. I forgot to run the tests with the dnf5daemon tag.

PR #1357, opened 2023-08-03
Update scenarios in upgrade.feature for dnf5

glum23:
Update remaining dnf4 scenarios in upgrade.feature for dnf5. 
Depends on https://github.com/rpm-software-management/ci-dnf-stack/pull/1355

### DIFF
--- a/dnf-behave-tests/dnf/upgrade.feature
+++ b/dnf-behave-tests/dnf/upgrade.feature
@@ -117,8 +117,7 @@ Scenario: Upgrade list of packages with --skip-unavailable, one of them is not i
         | upgrade       | flac-0:1.3.3-3.fc29.x86_64                |
 
 
-# @dnf5
-# TODO(nsella) rpmdb check fail
+@dnf5
 @tier1
 @bz1670776 @bz1671683
 Scenario: Upgrade all RPMs from multiple repositories with best=False
@@ -128,12 +127,14 @@ Scenario: Upgrade all RPMs from multiple repositories with best=False
   Given I configure dnf with
         | key  | value |
         | best | False |
-   When I execute dnf with args "upgrade"
+   When I execute dnf with args "--nogpgcheck upgrade"
    Then the exit code is 0
     And stderr is
     """
     Problem: cannot install the best update candidate for package SuperRipper-1.0-1.x86_64
       - nothing provides unsatisfiable needed by SuperRipper-1.3-1.x86_64
+
+    Warning: skipped PGP checks for 7 package(s).
     """
     And Transaction is following
         | Action        | Package                                   |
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/upgrade.feature
```diff
@@ -216,16 +224,20 @@ Examples:
...
+@dnf5
+# dnf4 @bz1585138
+# dnf5: we intentionally do not offer unrequested info
+Scenario Outline: Do nothing when updateinfo is available but other than requested <type> update is available
   Given I use repository "dnf-ci-fedora-updates"
     And I use repository "dnf-ci-thirdparty-updates"
    When I execute dnf with args "update <type>" 
    Then the exit code is 0
     And Transaction is empty
-    And stderr contains "No security updates needed, but 7 updates available"
```
kontura, line 228: I believe we actually want this: https://github.com/rpm-software-management/dnf5/issues/360
I would keep the test disabled.

dnf-behave-tests/dnf/upgrade.feature
```diff
@@ -207,7 +211,11 @@ Scenario Outline: Print correct number of available updates if update <type> is
    When I execute dnf with args "update <type>" 
    Then the exit code is 0
     And Transaction is empty
-    And stderr contains "No security updates needed, but 2 updates available"
```
kontura, line 210: I believe we actually want this: https://github.com/rpm-software-management/dnf5/issues/360
I would keep the test disabled.

dnf-behave-tests/dnf/upgrade.feature
```diff
@@ -187,17 +188,20 @@ Scenario: Print information about skipped packages
     And Transaction is following
         | Action        | Package                                   |
         | upgrade       | SuperRipper-0:1.2-1.x86_64                |
-        | broken        | SuperRipper-0:1.3-1.x86_64                |
    Then stdout section "Upgrading:" contains "SuperRipper\s+x86_64\s+1.2-1\s+dnf-ci-thirdparty-updates"
```
kontura, line 191: Since we have the step above "Transaction is following" this feels redundant to me.
We already check that `SuperRipper-0:1.2-1.x86_64` gets upgraded.
Is there some reason to keep it? 
It would also unblock the PR.

dnf-behave-tests/dnf/upgrade.feature
```diff
@@ -174,11 +176,10 @@ Scenario: Upgrade all RPMs from multiple repositories with best=True
         | upgrade       | wget-1:1.19.5-5.fc29.x86_64               |
         | upgrade       | SuperRipper-0:1.2-1.x86_64                |
         | upgrade       | abcde-0:2.9.3-1.fc29.noarch               |
-        | broken        | SuperRipper-0:1.3-1.x86_64                |
```
kontura, line 177: To match the test before the upgrade with best was added can you also add a check for `stderr` at the end?
To make sure dnf5 still reports the problem with best candidate when running with `--no-best`.

kontura (changes requested):

dnf-behave-tests/dnf/upgrade.feature
glum23: OK, I've discussed this personally with Jarda and he said it's not necessary to provide unrequested info, especially a number that says almost nothing (just there can be complaints that it is incorrect). I'd say - if necessary - just info that "other types of updates are available" will do. However, the scenario has been disabled as requested.

dnf-behave-tests/dnf/upgrade.feature
glum23: the same as previous issue

dnf-behave-tests/dnf/upgrade.feature
glum23: It was present in the original version of the scenario, I supposed it tests that the output is exactly as it is described in the scenario. If not, removing.

dnf-behave-tests/dnf/upgrade.feature
glum23: Done.

glum23:
Rebased and done requested changes. Passed on f-38 with the latest unstable dnf5.

kontura:
Thank you!

PR #1362, opened 2023-08-10
reposync: repos contain only compressed group.xml.zst

kontura:
createrepo_c no longer puts two comps.xml variants in repodata, there is just one compressed group.xml.zst with mdtype "group".

We use step: "the text file contents of.." because it uncompresses the inputs to compare them instead of direct binary compare with `diff`.


m-blaha (approved):
Thanks for fixing it!

PR #1364, opened 2023-08-15
Test printing nested exceptions in error messages

jan-kolarik:
Requires https://github.com/rpm-software-management/dnf5/pull/819.


evan-goode:
Thanks! I was looking at getting the tests in `repo-sync.feature` enabled now that we have `makecache`. But we should just do this here for now.

evan-goode (approved):

PR #1376, opened 2023-09-04
librpm is initialized even with --forcearch and --releasever

m-blaha:
Requires: https://github.com/rpm-software-management/dnf5/pull/856
For: https://github.com/rpm-software-management/dnf5/issues/855


jan-kolarik (approved):
Thanks!

PR #1377, opened 2023-09-04
Removed obsoleted comment

glum23:
Removed a comment that seems to be obsoleted. The scenario passes and the exit code is the same as in RHEL.


jan-kolarik:
Thanks!

PR #1379, opened 2023-09-07
Add comment for dnf5 transition

glum23:
upgrade-cached test currently fails for dnf5 due to localization. It has to be adjusted when localization is resolved for dnf5, related comment has been added.


j-mracek:
LGTM

PR #1381, opened 2023-09-15
keepcache: Add test for remove command scenario (RhBug:2237883)

jan-kolarik:
We don't want to remove package cache when just removing packages.

Requires: https://github.com/rpm-software-management/dnf5/pull/885.

### DIFF
--- a/dnf-behave-tests/dnf/keepcache.feature
+++ b/dnf-behave-tests/dnf/keepcache.feature
@@ -79,3 +79,30 @@ Scenario: Reseting keepcache does not remove previously kept packages from cache
    \./simple-base-[0-9a-f]{16}/packages/labirinto-1\.0-1\.fc29\.x86_64\.rpm
    \./simple-base-[0-9a-f]{16}/packages/vagare-1\.0-1\.fc29\.x86_64\.rpm
    """
+
+
+# Test that a package that was cached in the previous transaction (the 'vagare' package)
+# will be left untouched by the following successful remove operation (the 'labirinto' package)
+@bz2237883
+Scenario: Cached packages are not removed after remove command with keepcache set to false
+  Given I successfully execute dnf with args "install labirinto"
+   Then Transaction is following
+        | Action        | Package                                   |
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/keepcache.feature
```diff
@@ -79,3 +79,28 @@ Scenario: Reseting keepcache does not remove previously kept packages from cache
...
+
+
+@bz2237883
+Scenario: Cached packages are not removed after remove command with keepcache set to false
+  Given I use repository "scriptlets"
+    And I successfully execute dnf with args "install labirinto"
+   Then Transaction is following
+        | Action        | Package                                   |
+        | install       | labirinto-0:1.0-1.fc29.x86_64             |
+   When I execute dnf with args "install Package-pre-fail"
```
m-blaha, line 91: I might be missing something, but I think two install commands are over-complicating the test. I also do not like abusing `scriptlets` repository, I consider using --downloadonly switch clearer than relying on the transaction to fail due to a broken package. Would something like
```
Given I successfully execute dnf with args "install --downloadonly labirinto"
```
, then completely omitting "scriptlets" repository and Package-pre-fail installation, and then testing presence of `labirinto.rpm` in the cache work?

jan-kolarik: Hmm, you are right it is not intuitive. The labirinto package is not important here, I just want to be sure there is some package present which I use later for successful `remove` command usage. The point here is, that when packages are downloaded during **non-successful** transaction, they are kept after successfully running some `remove` command later. Which should match the use case from connected bz...

jan-kolarik: If it makes sense, I can try to add some comments there.

m-blaha: I think that having a failing install transaction is not important here (at least scenario name does not imply it). Important is to fill the cache with something and for this purpose will `install --downloadonly` serve well.

But I didn't realize you need to successfully remove something, so you need to also successfully install this something before. I mean - double install is needed, sorry for the confusion.

PR #1383, opened 2023-09-19
repoinfo: Add scenario with --add-values option

jan-kolarik:
Requires: https://github.com/rpm-software-management/dnf5/pull/896


jan-kolarik:
Closing this as the parent was also closed.

PR #1387, opened 2023-09-20
rpmlint: fix locale error

inknos:


.github/workflows/ci-host.yml
```diff
@@ -40,7 +40,7 @@ jobs:
            https://raw.githubusercontent.com/rpm-software-management/dnf5/main/dnf5.spec
           RUN dnf -y install ccache
           # needed for pre-commit hooks
-          RUN dnf -y install git python3-pip rpmlint
+          RUN dnf -y install git python3-pip rpmlint glibc-langpack-en
```
ppisar, line 43: Wouldn't be simpler to reset the locale at the beginning of the container?

~~~~
          docker build -t ${IMAGE}:${DATE} -t ${IMAGE}:latest -<<EOF
          FROM fedora:latest
+       ENV LC_ALL C.UTF-8
          RUN dnf -y update
~~~

You already do a similar thing in ./Dockerfile.

I looked where the locale is set in the only place is in the behave tests and that is already handled in dnf-behave-tests/requirements.spec. So I guess what we observe is a locale inherited from CI infrastructure. Or rpmlint forcefully sets a locale to en_US, and that it's a bug in rpmlint and rpmlint should depend on  glibc-langpack-en.

ppisar: It looks like a bug in rpmlint:

~~~~
$ LC_ALL=C.UTF-8 rpmlint dnf5.spec
[...]
dnf5.spec: E: specfile-error LANGUAGE = "en_US",
dnf5.spec: E: specfile-error LC_ALL = "en_US.UTF-8",
dnf5.spec: E: specfile-error LANG = "en_US.UTF-8"
[...]
$ rpm -qf /usr/bin/rpmlint
rpmlint-2.4.0-11.fc40.noarch
$ rpm -q --requires rpmlint | grep glibc-langpack-en
$ 
~~~~

ppisar: I opened a bug for rpmlint <https://bugzilla.redhat.com/show_bug.cgi?id=2239860>.

inknos: ok so let's wait for it

inknos: Agree, it's done

PR #1388, opened 2023-09-21
Adjust obsoletes test to new dnf behavior.

m-blaha:
Now install takes only obsoleters of the latest version into account.

Requires: https://github.com/rpm-software-management/libdnf/pull/1624
Requires: https://github.com/rpm-software-management/dnf/pull/1994


j-mracek (approved):
LGTM

PR #1399, opened 2023-10-30
Adjust dnf-automatic tests for dnf5

m-blaha:
- as automatic is a plugin in dnf5, it correctly handles installroot so no need for @no_installroot tag
- changed the way how automatic is called 'dnf-automatic' -> 'dnf5 automatic'

Requires: https://github.com/rpm-software-management/dnf5/pull/984

This PR only adjusts currently existing tests. We should cover more of the automatic functionality in follow-up pull requests.

### DIFF
--- a/dnf-behave-tests/dnf/dnf-automatic/reboot.feature
+++ b/dnf-behave-tests/dnf/dnf-automatic/reboot.feature
@@ -1,21 +1,16 @@
-@no_installroot
+@dnf5
 Feature: dnf-automatic reboots
 
-
 Background:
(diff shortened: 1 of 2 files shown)


jan-kolarik:
Seems like the dnf-automatic tests are failing...

m-blaha:
Oh yeah, they use the old automatic.conf location. I'll fix them.

m-blaha:
I also noticed that automatic checks are quite slow due to a bug in network availability detection. Here is the fix: https://github.com/rpm-software-management/dnf5/pull/1298

m-blaha:
The automatic.conf location is now fixed.

PR #1401, opened 2023-10-31
Remove dnf5 only step from obsoletes.feature

m-blaha:


glum23:
LGTM, has been tested manually, should be OK

PR #1403, opened 2023-11-08
dnf5: Enable needs-restarting tests

evan-goode:
Update the needs-restarting tests for DNF 5 and add a test for recommending a reboot when a package has an associated reboot_suggested advisory.

Requires https://github.com/rpm-software-management/dnf5/pull/887

For https://github.com/rpm-software-management/dnf5/issues/743

### DIFF
--- a/dnf-behave-tests/dnf/plugins-core/needs-restarting.feature
+++ b/dnf-behave-tests/dnf/plugins-core/needs-restarting.feature
@@ -1,3 +1,4 @@
+@dnf5
 @bz1639468
 Feature: Reboot hint
 
@@ -5,8 +6,12 @@ Background:
     Given I enable plugin "needs_restarting"
       And I use repository "dnf-ci-fedora"
       And I move the clock backward to "before boot-up"
-      And I execute dnf with args "install lame kernel basesystem glibc wget"
-      And I move the clock forward to "2 hours"
+      And I execute dnf with args "install lame kernel basesystem glibc wget lz4"
+      And I move the clock forward to "the present"
+      # Boot time is measured down to the second, so sleep for one second so
+      # the time packages are installed is at least one second after container
+      # boot
+      And I sleep for "1" seconds
       And I use repository "dnf-ci-fedora-updates"
 
 @bz1913962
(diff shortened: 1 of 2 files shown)


jan-kolarik:
There is 1 failing test after rebasing to the current dnf5 codebase with needs-restarting changes included.

evan-goode:
Interesting, it passes locally. I just double-checked to make sure I didn't forget to stage something, and that's not it. I'm going to trigger it to run again just for fun.

@inknos is there a way to make the CI do as "clean" a run as possible? Is there anything special needed to do since a file in `dnf-behave-tests/fixtures/specs` was changed in this PR?

kontura:
> Interesting, it passes locally. I just double-checked to make sure I didn't forget to stage something, and that's not it. I'm going to trigger it to run again just for fun.
> 
> @inknos is there a way to make the CI do as "clean" a run as possible? Is there anything special needed to do since a file in `dnf-behave-tests/fixtures/specs` was changed in this PR?

In the  `dnf-behave-tests/fixtures/specs` there is a `build.sh` script that needs to be run if a spec was changed.
If you are running the tests locally directly with `behave` you should run the script directly (you can even use the `-f` switch to force it to regenerate all the test packages.

If you use the `./container-test` to build a container for the tests you might have to rebuild it because it runs the `build.sh` during the container build. However if the change was already present during the container build it should be fine.

evan-goode:
OK, the CI is happy at long last. The problem was that the `NO_FAKE_STAT` setting for `faketime` doesn't work correctly on GitHub Actions. `needs-restarting` reads the mtime of `/proc/1`, and the test doesn't work correctly if faketime fakes that mtime.

The test worked fine when I ran the container on my machine, but not on GHA. It must have something to do with the kernel or the filesystem they use; maybe the filesystem stores mtimes in a way that is relative to the uptime rather than storing a full timestamp? That might take up less space and make sense for a CI environment. But I have no idea. I'd rather not spend the time definitively getting to the bottom of it since every debugging run I did took 50 minutes or so :)

To work around the issue, I had to add a 1-second sleep, instead of setting the time forward. It's not ideal, but I think it's a reasonable compromise of test complexity/development effort versus performance.

jan-kolarik (approved):

Issue #1400, opened 2023-10-30
Add "modules state is following" step alternative for DNF5


pkratoch:
This blocks many module tests that could be enabled now.

PR #1404, opened 2023-11-09
Enable and update some module tests

pkratoch:
Resolves: https://github.com/rpm-software-management/ci-dnf-stack/issues/1400

### DIFF
--- a/dnf-behave-tests/dnf/module/filtering.feature
+++ b/dnf-behave-tests/dnf/module/filtering.feature
@@ -1,3 +1,4 @@
+@dnf5
 Feature: Modular filtering must provide onlu relevant source packages
 
 Background:
@@ -14,6 +15,7 @@ Scenario: Check that only module packages including src are available
    Then the exit code is 0
     And stdout is
         """
+        <REPOSYNC>
         berry-0:1.0-1.wood.src
         berry-0:1.0-1.wood.x86_64
         """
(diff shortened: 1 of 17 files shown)


dnf-behave-tests/dnf/steps/module.py
```diff
@@ -46,19 +47,22 @@ def check_module_list(context):
 
 
 def get_modules_state(installroot):
-    cfgdir = os.path.join(installroot, 'etc/dnf/modules.d')
-    cfg = ConfigParser()
-    cfg.read(glob.glob(cfgdir + '/*.module'))
```
kontura, line 51: Since the `ConfigParser` and `glob` are no longer used should we also drop their imports?

dnf-behave-tests/dnf/module/filtering.feature
```diff
@@ -12,8 +13,9 @@ Scenario: Check that only module packages including src are available
         | berry-source | enabled   | wood       |           |
    When I execute dnf with args "repoquery berry"
    Then the exit code is 0
-    And stdout is
+    And stdout matches line by line
```
kontura, line 16: I am fine with the change, I am just curious why is it done?

pkratoch: Ah, for some reason I thought the "<REPOSYNC>" is checked only for "stdout matches line by line" and then forgot to check it and change it back. Thanks.

dnf-behave-tests/dnf/steps/module.py
pkratoch: Good catch! I removed them.

kontura (approved):
Great, thanks!

PR #1405, opened 2023-11-09
Onboard Tmt

inknos:
This PR onboards tmt to test DNF5 only.

The checks will run only when requested via `/packit tests` or `/packit tests rpm-software-management/dnf5#PRNUMBER`.

The PR also checks tmt plans with pre-commit tests.


Dockerfile
```diff
@@ -32,6 +32,12 @@ RUN set -x && \
     dnf -y upgrade; \
     dnf -y distro-sync --repo copr:copr.fedorainfracloud.org:rpmsoftwaremanagement:dnf5-unstable;
 
+RUN set -x && \
+    if [ -n "$COPR" ] && [ -n "$RPMS" ]; then \
```
m-blaha, line 36: Nitpick - for clarity I'd prefer to rename RPMS -> COPR_RPMS. At first glance it looks like independent variables and using logical and was not clear to me.

plans/main.fmf
```diff
@@ -0,0 +1,41 @@
...
+          BASE: "fedora:37"
+    - when: distro == fedora-38
+      environment:
+          BASE: "fedora:38"
+    - when: distro == fedora-39
+      environment:
+          BASE: "fedora:39"
+    - when: distro == fedora-rawhide
+      environment:
+          BASE: "fedora:rawhide"
```
m-blaha, line 14: This adjustments are a bit unfortunate - we will need to adjust them every time a new Fedora is branched. Would it be possible to use original distro value and do something like this?
```
...container-test build --base $(echo "$BASE" | tr '-' ':') ...
```

inknos: for fedora no problem but we'll need to adjust stuff for centos, rhel etc in the future

Dockerfile
inknos: sure, seems logical to me

plans/main.fmf
inknos: @m-blaha I tweaked the `--base` option but we will need to be smarter if we'll want to add Centos tests too

m-blaha: Thank you.

PR #1407, opened 2023-11-14
Remove dnf5-unstable repo and use dnf5 once it's available

inknos:
I would like to simplify the way how we build our ci-stack by cutting `dnf5-unstable` repo.

The repo `dnf5-unstable` does not serve any purpose since `DNF5` is built in `dnf-nightly`.

Furthermore, `dnf5-unstable` should be slowly decommissioned since the name unstable provides misleading information and the content is redundant with the nightly copr.

Also, I propose to use dnf5 once it becomes available.


jan-kolarik:
Nice, thanks.

PR #1423, opened 2023-12-11
Test makecache/mc alias

inknos:
https://github.com/rpm-software-management/dnf5/pull/1079


inknos:
packit still doesn't use the copr repo. waiting for dnf5 feature to be merged before retesting

PR #1424, opened 2023-12-11
Fix command to run createrepo_c suite

inknos:


inknos:
createrepo_c passed, the issue is fixed

PR #1427, opened 2023-12-13
Enable tests for module info that don't require install

pkratoch:
I considered installing modules using dnf4 and enabling the `info.feature`, but dnf5 then doesn't have information about how that changed the module states, since it's stored differently in dnf4 and dnf5, so it's not very usable.

Instead, I created a duplicit set of tests for module info without the need of module install. They can be removed once the install command is implemented and the `info.feature` enabled.

The order of module streams listed is different, because it's not sorted (at least yet), but the output of dnf4 also doesn't seem sorted.

Requires: https://github.com/rpm-software-management/dnf5/pull/1093
Requires: https://github.com/rpm-software-management/dnf5/pull/1094
Requires: https://github.com/rpm-software-management/dnf5/pull/1098

### DIFF
--- a/dnf-behave-tests/dnf/module/defaults.feature
+++ b/dnf-behave-tests/dnf/module/defaults.feature
@@ -26,6 +26,7 @@ Scenario: The default streams are identified in the output of module list
         | nodejs    | 11        | development, minimal, default |
 
 
+@dnf5
 @bz1618553
 Scenario: Default profiles are identified in the output of dnf info
    When I execute dnf with args "module info nodejs"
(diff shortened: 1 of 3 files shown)


jan-kolarik:
Seems like 2 tests are still expecting the output in different order.

pkratoch:
> Seems like 2 tests are still expecting the output in different order.

Yes sorry, I implemented the sorting now: https://github.com/rpm-software-management/dnf5/pull/1109

jan-kolarik (approved):

PR #1428, opened 2023-12-14
Test makecache as unprivileged user without system state

kontura:
Test makecache as unprivileged user without system state

For: https://github.com/rpm-software-management/dnf5/pull/1097


j-mracek:
LGTM

PR #1429, opened 2024-01-04
dnf5 config-manager: Sync w upstream - Add repoid to gener repo name, test warning messages

jrohel:
The new upstream code https://github.com/rpm-software-management/dnf5/pull/1128
- adds the repository id to the generated repository name
- writes warnings also to stderr (previously only to the log)

### DIFF
--- a/dnf-behave-tests/dnf/plugins-core/config-manager-opts.feature
+++ b/dnf-behave-tests/dnf/plugins-core/config-manager-opts.feature
@@ -130,7 +130,28 @@ Scenario: unset/remove an main option
       """
 
 
-Scenario: unset/remove an main option, trying to unset an not set option is OK
+Scenario: unset/remove unsupported main option is OK, but a warning is written
+  Given I create file "/etc/dnf/dnf.conf" with
+      """
+      [main]
+      best=True
+      unsupported=1
+      skip_unavailable=True
+      """
+   When I execute dnf with args "config-manager unsetopt best unsupported"
+   Then the exit code is 0
+    And file "/etc/dnf/dnf.conf" contents is 
+      """
+      [main]
+      skip_unavailable=True
+      """
+    And stderr is
+      """
+      config-manager: Request to remove unsupported main option: unsupported
+      """
+
+
+Scenario: unset/remove an main option, trying to unset an not set option is OK, but a warning is written
   Given I create file "/etc/dnf/dnf.conf" with
       """
       [main]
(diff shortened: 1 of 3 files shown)


dnf-behave-tests/dnf/plugins-core/config-manager-opts.feature
```diff
@@ -144,6 +165,20 @@ Scenario: unset/remove an main option, trying to unset an not set option is OK
...
+
+
+Scenario: trying to unset an not set main option (config file is not found) is OK, but a warning is written
+  Given I delete file "/etc/dnf/dnf.conf"
+   When I execute dnf with args "config-manager unsetopt best"
+   Then the exit code is 0
+    And stderr is
+      """
+      config-manager: Request to remove main option but config file not found: {context.dnf.installroot}/etc/dnf/dnf.conf
+      """
```
pkratoch, line 181: I am not sure what changed, but this scenario fails for me now. The problem is that dnf5 always fails with "Configuration file {config_path} not found" when the config file path is provided on the cli, which is also the case for the ci-dnf-stack tests.

jrohel: @pkratoch  It passed the tests locally on my PC. But you're right. If CI stack runs tests with config file path on cli it should fail. I have to look into it.

jrohel: @pkratoch 
I looked into it. Why does it pass in local CI? And I saw the command:
`dnf5 -y --installroot=/tmp/dnf_ci_installroot_q04aelti --releasever=29 --setopt=module_platform_id=platform:f29 --disableplugin='*' config-manager unsetopt best`

This means that the path to the configuration file is not provided on the cli. And that's why it goes right.

jrohel: OK. The problem is in the CI stack in the `get_dnf5_cmd` method.
Compared to `get_dnf4_cmd`, it adds:
```
result.append("--setopt=reposdir=%s" % self.installroot + "/etc/yum.repos.d")
result.append("--setopt=config_file_path=%s" % self.installroot + "/etc/dnf/dnf.conf")
result.append("--setopt=cachedir=%s" % self.installroot + "/var/cache/dnf")
```
I started CI tests without `--tags=dnf5`, so `get_dnf4_cmd` was used.

pkratoch: The tests pass with the change https://github.com/rpm-software-management/ci-dnf-stack/pull/1467

PR #1430, opened 2024-01-04
Add test for remove command to ensure not removing according to provides

j-mracek:
The new behavior is more stable for cases when the same spec is used multiple times with remove command.


dnf-behave-tests/fixtures/specs/dnf-ci-provides-alternatives/bar-1.0.spec
```diff
@@ -0,0 +1,23 @@
+%undefine _debuginfo_subpackages
+
+Name:           bar
+Version:        1.0
+Release:        1
+
+License:        Public Domain
```
ppisar, line 7: Is there a reason why bar-1.0.spec URL, License and %install differ from foo.spec? Could you make them the same, so it's clear in how the two spec files differ?

ppisar (commented):
Could you link the related DNF5 change <https://github.com/rpm-software-management/dnf5/pull/1126> from this commit message?

dnf-behave-tests/fixtures/specs/dnf-ci-provides-alternatives/bar-1.0.spec
j-mracek: Good point, I will unify it.

j-mracek:
@ppisar I think I've fixed it

dnf-behave-tests/dnf/remove-no-provides.feature
```diff
@@ -0,0 +1,27 @@
+@dnf5daemon
+Feature: Remove RPMs by provides
+
+
+Background: Install glibc
```
ppisar, line 5: The Backround text is wrong. It should read "Install two packages, both providing foo".

dnf-behave-tests/dnf/remove-no-provides.feature
```diff
@@ -0,0 +1,27 @@
...
+   Then the exit code is 0
+    And Transaction is following
+        | Action        | Package                         |
+        | install       | foo-0:1.0-1.noarch              |
+        | install       | bar-0:1.0-1.noarch              |
+
+
+
+@dnf5
+Scenario: Remove an RPM by name and nott by provide
```
ppisar, line 17: "nott" is a typo. Did you mean "not"?

dnf-behave-tests/dnf/remove-no-provides.feature
j-mracek: Thank you, fixed

dnf-behave-tests/dnf/remove-no-provides.feature
j-mracek: Thank you, fixed

PR #1434, opened 2024-01-09
Adjust versionlock locking tests to dnf5

m-blaha:
- change config format to new versionlock.toml format
- drop test for different patterns of locked versions. Globs are not
  supported any more.
- version pattern (1.3.*) test changed to proper versions interval
  (>=1.3 && < 1.4)

Requires: https://github.com/rpm-software-management/dnf5/pull/1121


m-blaha:
@jan-kolarik Ech sorry. I have it fixed locally, but forgot to commit and push...

PR #1437, opened 2024-01-11
Update actions versions and output method

rffontenelle:
Updates actions/checkout@v4 to the latest major version to drop deprecated nodejs version, and replace deprecated `::set-output` method with the [current method](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter).

This solves the warnings:

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

and

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/actions/copr-build/action.yml
+++ b/.github/actions/copr-build/action.yml
@@ -61,4 +61,4 @@ runs:
 
         # convert newlines to spaces (output variable assignment below doesn't support newlines)
         PACKAGE_URLS="$(tr '\n' ' ' < rpmlist)"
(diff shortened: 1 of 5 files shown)


inknos:
ah this was on my backlog for a long time! thanks

PR #1440, opened 2024-01-29
Tests for optional filelists metadata

jan-kolarik:
Basic tests for optional filelists metadata functionality taken and reworked from the existing dnf5 tests.
Also adjusted the existing dnf4 tests depending on filelists functionality.

Requires: https://github.com/rpm-software-management/libdnf/pull/1642.


jan-kolarik:
I am not sure about the microdnf test disabling, because [in the target PR](https://github.com/rpm-software-management/libdnf/pull/1642), there is still microdnf package used as the microdnf provider, although it should be already provided by dnf5 since F38.

PR #1444, opened 2024-02-08
Enable additional module tests

pkratoch:
Requires: https://github.com/rpm-software-management/dnf5/pull/1277

### DIFF
--- a/dnf-behave-tests/dnf/module/disable-errors.feature
+++ b/dnf-behave-tests/dnf/module/disable-errors.feature
@@ -1,3 +1,4 @@
+@dnf5
 Feature: Disabling module - error handling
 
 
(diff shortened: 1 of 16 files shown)


pkratoch:
The main differences from DNF4 in the stderr output are:

1. There is "Failed to resolve the transaction:" instead of "Error: Problems in request:"
2. There is "No match for argument: <modulespec>" instead of "Unable to resolve argument <modulespec>"
3. There is no summary of which groups or modules are missing or broken
4. The streams are sorted by alphabet (in e.g. "Argument 'nodejs' matches 4 streams ('10', '11', '12', '8') of module 'nodejs', but none of the streams are enabled or default.")

I think we might want to fix number 3, but the others seem fine to me unless there is some reason to change it.

PR #1449, opened 2024-02-16
Fix installonly pkgs tests with latest nightly libsolv

kontura:


m-blaha:
The patch looks good for me.
I'm only wondering whether this branch is also used for testing in environments with older libsolv installed?

jan-kolarik:
> But on this PR it seems the CI is not running at all for a change.. 😢
> 
> Missing `pgrep`, I will look into it.

There is some issue probably since containers were updated to F39 during the last week. Maybe try to ping Nicola.

m-blaha (approved):

m-blaha:
> Ok, it is working now.
> 
> @m-blaha If I understand correctly the daemon tests are still expected to fail.

Yes, till https://github.com/rpm-software-management/dnf5/pull/1235 gets merged.

PR #1452, opened 2024-02-19
Fix test for installing through provide of installonly pkg

kontura:
Since version 0.7.29 (currently unreleased) libsolv is improved and no longer tries to install unrequested pacakges.


m-blaha (approved):
Thank you!

PR #1454, opened 2024-02-20
Test upgrading kernel when over installonly limit with custom kernel

kontura:
Custom kernel doesn't contain `installonlypkg(kernel)` provide but provides `kernel`. libdnf needs to be able to handle such a situation even if the custom kernel is installed over limit.

https://bugzilla.redhat.com/show_bug.cgi?id=2263675


m-blaha (approved):
Thank you!

PR #1455, opened 2024-02-21
Add quay.io action

inknos:
Provides the images fedora-dnf5 and fedora-dnf5-nightly which come with dnf5 pre-installed

### DIFF
--- a/containerfiles/Containerfile
+++ b/containerfiles/Containerfile
@@ -1,0 +1,26 @@
+FROM fedora:$TAG
+
+ARG TAG
+ARG NAME
+
+RUN dnf -y update \
+    && dnf -y install 'dnf-command(copr)' \
+
+    # enable nightly repositories
+    && if [ "$NAME" == "dnf5-nightly" ]; then \
+        dnf copr -y enable rpmsoftwaremanagement/dnf-nightly ; \
+    fi \
+
+    # enable testing repositories
+    && if [ "$NAME" == "dnf5-testing" ] || [ "$NAME" == "dnf5-testing-nightly" ]; then \
+        dnf copr -y enable rpmsoftwaremanagement/"$NAME" ; \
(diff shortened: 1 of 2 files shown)


containerfiles/Containerfile
```diff
@@ -0,0 +1,27 @@
...
+
+# install all the DNF5 stack
+RUN dnf -y install dnf5 dnf-data dnf5-plugins
+
+# clean installed packages
+# from this on we need to be careful to specify which dnf to use
+RUN dnf4 -y remove 'dnf-command(copr)'
+#clean cache
+RUN dnf4 clean all
+RUN dnf5 clean all
```
praiskup, line 27: These cleanup actions need to be "squashed" into a single RUN command (single layer) otherwise the only effect is that it doesn't clean up anything (but it creates yet another layer).

inknos: oh right

praiskup: Actually :-) it seems that all the RUN commands should be merged into a single one here, looking like:
```
RUN dnf -y update \
    && dnf -y install 'dnf-command(copr)' \
    && if [ "$NAME" == "dnf5-nightly" ]; then \
        dnf copr -y enable rpmsoftwaremanagement/dnf-nightly ; fi \
    && ...
```

inknos: I could do that too. thanks

inknos: now that I think about it, I don't really know what's the recommended way of building containers, like, what options should be squashed together 🤔 is it ok now?

praiskup:
Looks good to me, thank you! (though I'd say that removing dnf4 from the image would be even better)

praiskup:
Agreed, this sounds like a good first step, thank you.  Can we merge?

PR #1460, opened 2024-02-26
Add test for case insensitive of list command

j-mracek:


dnf-behave-tests/dnf/list.feature
```diff
@@ -54,6 +54,18 @@ Given I drop repository "dnf-ci-fedora"
   setup.noarch +2.12.1-1.fc29 +dnf-ci-fedora
   """
 
+@dnf5
+Scenario: dnf list setup (when setup is installed) using incorrect case
```
ppisar, line 58: Could you rename the Scenario to "dnf list is case insensitive"? Your current text only paraphrases what's written in the next line without providing a motivation for the test.

j-mracek: Good suggestion. Thank you. Fixed

PR #1463, opened 2024-02-29
update repolist tests for containers for dnf5

glum23:
renamed:    microdnf/repolist.feature -> repolist-container.feature
renamed:    microdnf/repolist-disabled.feature -> repolist-disabled-container.feature
and updated for dnf5

Notes:
* microdnf alias instead of dnf is called intentionally
* scenarios with both --enable and --disable options have been changed to reflect current dnf5/microdnf behavior
* alias "repolist" in some scenarios is used intentionally
* tags no_installroot and destructive are used intentionally

The updated tests passed when run on a testing machine with dnf5-5.1.13-20240229004616.10.gc6d67a81.fc39

### DIFF
--- a/dnf-behave-tests/dnf/repolist-container.feature
+++ b/dnf-behave-tests/dnf/repolist-container.feature
@@ -1,6 +1,7 @@
+@dnf5
 @no_installroot
 @destructive
-Feature: Repolist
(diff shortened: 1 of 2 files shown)


kontura:
Thank you!

PR #1467, opened 2024-03-07
get_dnf5_cmd: Do not set "reposdir" and "config_file_path"

jrohel:
They are taken from installroot by default.
Setting it explicitly via the CLI will hide the default behavior. And it changes the behavior if dnf.conf is not present.

Moreover, it is more compatible with DNF4 tests. `get_dnf4_cmd` does not set "reposdir" and "config_file_path".

### DIFF
--- a/dnf-behave-tests/dnf/environment.py
+++ b/dnf-behave-tests/dnf/environment.py
@@ -235,8 +235,6 @@ class DNFContext(object):
         for key,value in setopts.items():
             result.append("--setopt={0}={1}".format(key, value))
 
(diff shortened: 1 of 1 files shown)


jan-kolarik:
Created a follow-up issue to address this automatic overriding of parameters in CI scenarios generally: https://github.com/rpm-software-management/ci-dnf-stack/issues/1468.

pkratoch:
The scenario passes with older librepo (when built without commit https://github.com/rpm-software-management/librepo/commit/66c99da1206c96fefc1b8279f777afefb79dc614 ), so the failure seems independent on this PR.

PR #1470, opened 2024-03-11
Add extended test for installonly packages

j-mracek:
The original test is ok, but this test is for a random behavior of the code, therefore it uses more additional packages in test to triger random behavior.

Required for: https://issues.redhat.com/browse/RHEL-15902

### DIFF
--- a/dnf-behave-tests/dnf/installonly.feature
+++ b/dnf-behave-tests/dnf/installonly.feature
@@ -258,6 +258,61 @@ Scenario: Do not remove or change reason after remove of one of installonly pack
         | action  | package                               | reason | repository |
         | Remove  | kernel-core-0:4.19.15-300.fc29.x86_64 | User   | @System    |
 
+# https://issues.redhat.com/browse/RHEL-15902
+# The test expects that installed packages in pool are in order in which they were installed. This is required for fail
+# without patch but it is not guaranteed by RPM
+Scenario: Do not remove or change reason after remove of one of installonly packages - more complex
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/installonly.feature
```diff
@@ -258,6 +258,58 @@ Scenario: Do not remove or change reason after remove of one of installonly pack
...
+        | install       | abcde-0:2.9.2-1.fc29.noarch           |
+        | install-dep   | wget-0:1.19.5-5.fc29.x86_64           |
+        | install-weak  | flac-0:1.3.2-8.fc29.x86_64            |
+   When I execute dnf with args "mark remove abcde"
+   Then the exit code is 0
+    And package reasons are
+        | Package                                | Reason          |
+        | abcde-2.9.2-1.fc29.noarch              | dependency      |
+        | flac-1.3.2-8.fc29.x86_64               | weak-dependency |
+        | wget-1.19.5-5.fc29.x86_64              | dependency      |
```
m-blaha, line 278: For the sake of future maintainers, I'd like to include a comment here explaining why the `abcde` package is needed on the system, with a `dependency` reason. I'm concerned that without it, anyone attempting to understand the test in a year or two might become confused.
What do you think?

j-mracek: Good point

j-mracek: I have created a comment.

m-blaha (approved):
Thank you!

PR #1471, opened 2024-03-11
Enable multiarch container builds

inknos:


praiskup (approved):
LGTM, needed for: https://github.com/rpm-software-management/mock/pull/1347

inknos:
@praiskup one thing that I am not sure about is if the tags for containers are the same archs references that fedora uses. like, is it aarch64 or arm64? 🤔

praiskup:
Mock uses just `podman pull <image_id>`, and Podman by default selects the appropriate architecture, if exists in the registry. See [this PR](https://github.com/rpm-software-management/mock/pull/1199) for more info. I think we should be fine.

praiskup:
Yeah, actually, I don't know how the "build time" architecture selection is implemented.  I rather meant that we (Mock) - at runtime - can use those images just fine (multiarch metadata/manifests). Hm.

inknos:
now it should be good to go.

I tested it on aarch64 images on quay.io and it works. Some images will fail if the nightlies are not ready but all the fedora-dnf5 images should be fine

.github/workflows/container-build-quay-io.yml
```diff
@@ -31,12 +31,18 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - name: Install qemu dependency
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-user-static
```
praiskup, line 37: This installation is emulated though, at least it seems (with the emulated runtime for rpm, but also installation scriptlets, etc.). This could make quite some difference compared to the official Fedora images..?

praiskup: > What difference should it do if we use emulated runtimes? 

I don't have anything specific in mind. It's just a bad feeling :-) because it's quite common for Mock builds with forcearch to fail for various reasons.

For example, the entire DNF in the build container is emulated (including RPM) for additional package installation. DNF/RPM interacts with the underlying databases, and these databases are supposed to be utilized by native executors later on.

I don't know; I wouldn't love it if my overexcitement caused issues with our production Copr builders :-) nb, we should have a "natively built" DNF5-based default Fedora 41 image soon.

inknos: > I don't know; I wouldn't love it if my overexcitement caused issues with our production Copr builders :-) nb, we should have a "natively built" DNF5-based default Fedora 41 image soon.

Agree, I'll dig some more I guess

PR #1475, opened 2024-03-20
System-upgrade, offline, and --offline tests

evan-goode:
Add tests for the `system-upgrade` and `offline` commands and the `--offline` argument.

The `system-upgrade` tests are adapted from DNF 4, but moved out of `dnf-behave-tests/dnf/plugins-core` since `system-upgrade` is now a regular command and not a plugin.


evan-goode:
Since this PR was still open, I added the tests for offline-distrosync and offline-upgrade here. So this patch now requires https://github.com/rpm-software-management/dnf5/pull/1344.

PR #1477, opened 2024-03-22
update scenarios for dnf5

glum23:
Various scenarios have been updated for dnf5. Primarily, tests that called (but not directly tested) "repolist" have been selected for this update. However, all scenarios that could be currently updated for dnf5 have been updated (not only those with repolist calls).
All updated tests passed when run directly in behave and via container-test for the latest available dnf5.

### DIFF
--- a/dnf-behave-tests/dnf/config-repos-overrides.feature
+++ b/dnf-behave-tests/dnf/config-repos-overrides.feature
@@ -24,7 +24,7 @@ Scenario: Test the overrides in "/usr/share/dnf5/repos.override.d"
       [dnf-ci-fedora-updates]
       enabled=1
       """
-   When I execute dnf with args "repolist"
+   When I execute dnf with args "repo list"
    Then the exit code is 0
     And stdout does not contain "dnf-ci-fedora\s+dnf-ci-fedora"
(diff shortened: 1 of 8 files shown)


glum23:
Some checks for dnf5daemon failed due to usage of "repo list" instead of "repolist", e.g. 
```
dnf5daemon-client -y --installroot=/tmp/dnf_ci_installroot_67ltmmd7 --releasever=29 --setopt=module_platform_id=platform:f29 --setopt=cachedir=/tmp/dnf_ci_installroot_67ltmmd7/var/cache/dnf repo list
...
Unknown argument "list" for command "repo". Add "--help" for more information about the arguments.
```
However, "repo list" should be now the preferred variant for dnf5, "repolist" is just for compatibility. Shouldn't dnf5daemon 
respect this, too?

pkratoch:
The dnf5daemon doesn't have command "repo list", but still uses "repolist". This causes a failure of scenarios in "gpg.feature" and "gpg-noeol.feature" when run for dnf5daemon. I filed an issue for this problem: https://github.com/rpm-software-management/dnf5/issues/1339 Until the issue is fixed, the failing tests cannot be modified to "repo list" and I am not even sure if it's a good idea to modify the passing ones, because it will prevent enabling them for dnf5daemon if needed.

pkratoch:
Thanks!

PR #1478, opened 2024-03-27
Don't include package download size in test outputs

pkratoch:
The size sometimes slightly differs and it causes random test failures. The rpm packages are not stored, but built from source each time the container is built, so it's reasonable to expect that package sizes may change.


kontura:
Thank you!

PR #1485, opened 2024-04-15
Update test to match curl error in all fedoras

kontura:
In f40 the curl error message changed from:
```
Protocol "xxxx" not supported or disabled in libcurl
```
to
```
Protocol "xxxx" not supported
```
update the test to match both.

### DIFF
--- a/dnf-behave-tests/dnf/config.feature
+++ b/dnf-behave-tests/dnf/config.feature
@@ -195,7 +195,7 @@ Scenario: Dnf prints reasonable error when remote config file is not downloadabl
    And stderr matches line by line
    """
    Config error: Configuration file URL "xxxx://localhost:[\d]+/does-not-exist\.conf" could not be downloaded:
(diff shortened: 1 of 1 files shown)


ppisar:
The error message changed in curl-8.5.1. See its CHANGES file:

~~~~
  Bumped to 8.5.1

- url: for disabled protocols, mention if found in redirect

  To help users better understand where the URL (and denied scheme) comes
  from. Also removed "in libcurl" from the message, since the disabling
  can be done by the application.

  The error message now says "not supported" or "disabled" depending on
  why it was denied:
   
   Protocol "hej" not supported
   Protocol "http" disabled
  
  And in redirects:

   Protocol "hej" not supported (in redirect)
   Protocol "http" disabled (in redirect)

  Reported-by: Mauricio Scheffer
  Fixes #12465
  Closes #12469
~~~~

Could you rephrase the commit message from "all fedoras" to "curl 8.5.1"? It will help people with picking the commit to different distributions.

kontura:
> Could you rephrase the commit message from "all fedoras" to "curl 8.5.1"? It will help people with picking the commit to different distributions.

Yes, that is better.
Done.

ppisar (approved):
Good.

PR #1488, opened 2024-04-19
Tests for repoquery -f, -l

jan-kolarik:
Related: https://github.com/rpm-software-management/dnf/pull/2083.

### DIFF
--- a/dnf-behave-tests/dnf/repoquery/files.feature
+++ b/dnf-behave-tests/dnf/repoquery/files.feature
@@ -1,0 +1,37 @@
+@dnf5
+Feature: Repoquery tests working with files
+
+
+Scenario: list files in an rpm including files in filelists.xml
+Given I use repository "repoquery-files"
+ When I execute dnf with args "repoquery a.x86_64 -l"
+ Then the exit code is 0
+  And stdout is
+      """
+      <REPOSYNC>
+      /root-file
+      /usr/bin/a-binary
+      """
+
+
+Scenario: filter by file in primary.xml
+Given I use repository "repoquery-files"
+ When I execute dnf with args "repoquery --file /usr/bin/a-binary"
+ Then the exit code is 0
+  And stdout is
+      """
+      <REPOSYNC>
+      a-0:1.0-1.fc29.x86_64
+      """
+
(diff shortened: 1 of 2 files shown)


jan-kolarik:
After I've created the tests for dnf4 branch, I realized these are already present in the dnf5, so no need to backport it there.

m-blaha:
Just thinking whether it would have some maintenance benefit to keep `repoquery/files.feature` the same (or as similar as possible) for dnf-4-stack and main branches. What do you think?

dnf-behave-tests/dnf/repoquery/files.feature
```diff
@@ -0,0 +1,23 @@
...
+Scenario: List of files in the package
+  Given I successfully execute dnf with args "repoquery -l filedep-dependency"
+   Then stdout is
+    """
+    /dummy/file
+    /filedep-dependency-1.0.spec
+    """
+
+
+Scenario: Query packages that owns a file
```
m-blaha, line 18: please add `@bz2276012` tag

jan-kolarik:
> Just thinking whether it would have some maintenance benefit to keep `repoquery/files.feature` the same (or as similar as possible) for dnf-4-stack and main branches. What do you think?

You're right, I'll try to reuse the tests from dnf5, drop the unusable ones and add the missing test for -l in dnf5.

m-blaha:
The `-l` is not completely missing - it just uses long variant of the option `--files`.

jan-kolarik:
I've backported the tests from the dnf5 branch and did small adjustments.

jan-kolarik:
Hmm, it seems the CI tests are failing because F39 environment is used for testing, though the filelists changes were introduced in >= F40.

jan-kolarik:
I think we should better migrate also the dnf4 branch of CI stack to use TMT testing, so we get the different chroots testing "for free".

jan-kolarik:
Hmm, so I forgot to actually add the testing repository that is used in the tests. Anyway, it's interesting no error about that non-existing repo was thrown.

m-blaha (approved):
Thanks!

PR #1492, opened 2024-04-24
Drop enabling of dnf5 plugins

jan-kolarik:
Since the enable/disable plugin CLI options are now used exclusively for libdnf5 plugins, drop the explicit enabling of any dnf5 plugins, as it will have no effect and could cause issues with warning messages in the output.


m-blaha:
Thanks for fixing it! Let me just wait for the test results...

m-blaha (approved):

PR #1495, opened 2024-04-24
Enable config-best.feature tests

m-blaha:
Since they are passing with dnf5...


m-blaha:
Oh, it requires skipped packages in the transaction table. I'll add the commit to https://github.com/rpm-software-management/ci-dnf-stack/pull/1494

PR #1498, opened 2024-04-29
Modify tests or outpit check according to new format with repo names

j-mracek:
Message was modified from `Warning: skipped PGP checks for 1 package(s).` to
`Warning: skipped PGP checks for 1 package(s) from comps-group repository(s).`

Requires: https://github.com/rpm-software-management/dnf5/pull/1459


pkratoch:
Just a small typo in the commit message: "Modify tests or outpit check" -> "Modify output check". Also, the following lines of the commit message seem unnecessary, since the change is clear from the diff, but it's not a problem.

PR #1500, opened 2024-05-02
Adjust to error messages containing repo

m-blaha:
Solver errors now also include the repository from which the problematic
package originates.

Requires: https://github.com/rpm-software-management/dnf5/pull/1464

### DIFF
--- a/dnf-behave-tests/dnf/distro-sync.feature
+++ b/dnf-behave-tests/dnf/distro-sync.feature
@@ -182,9 +182,9 @@ Scenario: distro-sync all with a broken dependency and without best
     And stderr is
     """
     Failed to resolve the transaction:
-    Problem: package labirinto-2.0-1.noarch requires labirinto-libs = 2.0-1, but none of the providers can be installed
+    Problem: installed package labirinto-2.0-1.noarch requires labirinto-libs = 2.0-1, but none of the providers can be installed
       - labirinto-libs-2.0-1.noarch does not belong to a distupgrade repository
-      - problem with installed package 
+      - problem with installed package
     You can try to add to command line:
       --skip-broken to skip uninstallable packages
     """
--- a/dnf-behave-tests/dnf/broken-dependencies-report.feature
+++ b/dnf-behave-tests/dnf/broken-dependencies-report.feature
@@ -16,10 +16,10 @@ Scenario: Broken dependencies are reported when strict and best options are off
    Then the exit code is 0
     And stderr is
     """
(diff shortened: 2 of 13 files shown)


dnf-behave-tests/dnf/distro-sync.feature
```diff
@@ -182,9 +182,9 @@ Scenario: distro-sync all with a broken dependency and without best
     And stderr is
     """
     Failed to resolve the transaction:
-    Problem: package labirinto-2.0-1.noarch requires labirinto-libs = 2.0-1, but none of the providers can be installed
-      - labirinto-libs-2.0-1.noarch does not belong to a distupgrade repository
-      - problem with installed package 
+    Problem: package labirinto-2.0-1.noarch from @System requires labirinto-libs = 2.0-1, but none of the providers can be installed
+      - labirinto-libs-2.0-1.noarch from @System does not belong to a distupgrade repository
```
j-mracek, line 186: What about to split message and when the problem is related to installed packages write something like `installed package {} does not ...`. What do you think?

m-blaha: I was thinking about it - the `@System` repo name is a bit cryptic for a new user.
We can split `SOLVER_RULE_PKG_REQUIRES` according the repo into two variants of messages - `ProblemRules::RULE_PKG_CONFLICTS` or `ProblemRules::RULE_INSTALLED_PKG_CONFLICTS`.
But there are more solver rules that would need to be split. For example `SOLVER_RULE_PKG_CONFLICTS` would need three variants - source installed, target installed, none installed (not sure whether there can be variant where both source and target packages are installed).
I was even thinking about having only one `{}` placeholder in the message and filling it either with `package <nevra> from <repoid>` or with `installed package <nevra>` "submessage" but considered this variant complicated for translators - I was not sure whether splitting a sentence into multiple translation strings would correctly work for every language.

j-mracek: I was thinking about approach that we used in DNF4 for `    {RULE_PKG_INSTALLED_OBSOLETES, M_("installed package %s obsoletes %s provided by %s from %s")},
    {RULE_PKG_IMPLICIT_OBSOLETES, M_("package %s from %s implicitly obsoletes %s provided by %s from %s")},`. 
This approach is easy for translation but requires to extend enum. The new items must be added as last items. What do you think?

j-mracek:
/packit test https://github.com/rpm-software-management/dnf5/pull/1464

PR #1501, opened 2024-05-09
Add coverage for improved countme system age

dmnks:
This updates the countme feature file in accordance with https://github.com/rpm-software-management/libdnf/pull/1662.


ppisar:
I think we also need to add the machine-id code into dnf5.

dmnks:
Yup, porting this to dnf5 was the intention, indeed. I made the patch in dnf4 just because that's what I was most familiar with.

It's probably best to first let the dnf4 patch go through the usual review process and only then create a dnf5 version of it, to keep things more manageable. I can help with the latter too (or just do it myself, the related code is basically the same so it shouldn't be too hard).

mattdm:
> It's probably best to first let the dnf4 patch go through the usual review process and only then create a dnf5 version of it, to keep things more manageable. I can help with the latter too (or just do it myself, the related code is basically the same so it shouldn't be too hard).

Can we make sure that the change only lands in rawhide and f41+? That way, I'll know where/when the change happened and can take that into account in interpreting the data.

jan-kolarik:
> > It's probably best to first let the dnf4 patch go through the usual review process and only then create a dnf5 version of it, to keep things more manageable. I can help with the latter too (or just do it myself, the related code is basically the same so it shouldn't be too hard).
> 
> Can we make sure that the change only lands in rawhide and f41+? That way, I'll know where/when the change happened and can take that into account in interpreting the data.

Yep, I'll sync about that with our colleague doing releases.

jan-kolarik:
> @jan-kolarik is dnf4 shipped in f41/rawhide at all? I'm thinking if it makes sense to make this change in dnf4 and if we shouldn't just skip dnf4 entirely.

Ehm, good point :smile: Well, basically it is not, as the default is already the dnf command from dnf5 package. So this would only affect users explicitly using `dnf-3` binary (or `dnf4` symlink).

dmnks:
OK, therefore the dnf4 patch still makes sense to pursue, I suppose.

I wonder, however, if we shouldn't apply the patch conditionally (in both the libdnf4 and libdnf5 specs) only on F41+ then? 

Having a dnf config option to choose whether the age reported should be absolute (based on the machine-id timestamp) or relative (based on the first countme flag sent for the given repo) would be another option. It would make it easier/cleaner to control on which distros this is enabled, although it might as well be overkill, so I'm not sure.

WDYT?

jan-kolarik:
@dmnks Sorry, I somehow missed your comment.

> OK, therefore the dnf4 patch still makes sense to pursue, I suppose.

I think so. If we then port it to dnf5, the work should be easier, and we'll cover all dnf clients in Rawhide, as some may use dnf5 and others dnf4 in F41.

> I wonder, however, if we shouldn't apply the patch conditionally (in both the libdnf4 and libdnf5 specs) only on F41+ then?

We usually try to keep the latest Fedora and upstream in sync, so I'd prefer to apply the revert patch for F40 and F39.

If no one else takes it, I'd like to review your PR in the coming days. Thanks!

dmnks:
> @dmnks Sorry, I somehow missed your comment.

No worries.

> I think so. If we then port it to dnf5, the work should be easier, and we'll cover all dnf clients in Rawhide, as some may use dnf5 and others dnf4 in F41.

Ack!

> We usually try to keep the latest Fedora and upstream in sync, so I'd prefer to apply the revert patch for F40 and F39.

Ack, whatever works for you best :+1: 

> If no one else takes it, I'd like to review your PR in the coming days. Thanks!

Sounds like a plan, thank you! If/when you get around to it, feel free to ping me on Slack if you need any help.

jan-kolarik:
> I think we also need to add the machine-id code into dnf5.

I've created the dnf5 counterpart for this: https://github.com/rpm-software-management/dnf5/issues/1525.

jan-kolarik:
@dmnks Oh, I just noticed that now - this needs to be targeted against `dnf-4-stack` as the `main` branch is for dnf5 stuff...

dmnks:
Oh... Thanks. I should open a new PR against `dnf-4-stack`, correct?

jan-kolarik:
> Oh... Thanks. I should open a new PR against `dnf-4-stack`, correct?

Probably yes, I don't know if you can switch the target branch of the existing PR.

dmnks:
Ported: https://github.com/rpm-software-management/ci-dnf-stack/pull/1512

kontura:
I have added a couple more commits to enable the tests for dnf5.
The tests still require both: https://github.com/rpm-software-management/dnf5/pull/1613 and https://github.com/rpm-software-management/dnf5/pull/1590.

jan-kolarik (approved):

jan-kolarik:
@dmnks and @kontura Thanks for making this functionality extensively covered with tests guys!

PR #1502, opened 2024-05-10
Add tests for --providers-of=.. with --installed

kontura:
Requires: https://github.com/rpm-software-management/dnf5/pull/1481


jan-kolarik:
Can we also explicitly check that `<REPOSYNC>` is not present when using `--installed`?

kontura:
Right I should actually remove it from the `stdout is` step.

Or do you mean something like: stdout doesn't contains "Updating and loading repositories"?

jan-kolarik:
> Right I should actually remove it from the stdout is step.

Yep, this is what I was thinking about.

jan-kolarik:
Thanks!

PR #1503, opened 2024-05-10
Adjust the container definition to work when the dnf command is dnf5

jan-kolarik:


dnf-behave-tests/requirements.spec
```diff
@@ -27,6 +27,7 @@ BuildRequires:  openssl
 BuildRequires:  python3
 BuildRequires:  python3-distro
 BuildRequires:  python3-pip
+BuildRequires:  python3-rpm
```
kontura, line 30: Can you add a comment or a note into the commit message why is this needed?

jan-kolarik: I didn't investigate deeply, but I think it should've already been there before, as it is a direct dependency of the CI tests. But it was probably fetched by some dnf4 package instead.

PR #1505, opened 2024-05-13
createrepo_c: account for repo sync messages when checking stdout

kontura:
Sometimes repository synchronization messages can be present in `stdout`.


jan-kolarik:
Thanks!

PR #1506, opened 2024-05-16
Enable append options tests for dnf5

m-blaha:
Requires: https://github.com/rpm-software-management/dnf5/pull/1495


jan-kolarik:
Thanks!

PR #1511, opened 2024-05-30
Add several tests for check-upgrade with advisories and minimal options

j-mracek:
Some of the tests are related to new option --minimal for check-upgrade

### DIFF
--- a/dnf-behave-tests/dnf/security-advisory.feature
+++ b/dnf-behave-tests/dnf/security-advisory.feature
@@ -1,5 +1,5 @@
 @dnf5
-Feature: Test upgrade-minimal with advisory, cve, secseverity
+Feature: Test upgrade, upgrade-minimal, and check-ugrade with advisory, cve, secseverity
 
 
 Background: Use repository with advisories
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/security-advisory.feature
```diff
@@ -55,6 +98,15 @@ Scenario: upgrade-minimal sec-severity
         | Action        | Package                   |
         | upgrade       | advisory_B-0:1.0-2.x86_64 |
 
+Scenario: check-upgrade --minimal with sec-severity
+   When I execute dnf4 with args "upgrade-minimal --sec-severity Moderate"
```
kontura, line 102: I believe the `I execute dnf4` step shouldn't be here, it is not being run, it is a wrong command and these tests are only for dnf5.

dnf-behave-tests/dnf/security-advisory.feature
```diff
@@ -64,6 +116,15 @@ Scenario: upgrade-minimal with pkgs specified sec-severity
         | Action        | Package                   |
         | upgrade       | advisory_B-0:1.0-2.x86_64 |
 
+Scenario: check-upgrade --minimal with pkgs specified sec-severity
+   When I execute dnf4 with args "upgrade-minimal advisory_B --sec-severity Moderate"
```
kontura, line 120: I believe the `I execute dnf4` step shouldn't be here, it is not being run, it is a wrong command and these tests are only for dnf5.

kontura (changes requested):

dnf-behave-tests/dnf/security-advisory.feature
j-mracek: Thank you - I've replaced dnf4 and dnf5 by simple `dnf`

dnf-behave-tests/dnf/security-advisory.feature
j-mracek: fixed

kontura (approved):
Thank you

PR #1517, opened 2024-06-25
fileprovides: create fileprovides with already present solv files

kontura:
Since dnf4 now also conditionally load filelists it ran into the same problem as dnf5 here: For: https://github.com/rpm-software-management/dnf5/issues/520 Backport the test.

For: https://github.com/rpm-software-management/libdnf/pull/1670


ppisar (approved):
It's the same patch as in commit 92211173018da5c0f33d41c0e7f0a1ac60e49e60 for dnf5. The only failed CI test was this new test because it was run against old libdnf.

PR #1523, opened 2024-07-01
Enable several tests for debuginfo-install command

j-mracek:
There are remaining 6 scenario that requires additional reporting functionality.


m-blaha (approved):
Thank you!

PR #1524, opened 2024-07-02
Enable tests for missing debug packages

j-mracek:


m-blaha:
Thank you! If I'm not mistaken, now each of the scenarios is tagged with `dnf5`. Would you mind removing all those tags and replacing them with one on the `Feature`?

PR #1525, opened 2024-07-02
Update and enable tests for transaction replay

kontura:
Requires: https://github.com/rpm-software-management/dnf5/pull/1569

### DIFF
--- a/dnf-behave-tests/dnf/transaction-sr/replay.feature
+++ b/dnf-behave-tests/dnf/transaction-sr/replay.feature
@@ -463,13 +463,9 @@ Given I create file "/{context.dnf.tempdir}/transaction.json" with
       | mid-a2-1.0-1.x86_64    | Weak Dependency |
       | top-a-1:2.0-1.x86_64   | User            |
       | top-b-1.0-1.x86_64     | Group           |
-  And I execute "echo 'select * from comps_group_package;' | sqlite3 -noheader -list {context.dnf.installroot}/var/lib/dnf/history.sqlite"
-  And stdout is
-      """
-      1|6|top-a|1|4
-      2|6|top-b|1|2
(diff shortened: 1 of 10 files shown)


dnf-behave-tests/dnf/transaction-sr/replay.feature
```diff
@@ -1140,6 +1140,9 @@ Given I successfully execute dnf with args "install archchange-1.0"
       | top-a-1:1.0-1.x86_64    | User            |
 
 
+# This should fail or there should be at least a warning.
+# Reported as https://github.com/rpm-software-management/dnf5/issues/1573
+@xfail
```
pkratoch, line 1145: I was a bit confused by this, because the `xfail` tests are usually written according to the expected behavior, but here there is the undesired (though currently working) behavior. Based on the discussion in the issue, I think dnf should report an error. Do I understand it correctly? If so, can you please change the expected outcome?

pkratoch (commented):
Thanks for all the patches!

Based on the comment in https://github.com/rpm-software-management/dnf5/pull/1569 regarding the missing `package_types` in the json format, I think it would be good to have a test for that as well (either test that it's a required field or test that a default value is used).

dnf-behave-tests/dnf/transaction-sr/replay.feature
kontura: Yes, sure, updated.

kontura:
> Thanks for all the patches!
> 
> Based on the comment in [rpm-software-management/dnf5#1569](https://github.com/rpm-software-management/dnf5/pull/1569) regarding the missing `package_types` in the json format, I think it would be good to have a test for that as well (either test that it's a required field or test that a default value is used).

That is a good idea.
I have added a new commit with that test.

PR #1532, opened 2024-07-26
Add `--json` tests for `repo` and `advisory list` commands  

kontura:
- Add a step to compare json outputs
- Add `--json` tests for `repo` and `advisory list` commands 
- Removes unused dnf4 parts from updateinfo.feature

### DIFF
--- a/dnf-behave-tests/dnf/repo-json.feature
+++ b/dnf-behave-tests/dnf/repo-json.feature
@@ -1,0 +1,199 @@
+@dnf5
+Feature: Repo command with --json option
+
+
+Background: Using repositories dnf-ci-fedora and dnf-ci-thirdparty-updates
+  Given I use repository "dnf-ci-fedora"
+    And I use repository "dnf-ci-thirdparty-updates"
+    And I use repository "dnf-ci-fedora-updates" with configuration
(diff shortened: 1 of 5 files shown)


dnf-behave-tests/dnf/repo-json.feature
```diff
@@ -0,0 +1,151 @@
...
+      {
+        "id":"dnf-ci-thirdparty",
+        "name":"dnf-ci-thirdparty test repository",
+        "is_enabled":false
+      }
+    ]
+    """
+
+
+Scenario: Repo list without arguments --json --refresh --quiet
```
m-blaha, line 134: Is there a reason for this scenario? It seems to be essentially the same as the `Repo list without arguments --json`, and I'm not sure why the --refresh and --quiet options are important. Could you clarify their significance?

kontura: It was meant to test that `<REPOSYNC>` doesn't get into the stdout.
It was really confusing and wrong.

The `--quiet` is not needed because its implicitly set by `--json` and I have changed it to `repo info` because `repo list` doesn't need metadata. 
I have updated it.
 
Thank you for catching that.

PR #1534, opened 2024-08-05
builddep: add --with/without tests and update err msg

kontura:
Requires: https://github.com/rpm-software-management/dnf5/pull/1509


kontura:
/packit test https://github.com/rpm-software-management/dnf5/pull/1509

PR #1535, opened 2024-08-14
Remove debug-* tests: "debug" plugin was removed from dnf-plugins-core

martinobanana:
Following tests don't make sense on RHEL-10 because the plugin is no longer available in dnf-plugins-core.

debug-dump.feature
debug-restore-obsoletes.feature
debug-restore.feature


glum23:
DNF CI / Integration Tests (dnf, dnf) (pull_request_target) failures unrelated to the removal, merging

PR #1540, opened 2024-08-29
Check for diagnostic messages on stderr, not stdout

evan-goode:
For https://github.com/rpm-software-management/dnf5/pull/1641.

Adds support for the `<REPOSYNC>` shorthand to `Then stderr is` and `Then stderr matches line by line`.

Introduces `Then stdout contains lines matching` and `Then stderr contains lines matching` steps which are similar to `Then stdout contains lines` but treat each expected line as a regex, like `Then stdout matches line by line`.

This patchset strengthens many tests; there were many instances of the pattern:

```cucumber
Then stdout is
     """
     <REPOSYNC>
     foo
     bar
     """
```

that were replaced with

```cucumber
Then stderr is
     """
     <REPOSYNC>
     """
 And stdout is
     """
     foo
     bar
     """
```

so now stderr is checked in many places where it wasn't before.

However, some tests were weakened, notably those following the pattern:

```cucumber
Then stderr is
     """
     No match for argument: nosuchpkg

     Warning: skipped PGP checks for 1 package from repository: dnf-ci-fedora-updates
     """
```

Because stderr now includes lots of other (non-error) information including progress bars for downloads and RPM transactions, these occurrences were changed to:

```cucumber
Then stderr contains lines
     """
     No match for argument: nosuchpkg
     Warning: skipped PGP checks for 1 package from repository: dnf-ci-fedora-updates
     """
```

We could maybe make `<RPMDOWNLOAD>` and `<RPMTRANSACTION>` markers that work similar to `<REPOSYNC>` to allow us to use `Then stderr is` again, but I'd rather do that in a follow-up issue.

These changes were made mostly manually with the help of Neovim macros.

### DIFF
--- a/dnf-behave-tests/common/output.py
+++ b/dnf-behave-tests/common/output.py
@@ -143,13 +143,38 @@ def then_stderr_is_empty(context):
 
 @behave.then("stderr is")
 def then_stderr_is(context):
+    """
+    Checks for the exact match of the test's stderr. Supports the <REPOSYNC>
+    placeholder on the first line, which will match against the repository
+    synchronization lines (i.e. the "Last metadata expiration check:" line as
+    well as the individual repo download lines) in the test's output.
+    """
     expected = context.text.format(context=context).strip().split('\n')
     found = context.cmd_stderr.strip().split('\n')
 
-    if expected == found:
+    if found == [""]:
+        found = []
+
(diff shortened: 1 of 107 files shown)


jan-kolarik:
/packit test https://github.com/rpm-software-management/dnf5/pull/1641

jan-kolarik (approved):
LGTM, tests were run locally successfully. And thanks for adding some docs to the python functions here in the CI stack!

PR #1546, opened 2024-09-04
Fix the format of file conflicts error message

jan-kolarik:
I think this was introduced with https://github.com/rpm-software-management/ci-dnf-stack/pull/1540 and I forgot to include/request the fix there...


evan-goode:
I'd rather fix this by https://github.com/rpm-software-management/dnf5/pull/1677/files, I didn't mean to remove that `-` in the output.

jan-kolarik:
> I'd rather fix this by https://github.com/rpm-software-management/dnf5/pull/1677/files, I didn't mean to remove that `-` in the output.

Agree, thanks!

PR #1550, opened 2024-09-17
Add missing `Recommends` to simulate real debuginfo pkgs

kontura:
Since `rpm-4.19.94-1.fc42` the rpm debuginfo code logic is not always executed. To match real debuginfo pkgs add debugsource recommends manually.

More details: https://github.com/rpm-software-management/rpm/issues/3310

### DIFF
--- a/dnf-behave-tests/fixtures/specs/debuginfo-install/bar-1.0.spec
+++ b/dnf-behave-tests/fixtures/specs/debuginfo-install/bar-1.0.spec
@@ -16,6 +16,7 @@ Dummy.
 
 %package debuginfo
 Summary:        Debug information for foo.
+Recommends:     %{name}-debugsource(x86-64) = %{version}-%{release}
 
 %description debuginfo
 Dummy.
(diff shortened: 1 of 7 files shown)


pkratoch:
Thanks! I am just thinking, shouldn't it be added also to other debuginfo packages, as it's something that rpm would do if these were real debuginfo packages? Also, dnf4 is not in Fedora rawhide, but the affected rpm version might get into RHEL or other Fedora releases, so we should perhaps make this change also for dnf-4-stack.

kontura:
> Thanks! I am just thinking, shouldn't it be added also to other debuginfo packages, as it's something that rpm would do if these were real debuginfo packages? 

Ok, I have added it to other debuginfo - debugsource pairs in the `debuginfo-install` repo. I have checked and there are two additional pairs in `dnf-ci-fedora-updates` (`lz4` and `zstd`) but those already have the recommends.

> Also, dnf4 is not in Fedora rawhide, but the affected rpm version might get into RHEL or other Fedora releases, so we should perhaps make this change also for dnf-4-stack.

That is a good suggestion I will backport the changes to `dnf-4-stack` branch. :+1:

PR #1552, opened 2024-09-18
Add test for repoquery on repository with xml:base

m-blaha:
See https://github.com/rpm-software-management/dnf/issues/2130


kontura:
Thanks

PR #1555, opened 2024-09-25
Update tests to match curl error messages

martinobanana:
Error messages in new curl changed and the test needed to be updated. The diffs were mostly "Couldn't" -> "Could not" and "host name" -> "hostname"


martinobanana:
This has been tested on RHEL-10 / centosstream10 to work correctly

glum23:
Integration tests fail because they were run on Fedora, not RHEL-10. The tests passed on RHEL-10 with the proper version of curl. Merging.

PR #1556, opened 2024-09-25
Update tests to match curl error messages

martinobanana:
Error messages in new curl changed and the test needed to be updated. The diffs were mostly "Couldn't" -> "Could not" and "host name" -> "hostname"


martinobanana:
This is actually a little bit tricky, it's the same commit I did for RHEL-10 (curl-8.9.1-4.el10.x86_64) where it works. But here it's complicated because due to how the outputs are checked (exact match) it depends on version of curl on the test system. Which can be different for all systems with dnf-4.

kontura:
Oh, sorry, I totally forgot about this PR :man_facepalming:. I did the same thing as part of https://github.com/rpm-software-management/ci-dnf-stack/pull/1569.

I am proposing to disable the tests with updated curl messages on Fedora < 41. That way they don't fail and going forward they will be run on all active fedoras again (in some time).

If you prefer we can finish the change here.

PR #1559, opened 2024-10-04
Update container to f40

kontura:


kontura:
This side-steps the failing builds (which fail due to: https://bugzilla.redhat.com/show_bug.cgi?id=2315859) and f39 will be EOL soon anyway.

PR #1560, opened 2024-10-07
Lowercase branches when constructing an image identifier

ppisar:
Executing a pull request test from a branch with upper case letters ended up with this error:

        Command '/var/ARTIFACTS/work-behave-createrepo_crye9wxe5/plans/integration/behave-createrepo_c/tree/tmt-prepare-wrapper.sh-Build-testing-container-default-0' returned 1.

        stderr (2 lines)
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        Error: tag container-test-unistd-for-STDOUT_FILENO-master: invalid reference format: repository name must be lowercase
        Error: Failed to build the container.
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

The cause was that -c argument of container-test script was built from a "container-test-unistd-for-STDOUT_FILENO" branch name. The container-test -c value is then directly passed as an image identifier to podman build and podman run commands. Problem is that podman, as well as docker, does not allow upper-case characters in the OCI identifier.

I wanted to circumvent it in the container-test script, to have a fix at one place. But one can pass a tag name there (foo/bar:tag) and the tag is handled case sensitively by docker.

So this patch fixes it on the invocation side.

NOTE: I could not really test it. I only tried a prepare phase of locally run TMT and it correctly transliterated the branch names. You should be able to fix by rerunning tests for
<https://github.com/rpm-software-management/createrepo_c/pull/439>.

### DIFF
--- a/plans/integration/behave-createrepo_c.fmf
+++ b/plans/integration/behave-createrepo_c.fmf
@@ -5,4 +5,4 @@ execute:
     script: |
         $TMT_PLANS_DATA/ci-dnf-stack/container-test \
         --suite createrepo_c \
-        -c container-test-$PACKIT_SOURCE_BRANCH-$PACKIT_TARGET_BRANCH run
+        -c container-test-${PACKIT_SOURCE_BRANCH,,?}-${PACKIT_TARGET_BRANCH,,?} run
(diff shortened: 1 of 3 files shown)


kontura:
Lets try it.

ppisar:
Thanks. It helped. Now most of tests for that crearerepo_c pull requests passed. The remaining three are a completely different errors.

PR #1561, opened 2024-10-09
Adjust reposync tests to xattrs renamed in librepo

m-blaha:
Requires: https://github.com/rpm-software-management/librepo/pull/327


m-blaha:
I was thinking about running the test on different Fedora/RHEL versions which could use older librepo.

kontura:
I believe we have branches for that but in this case I guess it is ok.

PR #1563, opened 2024-10-09
Recommend --use-host-config if --installroot is used and not all repositories can be enabled

ppisar:
Adapt tests to a new error message wording when an installroot is
used and a repository cannot be enabled or disabled.
    
Related: https://github.com/rpm-software-management/dnf5/pull/1760


kontura (approved):
Looks good.

PR #1564, opened 2024-10-09
revise config.feature for dnf5

m-blaha:
- there still is one failing scenario (remote config files, https://github.com/rpm-software-management/dnf5/issues/1767)
- passing scenarios were tagged with dnf5 (with some small adjustments)
- dnf5 does not support automatic search of the config file in host/installroot location, such tests were dropped / replaced with test for `--use-host-config` option


kontura:
Thank you!

PR #1567, opened 2024-10-14
Add obsoletes tests for yum behavior

kontura:
In https://issues.redhat.com/browse/RHEL-61475 I have noticed `dnf-3` installed the obsoleters as `Dependency`, it seems `dnf5` handles this better.

I am not sure if it is worth it to try to fix it in `dnf-3` since I believe the situation should be quite rare.


jan-kolarik:
Thanks for covering this!

PR #1568, opened 2024-10-15
Review [a-d]* tests for dnf5

m-blaha:
- added reference to related github issues for failing tests
- passing tests are enabled/adjusted
- tests that does not make sense any more are dropped
- test for aliases functionality is re-worked to cover dnf5 implementation


kontura (approved):
Great, thank you!

Issue #1554, opened 2024-09-20
Fix dnf4 tests for Fedora 41+


jan-kolarik:
Since dnf5 is the default on Fedora 41+, it's not currently possible to run dnf4 CI tests there (see [here](https://artifacts.dev.testing-farm.io/a331110e-1c46-4aeb-b2a8-473e13e1b685/)). We should address this by explicitly using the `dnf4` command from the `python3-dnf` package to ensure dnf4 is tested in these environments.

PR #1569, opened 2024-10-16
Update dnf4 tests for Fedora 41+

kontura:
For: https://github.com/rpm-software-management/ci-dnf-stack/issues/1554

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,9 +40,19 @@ RUN set -x && \
 # copy test suite
 COPY ./dnf-behave-tests/ /opt/ci/dnf-behave-tests
 
+# remove dnf5 and exclude it to ensure we test old dnf
+RUN set -x && \
+    dnf -y remove dnf5 --setopt=protected_packages=,; \
+    echo "excludepkgs=dnf5*" >> /etc/dnf/dnf.conf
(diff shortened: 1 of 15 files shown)


Dockerfile
```diff
@@ -40,9 +40,19 @@ RUN set -x && \
...
+# remove dnf5 and exclude it to ensure we test old dnf
+RUN set -x && \
+    dnf -y remove dnf5 --setopt=protected_packages=,; \
+    echo "excludepkgs=dnf5*" >> /etc/dnf/dnf.conf
+
+# On Fedora > 40 the symlinks to dnf-3 and yum are missing (because dnf5 provides dnf), add them manually
+RUN set -x && \
+    ln -sf /usr/bin/dnf-3 /usr/bin/dnf; \
+    ln -sf /usr/bin/dnf-3 /usr/bin/yum
+
```
jan-kolarik, line 52: That's nice, better than redirecting to dnf4 command everywhere. Thanks!

jan-kolarik (approved):
Thanks a lot for taking care of this! I think something similar will be also needed when moving from RHEL 10 to RHEL 11, so we can reuse this then.

PR #1571, opened 2024-10-18
Add test for dnf-automatic with non-existing config file

m-blaha:
See https://issues.redhat.com/browse/RHEL-46030

Requires: https://github.com/rpm-software-management/dnf/pull/2146

### DIFF
--- a/dnf-behave-tests/dnf/dnf-automatic/config.feature
+++ b/dnf-behave-tests/dnf/dnf-automatic/config.feature
@@ -1,0 +1,11 @@
+Feature: dnf-automatic configuration files testing
+
+# https://issues.redhat.com/browse/RHEL-46030
+Scenario: dnf-automatic fails if non-existing config file is specified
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/dnf-automatic/config.feature
```diff
@@ -0,0 +1,20 @@
+@no_installroot
+Feature: dnf-automatic configuration files testing
+
+Background:
+Given I delete file "/etc/yum.repos.d/*.repo" with globs
+  And I create file "/etc/dnf/dnf.conf" with
+    """
+    [main]
+    plugins=0
+    """
```
kontura, line 10: I think this is not needed here but I am fine with keeping it.

Could you add a comment or name the background (`Backgroud: setup clean environment`) to make its purpose clear?

m-blaha: Right. It's a copy paste from another automatic test, but it's not needed here. I'll remove the background section. It doesn't even need to be run outside of the installroot.

PR #1578, opened 2024-10-22
Review e-g features

kontura:
`encoding.feature`: It no longer expects an encoding error because dnf5 handles the `invalid_utf8_char` normally.


dnf-behave-tests/dnf/gpg.feature
```diff
@@ -229,7 +222,5 @@ Scenario: Expire repo when failed to install package with incorrect checksum
         | Action        | Package                               |
         | install       | broken-package-0:0.2.4-1.fc29.noarch  |
     And RPMDB Transaction is empty
-    And file "/var/cache/dnf/expired_repos.json" contents is
-        """
-        ["dnf-ci-gpg"]
-        """
+   When I execute dnf with args "rq"
+   Then stderr is empty
```
pkratoch, line 226: Can you please help me understand what is being checked here? Also, can you add a comment to the test that would describe it, so that it's immediately clear?

kontura: Oh, I forgot to finish that test.
I have updated it.

PR #1579, opened 2024-10-23
Review tests in microdnf directory

kontura:
A lot of the tests are duplicate to regular dnf5 tests but they are run through the `microdnf` command.


pkratoch:
Is it really worth it to keep the microdnf tests in the separate microdnf directory if it's just an alias for the dnf5 command? In past, we had run tests explicitly with "dnf", "dnf-3" and "yum", maybe it was even possible to switch the command for all the tests, I am not sure exactly. But having the test at one place and only switching the command name seems better than to have a somewhat duplicate sets of tests. I can imagine it might be a lot of work to change it, but it would be more maintainable later.

kontura:
I have update the PR to de-duplicate and merge the `microdnf` tests.
I really like this approach much better, great idea. :+1:

PR #1580, opened 2024-10-23
Test for groups with duplicate package entry

m-blaha:
See https://bugzilla.redhat.com/show_bug.cgi?id=2263762

Requires: https://github.com/rpm-software-management/dnf5/pull/1798

### DIFF
--- a/dnf-behave-tests/dnf/comps-group.feature
+++ b/dnf-behave-tests/dnf/comps-group.feature
@@ -697,3 +697,13 @@ Scenario: Remove group that is not installed
        <REPOSYNC>
        No groups to remove for argument: dnf-ci-testgroup
        """
+
+@dnf5
+@bz2263762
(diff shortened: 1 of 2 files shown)


ppisar:
Waiting a day for nightly repository to get dnf5 build from 47bd21dc6f7a193a64633f07bc839d1d0ad88668 commit added with https://github.com/rpm-software-management/dnf5/pull/1798.

ppisar:
The dnf-nightly repository has not been updated since 2024-10-10. There are dnf5 builds failing for EPEL9 because of the random PGP key bug in librepo (already fixed upstream). I tried rebuilding them. I'm not sure whether Copr stops publishing repositories until all buildroots succeeds. But I think it does not and there is maybe something else wrong with the nightly repository.

ppisar:
Now the repository has been publish and all builds are green (even the historical ones that failed?!).

PR #1582, opened 2024-10-29
Fix `encoding.feature` test with connected lines

kontura:
Requires: https://github.com/rpm-software-management/dnf5/pull/1805


m-blaha (approved):
Thank you!

PR #1584, opened 2024-11-01
Review k*-p* feature files for dnf5

pkratoch:
So far except for plugin* feature files, since these will probably need bigger changes.


kontura:
> So far except for plugin* feature files, since these will probably need bigger changes.

If you mean the tests in `plugins-core` those are a separate ticket that I am already working on.

dnf-behave-tests/dnf/metadata.feature
```diff
@@ -82,4 +82,8 @@ Scenario: present user understandable message when there is a mismatch between a
     And I execute "echo \"checksum mismatch\" >> /{context.dnf.repos[simple-base].path}/x86_64/labirinto-1.0-1.fc29.x86_64.rpm"
     When I execute dnf with args "install labirinto"
     Then the exit code is 1
-    And stdout contains "\[MIRROR\] labirinto-1.0-1.fc29.x86_64.rpm: Interrupted by header callback: Inconsistent server data, reported file Content-Length: .*, repository metadata states file length: .* \(please report to repository maintainer\)"
+    And stderr contains "Interrupted by header callback: Inconsistent server data"
```
kontura, line 82: I personally don't like the truncating and I have filed an issue for it: https://github.com/rpm-software-management/dnf5/issues/1829

We can keep the test like this and depending on what is decided in the issue it can be updated later (or not).

kontura (approved):
Thank you!

PR #1585, opened 2024-11-04
Fix downloading multiple packages of the same NEVRA test

m-blaha:
By default re.search() (used by "stdout does not contain") does not match across multiple lines. To bypass this limitation and check that the package name is not present on multiple lines, use "(.|\n)*" pattern instead of ".*".

Before this patch, the step didn't actually test anything and would always pass.

### DIFF
--- a/dnf-behave-tests/dnf/download-binary.feature
+++ b/dnf-behave-tests/dnf/download-binary.feature
@@ -106,9 +106,12 @@ Scenario: Download an existing RPM when there are multiple packages of the same 
     And stderr contains "filesystem-0:3.9-2.fc29.x86_64"
     And stderr contains "wget-0:1.19.5-5.fc29.x86_64"
       # check that each file was being downloaded only once
-    And stderr does not contain "setup-0:2.12.1-1.fc29.noarch.*setup-0:2.12.1-1.fc29.noarch"
-    And stderr does not contain "filesystem-0:3.9-2.fc29.x86_64.*filesystem-0:3.9-2.fc29.x86_64"
-    And stderr does not contain "wget-0:1.19.5-5.fc29.x86_64.*wget-0:1.19.5-5.fc29.x86_64"
+      # By default re.search() (used by "stderr does not contain") does not match
+      # across multiple lines. To bypass this limitation and check that the package
+      # name is not present on multiple lines, use "(.|\n)*" pattern instead of ".*".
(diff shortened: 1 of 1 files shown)


dnf-behave-tests/dnf/download-binary.feature
```diff
@@ -106,9 +106,12 @@ Scenario: Download an existing RPM when there are multiple packages of the same
...
       # check that each file was being downloaded only once
-    And stderr does not contain "setup-0:2.12.1-1.fc29.noarch.*setup-0:2.12.1-1.fc29.noarch"
-    And stderr does not contain "filesystem-0:3.9-2.fc29.x86_64.*filesystem-0:3.9-2.fc29.x86_64"
-    And stderr does not contain "wget-0:1.19.5-5.fc29.x86_64.*wget-0:1.19.5-5.fc29.x86_64"
+      # By default re.search() (used by "stderr does not contain") does not match
+      # across multiple lines. To bypass this limitation and check that the package
+      # name is not present on multiple lines, use "(.|\n)*" pattern instead of ".*".
+    And stderr does not contain "setup-0:2\.12\.1-1\.fc29\.noarch(.|\n)*setup-0:2\.12\.1-1\.fc29\.noarch"
+    And stderr does not contain "filesystem-0:3\.9-2\.fc29\.x86_64(.|\n)*filesystem-0:3\.9-2\.fc29\.x86_64"
+    And stderr does not contain "wget-0:1\.19\.5-5\.fc29\.x86_64(.|\n)*wget-0:1\.19\.5-5\.fc29\.x86_64"
```
kontura, line 114: Nice catch!

Personally I would find it cleaner if we replaced all the `contain` steps with:
```gherkin
And stderr matches line by line
"""
Updating and loading repositories:
 dnf-ci-fedora test repository .*
 dnf-ci-gpg test repository    .*
Repositories loaded.
filesystem-0:3.9-2.fc29.src    .*
filesystem-0:3.9-2.fc29.x86_64 .*
setup-0:2.12.1-1.fc29.noarch   .*
setup-0:2.12.1-1.fc29.src      .*
wget-0:1.19.5-5.fc29.src       .*
wget-0:1.19.5-5.fc29.x86_64    .*
"""
```

But I am also fine with you solution if you prefer.

m-blaha: With download, I'm not sure we can rely on the order of downloaded packages in the output. I believe that's the reason why we have this `stderr contains` / `stderr does not contain` couple of steps.

kontura: I see

PR #1586, opened 2024-11-04
Test that reposync downloads duplicated NEVRA only once

m-blaha:
See https://issues.redhat.com/browse/RHEL-64320
Requires: https://github.com/rpm-software-management/dnf-plugins-core/pull/566


kontura:
Tests pass locally.

PR #1589, opened 2024-11-07
Use dockerfile from the up-to-date dnf-4-stack branch

kontura:
Otherwise it uses dockerfile from the current work directory which is the outdated dockerfile from this branch (`enable-tmt-dnf-4-stack`).

This ensures changes from https://github.com/rpm-software-management/ci-dnf-stack/commit/8532ca86767ed83b08d400e45432faff8d17a9cd take effect.


kontura:
(The failing tests are not really relevant because they use old versions of the tests from `enable-tmt-dnf-4-stack`.)

PR #1590, opened 2024-11-07
Update `Dockerfile.ansible` to fc40 to match CI tests

kontura:
The same builds are used for both CI and ansible tests, their fedora versions have to match to be able to install tested packages.


jan-kolarik (approved):
Thanks!

PR #1595, opened 2024-11-19
Add libdnf5 and libdnf5-cli rpminspect ABI check plan

kontura:
It compares ABI of current PR build with the last nightly build. This means the ABI diff will mostly show only very recent changes.

Since it already uses `rpminspect` we could also enable additional checks if we wanted to.

The plan will be invoked from dnf5 upstream PRs.


kontura:
For: https://github.com/rpm-software-management/dnf5/issues/1651

jan-kolarik (approved):

jan-kolarik:
Thank you! This is much appreciated :slightly_smiling_face:

PR #1601, opened 2024-12-05
Use "OpenPGP" instead of "PGP"

evan-goode:
For https://github.com/rpm-software-management/dnf5/pull/1931.


kontura (commented):
I have noticed there still are some occurrences of only `PGP` in `dnf-behave-tests/dnf/distro-sync.feature` and `dnf-behave-tests/dnf/gpg.feature`.

It seems the steps `stdout/stderr contains lines matching` only checks if there is a line that contains the string. 
Based on the name of the step it would make more sense to me if it checked that the whole line matches.

kontura:
Lets merge this now to fix the CI, we can address my previous comment in a follow up PR if desired.

PR #1608, opened 2024-12-11
Use "OpenPGP" instead of "PGP" in distro-sync, gpg

evan-goode:
Following up from this comment: https://github.com/rpm-software-management/ci-dnf-stack/pull/1601#pullrequestreview-2483708419.

These lines do use `.*` in some places, so instead of converting them to `stderr contains lines`, I added `^` and `$` to further constrain the regex.


kontura:
I was thinking if the steps `stderr/stdout contains lines matching` should add the `^` and `$` automatically for all lines?

What do you think?

evan-goode:
Sounds good, tests are passing locally with that change.

kontura (approved):
Great, thank you!

PR #1612, opened 2024-12-13
Disable new dnf-automatic test on Fedora >= 41

kontura:
While merging https://github.com/rpm-software-management/ci-dnf-stack/pull/1610 I **again** forgot that the dnf-automatic tests can't run on fedora >= 41.


evan-goode (approved):
LGTM

PR #1251, opened 2023-04-20
Test for dnf-automatic reporting a transaction error

jan-kolarik:
Requires https://github.com/rpm-software-management/dnf/pull/1923.

PR #1614, opened 2024-12-18
automatic: Tests for error reporting and no updates functionality

jan-kolarik:
- Porting https://github.com/rpm-software-management/ci-dnf-stack/pull/1603 to dnf5.
- Porting https://github.com/rpm-software-management/ci-dnf-stack/pull/1251 to dnf5.
- Covering CI tests for the no updates functionality which requires: https://github.com/rpm-software-management/dnf5/pull/1884.


kontura (approved):
I wouldn't mind comparing the full `stdout` to make sure it has the expected form and there are not extras but that could be a separate test.

PR #1618, opened 2025-01-15
Add test for --transient and persistence option

evan-goode:
The `--transient` flag only works on a bootc system, but we should test that it throws the correct error and doesn't do anything unexpected on non-bootc systems.

Requires https://github.com/rpm-software-management/dnf/pull/2180 and https://github.com/rpm-software-management/libdnf/pull/1683.


dcantrell (approved):

dcantrell:
Looks good to merge.

PR #1619, opened 2025-01-15
Merge `bootc` branch to dnf-4-stack

evan-goode:
The `--transient` flag only works on a bootc system, and our current CI infrastructure is not yet set up to test on booted bootc systems. Still, we should add a test here to check that it throws the correct error and doesn't do anything unexpected on non-bootc systems.

Requires https://github.com/rpm-software-management/dnf/pull/2180 and https://github.com/rpm-software-management/libdnf/pull/1683.


jrohel:
Thank you. The tests are straightforward.

PR #1620, opened 2025-01-20
versionlock: add a test when `add`ing packages by wildcard

kontura:
Requires: https://github.com/rpm-software-management/dnf5/pull/2009


m-blaha (approved):
Thank you!

PR #1359, opened 2023-08-03
Add test for releasever_major, releasever_minor

evan-goode:
Requires https://github.com/rpm-software-management/dnf5/pull/800

For https://github.com/rpm-software-management/dnf5/issues/710

PR #1621, opened 2025-01-27
Add releasever_{major,minor} tests

evan-goode:
Backports https://github.com/rpm-software-management/ci-dnf-stack/pull/1359 to dnf-4-stack.

Adds a test for the new `--releasever-major` and `--releasever-minor` options.

Requires https://github.com/rpm-software-management/dnf/pull/2198 and https://github.com/rpm-software-management/libdnf/pull/1689.

For https://issues.redhat.com/browse/RHEL-68034.

### DIFF
--- a/dnf-behave-tests/dnf/vars-releasever.feature
+++ b/dnf-behave-tests/dnf/vars-releasever.feature
@@ -103,13 +103,36 @@ Scenario: Releasever is substituted in baseurl via vars in custom location
         | install       | setup-0:2.12.1-1.fc29.noarch  |
 
 
-@dnf5
 Scenario: Releasever gets split into releasever_major and releasever_minor
   Given I copy directory "{context.scenario.repos_location}/dnf-ci-fedora" to "/temp-repos/base-epel-12-34"
     And I use repository "dnf-ci-fedora" with configuration
         | key         | value |
         | baseurl     | file://{context.dnf.installroot}/temp-repos/base-epel-$releasever_major-$releasever_minor |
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/dnf/vars-releasever.feature
```diff
@@ -101,3 +101,40 @@ Scenario: Releasever is substituted in baseurl via vars in custom location
     And Transaction is following
         | Action        | Package                       |
         | install       | setup-0:2.12.1-1.fc29.noarch  |
+
+
+@dnf5
```
kontura, line 106: I think the `@dnf5` is not needed here.

kontura (commented):
I think it would be good to also test that `$releasever` is not affected by setting major/minor.

Otherwise it looks good to me.

kontura:
I was looking into it a bit more and I realized it would probably be good to have a test where the minor/major values are overridden by package provides since that is it intended use-case.
It also uses a different code path in both libdnf/dnf.

dnf-behave-tests/dnf/vars-releasever.feature
evan-goode: Removed

evan-goode:
> I think it would be good to also test that `$releasever` is not affected by setting major/minor.

> I was looking into it a bit more and I realized it would probably be good to have a test where the minor/major values are overridden by package provides since that is the intended use-case.

Both done, good suggestions.

kontura (approved):

PR #1626, opened 2025-02-07
Add tests for advisory info --json

kontura:
For: https://github.com/rpm-software-management/dnf5/pull/1970

I am using the `updateinfo` alias so the tests can be easily ported to dnf4.


m-blaha (approved):
Thank you!

PR #1630, opened 2025-02-11
Install systemd package early

m-blaha:
Since version 42 Fedora comes with pre-installed
systemd-standalone-sysusers package which conflicts with systemd
package which prevents dnf5daemon-server from being installed.

See errors during the container build here: https://artifacts.dev.testing-farm.io/8d2a79aa-95f5-41b4-b09c-42e6f65a5ede/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,9 @@ COPY ./repos.d/ /etc/yum.repos.d/
 # enable dnf5
 RUN set -x && \
     dnf -y --refresh upgrade; \
+    # Since Fedora 42, `systemd-standalone-sysuser` comes pre-installed.  This package 
+    # conflicts with `systemd`, preventing `dnf5daemon-server` from being installed.
+    dnf -y install systemd --allowerasing; \
     dnf -y install dnf5 dnf5-plugins; \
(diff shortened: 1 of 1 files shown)


jrohel:
I reproduced the issue locally. And this PR (adding `--allowerasing` to `dnf5 -y builddep ...`) fixed the problem.

There is another approach. To explicitly install `systemd` as soon as possible (right after the upgrade). But what's better...
```
# enable dnf5
RUN set -x && \
    dnf -y --refresh upgrade; \
    dnf -y install systemd --allowerasing; \
    dnf -y install dnf5 dnf5-plugins; \
    ...
```

kontura:
I think it can fail in either of the `builddep` step or the `dnf5 -y install $COPR_RPMS` I mentioned.
When the `COPR` and `COPR_RPMS` environment variables are set (like in the linked log) it fails there, if they are not set it fails in the `builddep` step.

I would probably explicitly install `systemd` as soon as possible to prevent these issues. It would also be good to add a comment explaining why its done.

m-blaha:
Good points guys, I've changed the fix to the proposed solution.

kontura (approved):

jrohel:
Thanks.

PR #1642, opened 2025-02-27
Enable "Missing baseurl/metalink/mirrorlist" test

evan-goode:
Requires https://github.com/rpm-software-management/dnf5/pull/2094.


kontura:
The failure is unrelated.
The test passes locally, thanks!

PR #1643, opened 2025-03-04
Test group and environment excludes

pkratoch:
Tests for https://github.com/rpm-software-management/dnf5/pull/2107

### DIFF
--- a/dnf-behave-tests/dnf/comps-excludes.feature
+++ b/dnf-behave-tests/dnf/comps-excludes.feature
@@ -1,0 +1,150 @@
+Feature: Group and environment upgrade
+
+# comps-upgrade-1 repo:
+#   a-group:
+#     mandatory: A-mandatory
+#     default: A-default
+#     optional: A-optional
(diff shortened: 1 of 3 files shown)


evan-goode (changes requested):
Nice, these tests are thorough. For better testing the specific use case in https://github.com/rpm-software-management/dnf5/issues/1910, could we add a test for installing an environment of two or more groups where one of the groups is excluded? This tests installing `AB-environment` when `a-group` is excluded but `a-group` is the only group in `AB-environment`. We may have to add another environment to the testing comps.xml.

pkratoch:
I added "Install an environment when some of its groups are excluded" scenario and modified the scenario with all the excluded groups, so that the difference is more obvious. As a bonus, now there is also `excludegroups` option with multiple values, which wasn't present anywhere before.

evan-goode (approved):
Thanks!

jan-kolarik:
There are some failures, but all is related to the newly added expired-pgp-keys functionality.

PR #1644, opened 2025-03-05
Tests for correct from_repo after offline transactions

m-blaha:
Requires: https://github.com/rpm-software-management/dnf5/pull/2093

### DIFF
--- a/dnf-behave-tests/dnf/offline.feature
+++ b/dnf-behave-tests/dnf/offline.feature
@@ -257,3 +257,16 @@ Given I use repository "simple-base"
   And Transaction is following
       | Action        | Package                       |
       | install       | labirinto-0:1.0-1.fc29.x86_64 |
+
+
+# https://github.com/rpm-software-management/dnf5/issues/1851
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/dnf/offline.feature
```diff
@@ -257,3 +257,16 @@ Given I use repository "simple-base"
   And Transaction is following
       | Action        | Package                       |
       | install       | labirinto-0:1.0-1.fc29.x86_64 |
+
+
+# https://github.com/rpm-software-management/dnf5/issues/1851
+Scenario: Test offline install and offline status
```
kontura, line 263: I think the scenario name is a left over from the copy.

kontura (changes requested):

dnf-behave-tests/dnf/offline.feature
m-blaha: Shame on me! Fixed.

kontura:
Thanks!

PR #1647, opened 2025-03-07
Test history info range

pkratoch:


kontura (approved):
Thanks, it would be great to have the test also for dnf5 to see how it handles such a situation.

kontura:
> > Thanks, it would be great to have the test also for dnf5 to see how it handles such a situation.
> 
> Such test doesn't make sense for dnf5, since it no longer merges the transactions. It's intentional, documented here: https://github.com/rpm-software-management/dnf5/blob/main/doc/changes_from_dnf4.7.rst?plain=1#L220

Oh, I forgot about that, thanks!

PR #1651, opened 2025-03-14
expired-pgp-keys: Change spelling from PGP to OpenPGP

ppisar:
I corrected a spelling of an DNF5 message in
<https://github.com/rpm-software-management/dnf5/pull/2135>.

This patch supports both spellings.


m-blaha (approved):
Thank you!

PR #1656, opened 2025-03-31
Update test outputs for parallel downloading API

kontura:
For: https://github.com/rpm-software-management/dnf5/pull/2167

### DIFF
--- a/dnf-behave-tests/dnf/repo-sync.feature
+++ b/dnf-behave-tests/dnf/repo-sync.feature
@@ -5,7 +5,7 @@ Background: Force column width
 # Some of the curl errors can be quite long and since they are
 # truncated: https://github.com/rpm-software-management/dnf5/issues/1829
 # we need to force the width to see them in full.
-Given I set environment variable "FORCE_COLUMNS" to "300"
+Given I set environment variable "FORCE_COLUMNS" to "400"
 
 
 @bz1763663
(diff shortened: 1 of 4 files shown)


dnf-behave-tests/dnf/repo-sync.feature
```diff
@@ -220,3 +217,258 @@ Scenario: Mirrorlist with invalid mirrors and one good mirror
...
+
+Scenario Outline: working repo and repo with invalid baseurl is an error when skip_if_unavailable is <value>
+  Given I use repository "dnf-ci-fedora"
+    And I configure a new repository "testrepo" with
+        | key     | value   |
+        | enabled | True    |
+        # the baseurl is a local file but doesn't exist
+        | baseurl | invalid |
+   When I execute dnf with args "rq dwm.x86_64 --setopt=skip_if_unavailable=<value>"
+   Then the exit code is 1
```
kontura, line 463: There is a change of exit code, when `skip_if_unavailable` is `1` this used to exit with 0.
This is related to https://github.com/rpm-software-management/librepo/pull/336#issuecomment-2872054045.

I don't have a strong opinion here.

m-blaha: I'm thinking about use-case where the user have configured some "local" repos from a NFS drive accessible from the office, but inaccessible from home. I would expect dnf to print a librepo error here, but with `skip_if_unavailable=1` continue with a limited set of repositories. But such user can still use a valid, even though non-existing path like "baseurl=file:///nfsmount/repo". Does it make sense to add a test for such use-case?

m-blaha: Hm, maybe one more thing - the error message should contain the invalid repository id. Currently it's unnecessarily hard to find the correct place to fix the configuration.

m-blaha: But feel free to file a new ticket for enhancing the error message.

kontura: > I'm thinking about use-case where the user have configured some "local" repos from a NFS drive accessible from the office, but inaccessible from home. I would expect dnf to print a librepo error here, but with `skip_if_unavailable=1` continue with a limited set of repositories. But such user can still use a valid, even though non-existing path like "baseurl=file:///nfsmount/repo". Does it make sense to add a test for such use-case?

I think that is a valid use-case, I have added two new tests for this. :+1: 

I have also updated the tests from the previous commit, the order in which the repositories are printed to `stderr` can change so I had to replace some `matches line by line` steps with `stderr contains lines matching`.



> Hm, maybe one more thing - the error message should contain the invalid repository id. Currently it's unnecessarily hard to find the correct place to fix the configuration.

That is a very good point. There is a general problem where the librepo error messages are meant really only for single repo download. When an error that ends all the downloads occurs during parallel repo download librepo doesn't propagate the information which repo caused it.

For the specific `Librepo error: Empty mirrorlist and no basepath specified!` I can see couple more issues:
1. We should change `basepath` to `baseurl` 
2. The error can happen when `baseurl` is set to relative path which doesn't exists so it actually is `specified`. I think this be even more confusing.

I will create a new ticket for these once the dnf5 PR is merged.

kontura:
@m-blaha I had to update one failing test due to different libcurl versions but it should be good to go now.

m-blaha (approved):
Thank you!

PR #1659, opened 2025-04-03
Enable test for length of history list output outside of terminal

pkratoch:
Test for https://github.com/rpm-software-management/dnf5/pull/2162

### DIFF
--- a/dnf-behave-tests/dnf/history-list.feature
+++ b/dnf-behave-tests/dnf/history-list.feature
@@ -207,19 +207,17 @@ Scenario: history lame (no transaction with such package)
   And stdout is empty
 
 
-@xfail
 # Reported as https://github.com/rpm-software-management/dnf5/issues/2025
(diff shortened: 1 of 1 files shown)


kontura:
I am not exactly sure what is causing it but the test is failing for me on fedora > 40 because there is one extra space.

I think it is caused by an outside change. I tried downgrading behave (and everything Python) to the f40 versions and then it passed again.

pkratoch:
The table width is for some reason always 0-2 characters shorter. For example, when `FORCE_COLUMNS` is set to 78, 79 or 80, it will be 78, when it's set to 75, 76 or 77, it will be 75. I think it's related to how the widths of the columns and paddings are calculated internally and I haven't found a way how to make the width exact, although I don't think it matters much. We can at least test that the width doesn't exceed the 80 characters. Not sure how to make it better. :/

PR #1661, opened 2025-04-04
Configure installed group/environment manually

kontura:
This is needed because daemon doesn't implement install for comps.

For https://github.com/rpm-software-management/dnf5/issues/2176

### DIFF
--- a/dnf-behave-tests/dnf/comps-environment-list.feature
+++ b/dnf-behave-tests/dnf/comps-environment-list.feature
@@ -4,8 +4,13 @@ Feature: Tests for environment list and info commands
 
 Background: Enable repo and mark an environment as installed
  Given I use repository "comps-group-list"
-   And I successfully execute dnf with args "group install test-environment"
+   And I create file "/usr/lib/sysimage/libdnf5/environments.toml" with
+       """
+       version = "1.0"
+       [environments]
 
+       test-environment = {groups = [], userinstalled = true}
+       """
 
 # https://github.com/rpm-software-management/dnf5/issues/1502
 @xfail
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/dnf/comps-environment-list.feature
```diff
@@ -1,11 +1,15 @@
-@dnf5daemon
 Feature: Tests for environment list and info commands
 
 
 Background: Enable repo and mark an environment as installed
  Given I use repository "comps-group-list"
-   And I successfully execute dnf with args "group install test-environment"
+   And I create file "/usr/lib/sysimage/libdnf5/environments.toml" with
```
pkratoch, line 6: Can you please add a comment here explaining that the config change is needed instead of calling `group install` because of the daemon? The same for the environments.

pkratoch (commented):
Thank you!

Let me know when the PR for fixing the display order for the daemon is ready.
If it should take more time, maybe we could temporarily remove the `dnf5daemon` tag also from the comps-group-list.feature and merge it and add it back when the PR for daemon is ready. But if it's a simple change, we can wait with this.

dnf-behave-tests/dnf/comps-environment-list.feature
kontura: Done.

kontura:
> Let me know when the PR for fixing the display order for the daemon is ready. If it should take more time, maybe we could temporarily remove the `dnf5daemon` tag also from the comps-group-list.feature and merge it and add it back when the PR for daemon is ready. But if it's a simple change, we can wait with this.

The PR is here: https://github.com/rpm-software-management/dnf5/pull/2179
It got more complicated than I initially thought, if there will be problems with it I will create another PR to temporarily disable the tests.

kontura:
@pkratoch this should be now ready to merge.

PR #1662, opened 2025-04-04
Change expected output of the clean command to also work for dnf5daemon

pkratoch:
The output for the daemon is different from dnf5. I think that duplicating the tests to customize them for each command is not necessary when the only difference is this output.

### DIFF
--- a/dnf-behave-tests/dnf/clean-cachefiles.feature
+++ b/dnf-behave-tests/dnf/clean-cachefiles.feature
@@ -30,7 +30,7 @@ Scenario: Cleanup of the whole cache (dnf clean all)
    Then the exit code is 0
     And stdout matches line by line
         """
-        Removed 4 files, 4 directories \(total of [0-9]* KiB\). 0 errors occurred.
+        (Removed 4 files, 4 directories \(total of [0-9]* KiB\). 0 errors occurred.|Cache successfully cleaned.)
         """
    When I execute "find | sort" in "{context.dnf.installroot}/var/cache/dnf"
    Then stdout matches line by line
@@ -57,7 +57,7 @@ Scenario: Cleanup of the whole cache, cache contains link pointing outside, cont
    Then the exit code is 0
     And stdout matches line by line
         """
-        Removed 6 files, 4 directories \(total of [0-9]* KiB\). 0 errors occurred.
+        (Removed 6 files, 4 directories \(total of [0-9]* KiB\). 0 errors occurred.|Cache successfully cleaned.)
         """
     And directory "/var/cache/dnf" is empty
     And file "/tmp/user_dir/user_file1" exists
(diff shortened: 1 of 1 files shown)


pkratoch:
I just thought of another possibility for the output checking - to introduce a separate steps to check dnf5 output vs. dnf5daemon output. This might come in handy if there are more differences like this one. I will look into it, marking it as draft for now.

pkratoch:
I added steps to check specifically dnf5 and dnf5daemon outputs.

dnf-behave-tests/dnf/steps/cmd.py
```diff
@@ -315,6 +316,20 @@ def then_stderr_contains_lines_matching(context):
             raise AssertionError("Stderr doesn't contain line matching: %s" % line)
 
 
+@behave.then("dnf5 stdout matches line by line")
+def then_dnf5_stdout_matches(context):
+    if context.dnf5daemon_mode:
+        return
```
kontura, line 322: To be honest I was happy we got rid of these steps that do nothing when the mode doesn't match.
I would like something like this more:
```gherkin
And stdout matches line by line
    """
    Removed 1 files, 1 directories \(total of [0-9]* KiB\). 0 errors occurred.
    """
    or in dnf5daemon mode
    """ 
    Cache successfully cleaned.
    """
```
However that doesn't seem to be possible.

The closest thing seems to be:
```gherkin
And stdout matches line by line depending on mode
    """
    Removed 1 files, 1 directories \(total of [0-9]* KiB\). 0 errors occurred.
    ---
    Cache successfully cleaned.
    """
```
Which is not great.

kontura: Perhaps for single line outputs we could do:
```gherkin
And stdout matches depending on mode
    | mode        | line                                                                       |
    | default     | Removed 1 files, 1 directories \(total of [0-9]* KiB\). 0 errors occurred. |
    | dnf5daemon  | Cache successfully cleaned.                                                |
```
But I suppose I could file a separate issue for this if you prefer since we should focus on fixing the CI now.

pkratoch: The second example (stdout matches line by line depending on mode) would be doable. The third I don't like very much because of limited usability.

I could also return to my original approach, which was to test for:
`(Removed 1 files, 1 directories \(total of [0-9]* KiB\). 0 errors occurred.|Cache successfully cleaned.)`
Which is similar to the "stdout matches line by line depending on mode" except for not defining which mode should match which. But the probability of a bug that would completely change the message and wouldn't be intentional seems quite low to me.

kontura: > I could also return to my original approach, which was to test for:
> `(Removed 1 files, 1 directories \(total of [0-9]* KiB\). 0 errors occurred.|Cache successfully cleaned.)`
> Which is similar to the "stdout matches line by line depending on mode" except for not defining which mode should match which. But the probability of a bug that would completely change the message and wouldn't be intentional seems quite low to me.

Agreed, for me it would be the preferred solution.

pkratoch: Ok, I changed it.

PR #1664, opened 2025-04-07
Test that user is warned about inaccessible module config

martinobanana:
Issue https://issues.redhat.com/browse/RHEL-62833

Requires: https://github.com/rpm-software-management/libdnf/pull/1695


martinobanana:
Failed checks expected (improper distro)

PR #1667, opened 2025-04-10
search: Separate a package from a summary with a tabulation character

ppisar:
For: https://github.com/rpm-software-management/dnf5/pull/2190
Resolves: https://github.com/rpm-software-management/dnf5/issues/2166

I did tested it yet. I will need to build RPMs and install them into a container and that will take some time. I cannot reuse builds from builds from code PR because somebody restarted the  PR tests and that deleted the old builds.

### DIFF
--- a/dnf-behave-tests/dnf/search-sort.feature
+++ b/dnf-behave-tests/dnf/search-sort.feature
@@ -16,33 +16,33 @@ Scenario: sort alphanumerically
    And stdout is
        """
        Matched fields: name (exact)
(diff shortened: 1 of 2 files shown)


ppisar:
According to local tests "\t" escape sequence is not expaned by behave. It seems I will need to use literal tabulation charater in the tests.

ppisar:
I replaced escape sequences with literal characters. Bot test features pass for me with the PR's code.

Issue #1666, opened 2024-10-18
Testing `dnf4 --transient`


evan-goode:
The DNF 4 bootc plugin needs integration tests, and [ci-dnf-stack](https://github.com/rpm-software-management/ci-dnf-stack) is not equipped to test things meant for OSTree systems. Some unanswered questions are:

- Where should test source code live? IMO, as much as possible should be in this upstream repository.
- What testing framework should we use?
- Where should the tests run (GHA, CoreOS Jenkins, Testing Farm)?
- Should we write the tests from scratch, or base them on existing rpm-ostree and/or bootc tests?
- Will any of the answers to the above need to change as development of the bootc plugin progresses?

evan-goode:
For now, let's follow what bootc has done upstream [1] and use TMT/Testing Farm with `bootc install-to-existing-root`.

[1] https://github.com/bootc-dev/bootc/pull/1249.

ppisar:
RHEL QE people already have some tests. Problem with Testing Farm is that it still does not support installing custom packages (i.e. a dnf build to test) into a bootc system.

cgwalters:
This is a hugely complex topic. One thing I personally find really painful about Testing Farm is that it doesn't by default at a low level solve this problem of injecting code-under-test; it's the responsibility of higher level driving tools to do so. For example https://packit.dev does this.

The way I'd personally much prefer things to work relates to https://issues.redhat.com/browse/TFT-2751 - in that flow, one would build a derived bootc container image with the content-under-test (e.g. a dockerfile that's just `FROM <base> RUN dnf -y upgrade dnf` basically) and then provision it using the to-existing-root flow above.

However, that from-a-registry path does complicate local testing; there's some inherent friction and duplication in requiring pushing containers-under-test to a local registry (space duplication and TLS/networking considerations).

> Is this the reason why we can't use the [bootc TMT provisioning plugin](https://tmt.readthedocs.io/en/latest/plugins/provision.html#bootc)?

I think you (or anyone) could, but it does require nested virt today. That path makes sense in some cases, but ultimately the path I want to streamline more looks closer to https://github.com/virt-manager/virt-manager/issues/739

PR #1670, opened 2025-04-21
Testing framework for bootc using TMT

evan-goode:
For https://github.com/rpm-software-management/ci-dnf-stack/issues/1666.

This PR implements a testing framework for bootc systems, based on the structure of bootc's own upstream tests (https://github.com/bootc-dev/bootc/pull/1249). We'll be able to write tests in ci-dnf-stack that run on booted bootc systems and can perform tests across multiple reboots, which will be useful for testing the `--transient` and `--persist` features.

We use `bootc install to-existing-root` and reboot the test VM directly into the testing container to avoid needing nested virtualization. Packit is used to trigger the tests on pull requests to ci-dnf-stack. Changes to https://github.com/rpm-software-management/dnf/blob/master/.packit.yaml and https://github.com/rpm-software-management/libdnf/blob/dnf-4-master/.packit.yaml will be needed to trigger the tests on dnf and libdnf PRs.

Currently, there is a simple test for `dnf4 --transient`.


evan-goode:
Note for reviewers: you can run this locally with

```
sudo dnf install -y tmt+all
sudo tmt run --debug
```

dnf-behave-tests/dnf/bootc.feature
```diff
@@ -0,0 +1,49 @@
...
+   And file "/usr/bin/hello" exists
+
+@reboot_count_1
+Scenario: After using DNF --transient, the overlay should exist
+  When I successfully execute "ostree admin status"
+  Then stdout contains lines
+     """
+     Unlocked: transient
+     """
+   And path "/usr/bin" is not writeable
```
kontura, line 39: The `Scenario: After using DNF --transient, the overlay should exist` depends on `Scenario: Install a package using --transient on a bootc system` right?

While these two perhaps could be combined I believe the next one with `@reboot_count_2` also depends on them and has to be separate.

Could we describe this somewhere? Since we usually try to have the scenarios completely separate.

tmt/plans/bootc-behave-dnf.fmf
```diff
@@ -0,0 +1,24 @@
...
+  package:
+  - podman
+  - skopeo
+  - jq
+
+execute:
+  how: tmt
+
+/test-01-bootc:
+  summary: Foo test
```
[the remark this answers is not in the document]
evan-goode: I added a better summary and changed this section to only test the "transient" feature since we'll want separate feature files (and separate test-00-bootc-install steps) for each feature we want to test. See how bootc does it: https://github.com/bootc-dev/bootc/blob/main/tmt/plans/integration.fmf.

I added some docs to `dnf-behave-tests/README.md`.

dnf-behave-tests/dnf/bootc.feature
evan-goode: Yes, I added some explanation to `dnf-behave-tests/README.md` that scenarios in the `dnf-bootc` suite are not independent.

I'm not sure how we would cleanly combine the two `@reboot_count_1` scenarios into one.

evan-goode:
> Is there a plan to someday run normal dnf tests in the bootc setup? I don't really understand this fully but it seems to me like the bootc tests will always be separate and quite different to normal dnf tests. If that is the case I would prefer if they were in a different test suite (the way `createrepo_c` is separated from dnf).

I fully agree, great idea. Even if we want to run some of the `@no_installroot` normal DNF tests on Image Mode in the future, it's still best to keep separate the tests that _need_ a bootc system and `tmt-reboot`.

I created a `dnf-bootc` test suite that shares the steps directory and `environment.py` with the `dnf` test suite via symlinks. 

> With that approach we wouldn't need to always disable tests with `@bootc` tag and the normal `./container-test run` would continue to work by default.

Yes, this is much better.

kontura (approved):
Thanks, this look good.
Like we discussed it would be ideal if the rebooting was just a behave step but I am also not sure its possible at this point.

I am also a little bit on the fence regarding sharing the steps via a symlink mostly because we already have a way to do that using the `common` directory. However since there are no `bootc` suite specific steps and we share the `environment.py` as well I suppose it is fine. We can always change it later.

PR #1671, opened 2025-04-22
Adjust to new exit codes for non-existing advisories

m-blaha:
Requires: https://github.com/rpm-software-management/dnf5/pull/2201


kontura:
Tests pass locally.

PR #1674, opened 2025-05-14
Include dnf-nightly in ci-host builds

kontura:
This can sometimes be useful when one package (libdnf5) depends on a new unreleased release of another (librepo) during our ccache builds.


m-blaha (approved):
Thank you!

PR #1675, opened 2025-05-19
Add additional `--transient` tests

kontura:


dnf-behave-tests/dnf-bootc/transient.feature
```diff
@@ -46,3 +44,62 @@ Scenario: After reboot, the transient overlay should disappear
...
+  Then the exit code is 0
+   And Transaction is following
+     | Action | Package                   |
+     | remove | hello-0:1.0-1.fc29.x86_64 |
+   And file "/usr/bin/hello" does not exist
+   And stdout does not contain "--transient"
+   And stderr is empty
+
+@reboot_count_2
+Scenario: dnf install --transient on unlocked system doesn't warn about transient overlay
```
evan-goode, line 78: Could this scenario be copied to `@reboot_count_3` as well, after what is currently the last test? This scenario is testing "If `ostree admin status` is already `Unlocked: development`, then `dnf4 --transient` should not print the 'A transient overlay will be created...' message". We should also test, "if `ostree admin status` is already "Unlocked: transient`, then the message is not printed".

`Unlocked: transient` is distinct from `Unlocked: development`. In "transient", the /usr overlay is read-only, in "development", it's writable.

Come to think of it, we should probably test each permutation:

On bootc system:

| DNF persistence setting/ostree admin unlock | Locked | Unlocked: transient | Unlocked: development |
| ------------- | ------------- | ------------|--------|
| auto  | fail | fail | succeed, no message | 
| transient  | succeed with message | succeed, no message | succeed, no message |
| persist (for now, probably just needs a single test) | fail | fail | fail |

evan-goode (changes requested):
Nice, thanks. I agree about creating a separate `environment.py` for the bootc tests.

dnf-behave-tests/dnf-bootc/transient.feature
kontura: Those are good points.
I have added the requested new tests and changed the wording a bit to make it clear what overlay is used at what stage.

evan-goode (approved):
Awesome, thank you!

evan-goode:
Let me just double-check that Packit has tested the most recent commit...

PR #1695, opened 2025-07-07
dnf5: Add basic "--from-repo" tests

jrohel:
requires: https://github.com/rpm-software-management/dnf5/pull/2303


evan-goode (approved):
Excellent, passes with https://github.com/rpm-software-management/dnf5/pull/2303. Thanks!

PR #1699, opened 2025-07-09
groups: specify display_order to match the expected order in tests

kontura:
Without this the order is implementation dependent.

This is a follow up to: https://github.com/rpm-software-management/ci-dnf-stack/pull/1696
It is required for: https://github.com/rpm-software-management/dnf5/pull/2345


ppisar (approved):
Good.

PR #1709, opened 2025-07-24
Add test to make sure we can run with fastestmirror

kontura:
The fastestmirror functionality is not actually checked because that is harder to test (it is partly non-deterministic).


kontura:
Unfortunately it doesn't always fail with the old build.
I believe its because accessing the freed memory is undefined behavior so sometimes it can work.

ppisar (approved):
Nice. With the old code (current nightly build) the new test sometimes reports segfault (-11).

PR #1710, opened 2025-07-28
bootc: enable tag-repository and test-artifacts

janblazek:
We need to get the packages out of the system under test into the container, where ci-dnf-stack is running. This change will allow us to enable ci-dnf-stack/bootc tests in gating.

### DIFF
--- a/tmt/tests/test-00-bootc-install.sh
+++ b/tmt/tests/test-00-bootc-install.sh
@@ -36,6 +36,25 @@ if [ "$TMT_REBOOT_COUNT" -eq 0 ]; then
     ls -al "$BOOTC_TEMPDIR"
 
     pushd "$TMT_TREE"
+        if [ -e /etc/yum.repos.d/tag-repository.repo ]; then
+            cp -v /etc/yum.repos.d/tag-repository.repo repos.d/
+        fi
+
+        if [ -e /etc/yum.repos.d/test-artifacts.repo ]; then
+            mkdir -p rpms
+            dnf repoquery --repo=test-artifacts --qf "%{SOURCERPM} %{NAME}" > T-A-pkgs.pkglist
+            cat T-A-pkgs.pkglist
+            set +e # following grep would fail whole script on first non-match
+            for P in dnf libdnf libsolv librepo librhsm libmodulemd1 libmodulemd microdnf createrepo_c; do
+                grep ^$P T-A-pkgs.pkglist | while read line ; do
+                    echo $line | awk '{print $2}' >> T-A-Packages
+                done
+            done
+            set -e
+            cat T-A-Packages
+            dnf download --destdir rpms/ --disablerepo=\* --enablerepo=test-artifacts `cat T-A-Packages | xargs echo -n`
+        fi
+
         ./container-test \
         --container localhost/dnf-bot/dnf-testing-bootc:latest \
         build \


tmt/tests/test-00-bootc-install.sh
```diff
@@ -36,6 +36,23 @@ if [ "$TMT_REBOOT_COUNT" -eq 0 ]; then
     ls -al "$BOOTC_TEMPDIR"
 
     pushd "$TMT_TREE"
+        find /etc/yum.repos.d/*.repo -name tag-repository.repo | xargs -I '{}' cp -v '{}' repos.d/
```
ppisar, line 39: Did you mean:

find /etc/yum.repos.d -name tag-repository.repo -exec cp -v '{}' repos.d/ \;

janblazek: I agree that /etc/yum.repos.d/*.repo is overhead, so I have changed it to /etc/yum.repos.d/
The find -exec didn't work for me in Testing Farm in container, so I would rather keep xargs there.

ppisar: I see. You are trying to workaround "set -e". find returns 1 if it cannot find anything.

ppisar: Do you want to copy /etc/yum.repos.d/tag-repository.repo, or any tag-repository.repo file in tree of /etc/yum.repos.d? I guess the first. What about:

~~~
if test -e /etc/yum.repos.d/tag-repository.repo; then
  cp -v /etc/yum.repos.d/tag-repository.repo repos.d
fi
~~~

tmt/tests/test-00-bootc-install.sh
```diff
@@ -36,6 +36,23 @@ if [ "$TMT_REBOOT_COUNT" -eq 0 ]; then
...
+
+        if [ -e /etc/yum.repos.d/test-artifacts.repo ]; then
+            mkdir -p rpms
+            repoquery --disablerepo=\* --enablerepo=test-artifacts --qf "%{SOURCERPM} %{NAME}" > T-A-pkgs.pkglist
```
ppisar, line 43: Please write it as <code>dnf repoquery --repo=test-artifacts</code>:

Bare repoquery program won't work with dnf5.
The --disablerepo --enablerepo can be simplified to --repo.

ppisar (commented):
Please squash the commits and document why we need this change.

tmt/tests/test-00-bootc-install.sh
janblazek: You're right, this improves readability.

tmt/tests/test-00-bootc-install.sh
```diff
@@ -36,6 +36,25 @@ if [ "$TMT_REBOOT_COUNT" -eq 0 ]; then
     ls -al "$BOOTC_TEMPDIR"
 
     pushd "$TMT_TREE"
+        if [ -e /etc/yum.repos./tag-repository.repo ]; then
```
ppisar, line 39: You are missing "d" in "yum.repos.d" directory name.

tmt/tests/test-00-bootc-install.sh
```diff
@@ -36,6 +36,23 @@ if [ "$TMT_REBOOT_COUNT" -eq 0 ]; then
...
     pushd "$TMT_TREE"
+        find /etc/yum.repos.d/*.repo -name tag-repository.repo | xargs -I '{}' cp -v '{}' repos.d/
+
+        if [ -e /etc/yum.repos.d/test-artifacts.repo ]; then
+            mkdir -p rpms
+            repoquery --disablerepo=\* --enablerepo=test-artifacts --qf "%{SOURCERPM} %{NAME}" > T-A-pkgs.pkglist
+            cat T-A-pkgs.pkglist
+            set +e # following grep would fail whole script on first non-match
+            for P in dnf dnf-plugins-core libdnf libsolv librepo librhsm libmodulemd1 libmodulemd microdnf createrepo_c; do
+                grep ^$P T-A-pkgs.pkglist | while read line ; do
```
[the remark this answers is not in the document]
ppisar: How was this resolved? You could use <code>grep "^${P} ". But I think more understandable would be:

~~~~
<T-A-pkgs.pkglist while read SRPM BRPM; do
  if [ "$P" = "$SRPM" ]; then
    echo "$BRPM" >> T-A-Packages
  fi
done
~~~~

The you could also remove "set +e".

tmt/tests/test-00-bootc-install.sh
janblazek: Thank you! <facepalm>

PR #1712, opened 2025-08-06
Add basic test for python_plugins_loader

kontura:
Requires: https://github.com/rpm-software-management/dnf5/pull/2378


pkratoch:
Sorry, this slipped under my radar. The test fails unless I add step `I enable plugin "test_plugin"`. I guess that's because all plugins are disabled by default and it probably passed before the checking for enabled plugins was fixed. Otherwise, it looks good.

PR #1713, opened 2025-08-07
store-option.feature: packages are not removed by following transaction

kontura:
Requires: https://github.com/rpm-software-management/dnf5/pull/2380


dnf-behave-tests/dnf/transaction-sr/store-option.feature
```diff
@@ -287,3 +287,15 @@ Given I execute dnf with args "install @test-group"
         """
         test-group.xml
         """
+
+
+# https://github.com/rpm-software-management/dnf5/issues/2160
+Scenario: stored packages are not removed by following transaction even with keepcache=false
+   When I execute dnf with args "install top-a-1.0 --store ./transaction"
+   Then file "/{context.dnf.tempdir}/transaction/packages/top-a-1.0-1.x86_64.rpm" exists
+   When I execute dnf with args "install top-a-1.0 --setopt=keepcache=false"
```
m-blaha, line 296: Shouldn't the `setopt=keepcache=false` be used with the first install command (the one with --store). It's this transaction that writes `temporary_files.toml` with paths to be purged.

kontura: I suppose that would also make sense.

I added it for the second install command because it is the one that did the removing based on the `keepcache` option: https://github.com/rpm-software-management/dnf5/blob/main/libdnf5/base/transaction.cpp#L1261

But you are correct that there are two parts to this, one writes the packages and the other reads and removes them.

I will add one more test for this. :+1:

m-blaha: Thank you.

m-blaha (approved):

Issue #1751, opened 2025-09-03
Switch nighly builds workflow to use `packit`


kontura:
The nightly builds workflow should be the last thing that actively uses `rpm-gitoverlay`: https://github.com/rpm-software-management/ci-dnf-stack/blob/main/.github/workflows/nightly.yml#L40

I think we should switch it to use `packit` like the rest of our CI if possible.

`rpm-gitoverlay` will require more maintenance in f43 due to breaking changes in its dependencies.
We should also clean up old overlays.

kontura:
Resolved by: https://github.com/rpm-software-management/ci-dnf-stack/pull/1771

PR #1763, opened 2025-10-06
Do not build libpkgmanifest on RHEL 9

ppisar:
It requires yaml-cpp ≥ 0.7.0 and swig ≥ 4.2.0 which do not exist in RHEL 9. The team decided on 2025-10-06 meeting that libpkgmanifs won't be supported there.

### DIFF
--- a/overlays/dnf-ci/overlay.yml
+++ b/overlays/dnf-ci/overlay.yml
@@ -8,6 +8,14 @@ components:
     version-from: git
     git:
       src: github:rpm-software-management/libpkgmanifest.git
+    distgit-overrides:
+      - chroots:
+          - epel-10-x86_64
+          - fedora-eln-x86_64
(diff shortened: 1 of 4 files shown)


kontura:
I believe this will not work.
The `distgit-overrides` are used to override `distgit` config for a particular `chroot` it doesn't disable the rest (it also requires more mandatory settings).

Looking into `rpm-gitoverlay` I don't think there is a way to disable particular `chroot(s)` only for some component(s).
Hopefully we will get more flexibility after switching to packit: https://github.com/rpm-software-management/ci-dnf-stack/issues/1751.

I did look into the nightly builds and it seems maybe half? of the components are also failing on `centos-stream+epel-next-9-x86_64 ` and `epel-9-x86_64 `. Perhaps we could disable those `chroots` for the whole COPR project?

Issue #1715, opened 2025-08-13
behave >= 1.3.0  dies on "from .rpm import normalize_epoch": KeyError: "'__name__' not in globals"


ppisar:
dnf/steps/lib/dnf.py does:

~~~~
from __future__ import absolute_import
from __future__ import print_function

import re


from .rpm import normalize_epoch
from .rpm import RPM
~~~~

When running ci-dnf-stack on RHEL 10, behave is not installed from RPM, it is installed from PyPi.

Since the beginning of August, 2025 the "behave dnf" command dies:

~~~~
# dnf behave
Unknown argument "behave" for command "dnf5". Add "--help" for more information about the arguments.
It could be a command provided by a plugin, try: dnf5 install 'dnf5-command(behave)'
[root@e6c0d0d4ab9d dnf-behave-tests]# behave dnf
USING RUNNER: behave.runner:Runner
Exception KeyError: "'__name__' not in globals"
Traceback (most recent call last):
  File "/usr/local/bin/behave", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/local/lib/python3.14/site-packages/behave/__main__.py", line 291, in main
    return run_behave(config)
  File "/usr/local/lib/python3.14/site-packages/behave/__main__.py", line 113, in run_behave
    failed = runner.run()
  File "/usr/local/lib/python3.14/site-packages/behave/runner.py", line 1138, in run
    return self.run_with_paths()
           ~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.14/site-packages/behave/runner.py", line 1143, in run_with_paths
    self.load_step_definitions()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.14/site-packages/behave/runner.py", line 1130, in load_step_definitions
    load_step_modules(step_paths)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/behave/runner_util.py", line 596, in load_step_modules
    exec_file(os.path.join(path, name), step_module_globals)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/behave/runner_util.py", line 567, in exec_file
    exec(code, globals_, locals_)
[...]

PR #1775, opened 2025-10-27
Behave 3.0.0 compatibility

kontura:
Closes: https://github.com/rpm-software-management/ci-dnf-stack/issues/1715


ppisar:
Thanks for the patch. It looks great. The only issue is that dnf5-daemon tests fail because something (dnf5daemon-client?) cannot find a D-Bus socket:

~~~~
Failed to open bus (No such file or directory)
Is D-Bus daemon running?
~~~~

PR #1779, opened 2025-11-04
Rotating logger preserves file mode and ACL

ppisar:
This checks that the rotating logger preserves file mode and file access control list which a user set on the log file before executing dnf.

It enables a test for the file mode and adds a new test for ACL.

Because log rotation only works outside of installroot, it adds no_installroot tag to the tests.

Because behave fails to execute multiple no_installroot scenarios from one feature file, it moves them into separate files.

For: https://github.com/rpm-software-management/dnf5/issues/1820
For: https://github.com/rpm-software-management/dnf5/issues/2487
Requires: https://github.com/rpm-software-management/dnf5/pull/2494

### DIFF
--- a/dnf-behave-tests/common/cmd.py
+++ b/dnf-behave-tests/common/cmd.py
@@ -7,7 +7,7 @@ import behave
 import glob
 import os
 
-from common.lib.cmd import assert_exitcode, run_in_context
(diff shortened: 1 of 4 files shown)


kontura (approved):
Thanks!

PR #1787, opened 2025-11-20
dnf4: Rotating logger preserves ACL

ppisar:
Upstream commit: f467a2df6e5b9fb5d7b72b65bda55e9712a9c76c

This checks that the rotating logger preserves file access control list which a user set on the log file before executing dnf.

It also moves a test for the file mode to better place.

For: https://github.com/rpm-software-management/dnf/issues/2279

### DIFF
--- a/dnf-behave-tests/common/cmd.py
+++ b/dnf-behave-tests/common/cmd.py
@@ -7,7 +7,7 @@ import behave
 import glob
 import os
 
-from common.lib.cmd import assert_exitcode, run_in_context
+from common.lib.cmd import assert_exitcode, run, run_in_context
 from common.lib.file import prepend_installroot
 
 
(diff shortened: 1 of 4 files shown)


dnf-behave-tests/common/cmd.py
```diff
@@ -60,6 +60,19 @@ def file_has_mode(context, filepath, octal_mode_str):
     assert oct(octal_mode) == oct(octal_file_mode), \
         "File \"{}\" has mode \"{}\"".format(matched_files[0], oct(octal_file_mode))
 
+@behave.step("file \"{filepath}\" has ACL entry \"{entry}\"")
```
dcantrell: Rather than using getfacl, what about using pylibacl?  python-pylibacl package.  For example:

```
def file_has_acl_entry(context, filepath, entry):
    import posix1e
    filepath = prepend_installroot(context, filepath)
    acl = posix1e.acl(file=filepath)
    for line in acl.to_any_text().split("\n"):
        if line == entry:
            return
    raise AssertionError("File \"{}\" has ACL:\n{}".format(filepath, acl.to_any_text()))
```

PR #1803, opened 2025-12-09
Enable and update tests for marking packages with reason GROUP

pkratoch:
Requires: https://github.com/rpm-software-management/dnf5/pull/2549


jrohel:
Are the duplicate lines intentional?
```
153  When I execute dnf with args "mark group dnf-ci-testgroup lame"
154  Then the exit code is 0
155  When I execute dnf with args "mark group dnf-ci-testgroup lame"
156  Then the exit code is 0
```

PR #1807, opened 2026-01-06
fix: reorder packages after adjusting the sorting

mfocko:
Caused by rpm-software-management/dnf5#2522

Asking for review from
- @evan-goode since you were merging the original PR that is suspected as the cause of the issue
- @jrohel as you were, IIRC, working recently on the changes related to vendors


jlebon:
Just happened to see this. https://github.com/rpm-software-management/dnf5/pull/2522 should only have an effect if `SOURCE_DATE_EPOCH` is defined when running `dnf`. Is that the case in this testsuite?

kontura:
> Just happened to see this. [rpm-software-management/dnf5#2522](https://github.com/rpm-software-management/dnf5/pull/2522) should only have an effect if `SOURCE_DATE_EPOCH` is defined when running `dnf`. Is that the case in this testsuite?

It is not, we set it only for some `rpmbuild` runs.

PR #1810, opened 2026-01-12
protected-packages: test protected dependencies

kontura:
Requires: https://github.com/rpm-software-management/libdnf/pull/1739


kontura:
> Is this missing the files for the `dependency-chain` repo? I'm getting `No match for argument: top`

Oh, yes, fixed, thanks!

PR #1812, opened 2026-01-13
dnf-4-stack: Do not use installroot in dnf-automatic/command_email.feature

ppisar:
The test sometimes failed because dnf-automatic sometimes did not print any output.

The cause was that the test added in commit
e942bcf8c045098243fc192faa211669f1d85299 ("dnf-automatic: email_command: Pass recipients as separate arguments to a command") used an installroot to avoid clobbering host system. But dnf-automatic does not have --installroot option, causing dnf-automatic to always operate on the host system, disregarding what happened in the installroot:

If the host system had a pending update, dnf-automatic emitted an notification and thus the test passed.  On the other hand, if the host had no pending update, there was no notification and the test failed.

This patch fixes it by disabling installroot.


m-blaha (approved):
Thank you!

PR #1816, opened 2026-01-19
Packit: checkout release-5.2.x.x branch

evan-goode:
Packit always checks out the main branch here, even when run from the DNF5 repository with `fmf_ref: release-5.2.x.x`. So I suppose we need to amend the `main.fmf` here. Though maybe there is a better way to get the correct branch name other than hardcoding it here?

See also https://github.com/rpm-software-management/dnf5/pull/2577/files.


dcantrell (approved):
I don't have a better suggestion other than hardcoding the branch name.

PR #1820, opened 2026-01-28
fix: enable the regression test

mfocko:
Related to RHEL-81778
Related to RHEL-81779

Merge with rpm-software-management/libdnf#1742


ppisar:
> I reported the persistency error as #1744. Without fixing it we should not merge this to prevent from introducing false negatives.

A possible workaround is removing the `Persistence` line from the match table.

PR #1823, opened 2026-01-30
autoremove: switch unneeded protected package msg to stderr

kontura:
Requires: https://github.com/rpm-software-management/dnf5/pull/2589


evan-goode (approved):
Passing with https://github.com/rpm-software-management/dnf5/pull/2589.

PR #1837, opened 2026-03-03
Test that offline and stored transactions use cached packages

kontura:
Requires: https://github.com/rpm-software-management/dnf5/pull/2634


m-blaha (approved):
Thank you!

PR #1838, opened 2026-03-03
Do not install libdnf5-plugin-local on RHEL >= 10

ppisar:
Building the plugin is intentionally disabled for RHEL >= 10 in dnf5.spec.


mfocko:
Is that also related to the Fedora CI issue with the tests? Or it was just installability?

PR #1839, opened 2026-03-03
Adapt behave tests for allow_vendor_change

fhbash:
Adapt behave tests for dnf5 `allow_vendor_change=false` default


Conan-Kudo (approved):
Since we haven't changed the default just yet, these tests need to be adapted to test both false and true cases properly

fhbash:
/packit test https://github.com/rpm-software-management/dnf5/pull/2641

evan-goode (approved):
LGTM

Issue #1845, opened 2026-03-11
Testing on fedora:45 fails due to missing signatures


kontura:
Since https://fedoraproject.org/wiki/Changes/Enforcing_signature_checking_by_default rpm requires signatures by default but we don't sign most packages.

Should we start signing all packages (except when the test requires not signed)?

kontura:
We also need to backport the fix to dnf4.

kontura:
This should be now resolved both for dnf4 and dnf5.

PR #1857, opened 2026-03-31
[dnf-4-stack] Automatically sign all packages

kontura:
This is a backport of https://github.com/rpm-software-management/ci-dnf-stack/pull/1852

Closes: https://github.com/rpm-software-management/ci-dnf-stack/issues/1845

Unfortunately while the commits are similar in approach most of them had to be updated due to difference with the `main` (dnf5) branch.
Another complication is that dnf4 `--nogpgcheck` doesn't turn off rpm verification, it has to be disabled manually by setting the `%_pkgverify_level` macro.


evan-goode (approved):
Thank you, everything LGTM!

Tests are passing on rawhide.

PR #1858, opened 2026-04-07
Packit+TMT on dnf-4-stack

evan-goode:
These patches set up Packit to test PRs on dnf-4-stack similar to the main branch. I also fixed some miscellaneous things which should allow the tests (including bootc tests) to be run on any supported Fedora version or CentOS 9 Stream or CentOS 10 Stream. Marking as a draft for now to test Packit.

dnf, libdnf, and dnf-plugins-core changes will also be needed.


evan-goode:
I experimented with enabling the tests for c9s and c10s, but the dnf-4-stack builds and/or tests are broken on CentOS for various reasons. We can keep them disabled on the dnf-4-stack branch and enable them on the rhel-x.y branches.

m-blaha:
The only thing I've noticed is that there is a new `behave-createrepo_c.fmf` integration plan, but it's run. Is that intentional?

evan-goode:
Thanks for looking.

> but it's run

Assume you mean it's not run? `.packit.yaml` did not run it but if you ran TMT manually I suppose it might.

Yes, good eye, that was carried over from the `enable-tmt-dnf-4-stack` but was seemingly not used anywhere. The createrepo_c repo runs tests from the main branch of ci-dnf-stack, and dnf, libdnf, dnf-plugins-core don't run them.

I removed it since we only need those tests on the main branch. But I think we'll want to add it back when backporting this PR to the rhel-x.y branches.

m-blaha (approved):
Thank you!

PR #1862, opened 2026-04-22
Adds explicit "repo_gpgcheck=0" into generated repositories

fhbash:
Adds explicit "repo_gpgcheck=0" into generated repositories


ppisar (approved):
Good.

PR #1864, opened 2026-04-24
Add swap-conflicts.feature

janblazek:
Test swap of conflicting packages that both
provide a dependency for another package.
This shouldn't cause a downgrade of the higher
level package.

Related to RHEL-141449


m-blaha (approved):
Works as expected, thank you!

PR #1873, opened 2026-04-30
repoinfo/repolist -v: test unique package count

kontura:
Requires: https://github.com/rpm-software-management/dnf/pull/2324


evan-goode (approved):
Tests pass locally on Rawhide.

PR #1875, opened 2026-05-05
Normalize percent-encoding in "HTTP log contains"

m-blaha:
libcurl started to normalize percent-encoding hexadecimal digits to uppercase (following RFC 3986 section 2.1) (see
https://github.com/curl/curl/commit/0b4ebebb06b3148a234a5a7bc95f8253dd6cb8df).

This change makes HTTP log comparison case insensitive with regard to percent-encoded parts.

This change should fix CI stack failures on Fedora rawhide (e.g. https://github.com/rpm-software-management/dnf5/pull/2724/checks?check_run_id=74253503641)

### DIFF
--- a/dnf-behave-tests/dnf/steps/server.py
+++ b/dnf-behave-tests/dnf/steps/server.py
@@ -7,6 +7,7 @@ import behave
 from fnmatch import fnmatch
 import os
 import parse
+import re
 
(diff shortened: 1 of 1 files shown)


kontura (approved):
Thank you!

PR #1876, opened 2026-05-05
bootc: Port framework and tests from dnf-4-stack

evan-goode:
Implement a framework for running tests on bootc systems, and add tests for `dnf5 --transient`

This is mostly a straightforward merge of the bootc testing framework from the dnf-4-stack branch to the main DNF5 branch.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -13,6 +13,14 @@ jobs:
 
   - job: tests
     trigger: pull_request
+    identifier: "dnf5-bootc-tests"
+    targets:
+      - fedora-all
+    skip_build: true
+    tmt_plan: "^/plans/integration/behave-dnf5-bootc/"
+
+  - job: tests
+    trigger: pull_request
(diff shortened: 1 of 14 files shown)


evan-goode:
/packit test https://github.com/rpm-software-management/dnf5/pull/2727

evan-goode:
The CI failures on F42 and Rawhide are due to Packit trying to install RPMs it shouldn't. On F42, it tries to install dnf5-plugin-manifest (fails because libpkgmanifest is not available on F42), and on Rawhide, it tries to install dnf5 from both the regular build and the WITH_MODULEMD=OFF build. The CI on the dnf5 PR runs correctly (although without this ci-dnf-stack PR), the problem only happens when testing across two PRs like this. https://github.com/packit/packit-service/issues/1720 might be related. It shouldn't matter once we move the DNF5 integration tests under the dnf5 repo.

PR #1881, opened 2026-05-12
reposync: test --min-buildtime option

kontura:
requires: https://github.com/rpm-software-management/dnf5/pull/2733


ppisar (approved):
Good.

Directly setting _buildtime macro to the desired UNIX time would be shorter, but your code is without private underscored names.

PR #1884, opened 2026-05-27
Add test for RHEL-80156: non-modular advisory filtered for modular packages

m-blaha:
Add a behave scenario that verifies 'updateinfo --list' does not show non-modular advisories when a modular stream of the same package is enabled.

Requires: https://github.com/rpm-software-management/libdnf/pull/1771

### DIFF
--- a/dnf-behave-tests/fixtures/specs/dnf-ci-fedora-non-modular-advisory/updateinfo.xml
+++ b/dnf-behave-tests/fixtures/specs/dnf-ci-fedora-non-modular-advisory/updateinfo.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<updates>
+  <update from="secresponseteam@foo.bar" status="final" type="bugfix" version="3">
+    <id>NONMOD-2024-9291</id>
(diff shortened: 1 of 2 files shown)


dnf-behave-tests/fixtures/specs/dnf-ci-fedora-non-modular-advisory/updateinfo.xml
```diff
@@ -0,0 +1,20 @@
...
+    <issued date="2024-09-29 00:00:00"/>
+    <updated date="2024-09-29 00:00:00"/>
+    <severity>none</severity>
+    <summary>Non-modular nodejs advisory</summary>
+    <description>Bug fix for non-modular nodejs</description>
+    <pkglist>
+      <collection short="named collection">
+        <name>Non-modular component</name>
+        <package name="nodejs" version="5.12.1" release="20.fc29" epoch="2" arch="x86_64">
+          <filename>nodejs-2:5.12.1-20.fc29.x86_64.rpm</filename>
```
kontura, line 15: I think we need to add this nodejs version rpm package so it can be excluded by module filtering.

PR #1886, opened 2026-05-28
context libdnf (microdnf): $arch variable set and substituted

kontura:
requires: https://github.com/rpm-software-management/libdnf/pull/1772


jrohel (approved):
LGTM

PR #1891, opened 2026-06-10
Adjust copr test for the latest curl

m-blaha:
Curl changed the format of an error message in its latest release (see curl
upstream commit bc40e09f638). Adjust the test to cover both variants.


ppisar (approved):
I confirm it fixes the test with curl-8.21.0~rc1-1.fc45.

PR #1892, opened 2026-06-10
Adjust download-source and repo-sync tests for the latest curl

m-blaha:
Curl changed the format of an error message in its latest release (see curl upstream commit bc40e09f638). Adjust the tests to cover both variants.


ppisar (approved):
Good. I verified that it restores a compatibility with curl-8.21.0~rc1-1.fc45.
The dnf/encoding.feature:85 ("non-UTF-8 character in an option when using corresponding locale") failure is unrelated, probably triggered by Python 3.15.0b2.

ppisar:
> The dnf/encoding.feature:85 ("non-UTF-8 character in an option when using corresponding locale") failure is unrelated, probably triggered by Python 3.15.0b2.

#1893

PR #1897, opened 2026-06-12
Update to f43

kontura:


ppisar (commented):
Maybe we should also remove "42" tag from .github/workflows/container-build-quay-io.yml file.

kontura:
> Maybe we should also remove "42" tag from .github/workflows/container-build-quay-io.yml file.

Removed, thanks.

ppisar (approved):
Good.

PR #1898, opened 2026-06-15
Adapt tests to DNF5 in ELN

ppisar:
This patch set makes the tests passing in ELN (RHEL 11). You can see the addressed failures in <https://github.com/rpm-software-management/dnf5/pull/2669> (results at <https://artifacts.dev.testing-farm.io/4e3b8ee4-6499-4a18-948c-d083bfa300e6/>).

This ci-dnf-stack patch should be merged before merging the linked dnf5 merge request.

### DIFF
--- a/dnf-behave-tests/dnf/protected-dnf.feature
+++ b/dnf-behave-tests/dnf/protected-dnf.feature
@@ -1,0 +1,13 @@
+# - doesn't makes sense to test in installroot
+# - a potentially destructive test
+@no_installroot
+Feature: Dnf protects itself
+
+
+@not.with_os=rhel__lt__11
+Scenario: Dnf5 when installed protects itself
+  Given I execute dnf with args "install dnf5"
+   Then the exit code is 0
+   When I execute dnf with args "remove dnf5"
+   Then the exit code is 1
+    And stderr contains "operation would result in removing the following protected packages: dnf5"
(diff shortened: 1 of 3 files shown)


dnf-behave-tests/dnf/protected-dnf.feature
```diff
@@ -0,0 +1,14 @@
+# - doesn't makes sense to test in installroot
+# - a potentially destructive test
+@no_installroot
+Feature: Dnf protects itself
+
+
+@not.with_os=rhel__lt__11
+Scenario: Dnf5 when installed protects itself
+  Given I use repository "dnf-ci-fedora"
+   When I execute dnf with args "install dnf5"
```
kontura: `dnf-ci-fedora` has only the `yum` package not `dnf5`, `dnf-ci-fedora` is not used and we could drop it, so this checks if the host has dnf5 from its own repositories right?

ppisar: Yes, it checks whether the host's configuration protects the package manager.  I will remove the use of "dnf-ci-fedora" repository from this test.

(A generic test for transitive protection already exists in "Dependency of a protected package cannot be removed" test.)

ppisar: Done.

kontura (approved):

dnf-behave-tests/dnf/protected-dnf.feature
kontura: Thanks

PR #1899, opened 2026-06-16
Pin pyftpdlib to 2.1.0

janblazek:
Newer pyftpdlib 2.2.0 requires newer python 3.9.


ppisar (approved):
I confirm that pyftpdlib 2.2.0 does not work with distribution wheel, while 2.1.0 does.

PR #1901, opened 2026-06-17
Add more test suite dependencies

ppisar:
Some tests failed on ELN (based on F45) with:

      ASSERT FAILED: Running command "id -u testuser || useradd testuser" failed: 127

The cause was that the /usr/bin/useradd program was not available because shadow-utils package was not listed in the test suite dependencies.

This patch declares RPM dependencies for all the commands executed by run_in_context() function.

### DIFF
--- a/dnf-behave-tests/requirements.spec
+++ b/dnf-behave-tests/requirements.spec
@@ -17,23 +17,39 @@ License:        GPLv3
 
 # test suite dependencies
 BuildRequires:  acl
+# attr for "getfattr" command
 BuildRequires:  attr
+# coreutils for "chmod", "head', "id", "ln", "ls", "mv", "sha256sum", "sleep",
+# "sort", "stat", "touch" commands
+BuildRequires:  coreutils
 BuildRequires:  createrepo_c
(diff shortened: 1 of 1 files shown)


kontura (commented):
I have also noticed we used `sed` (for example here: [link](https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/plugins-core/manifest.feature#L98)) and `diff` ([link](https://github.com/rpm-software-management/ci-dnf-stack/blob/main/dnf-behave-tests/dnf/plugins-core/manifest.feature#L57)), should we also add those?

Even better there seems to be only one `diff` perhaps we could replace it with `And the files "{context.dnf.tempdir}/real-checksum" and "{context.dnf.tempdir}/manifest-checksum" do not differ`?

ppisar:
I missed the "I execute" step. I will amend the dependencies.

ppisar:
Done.

kontura (approved):

kontura:
Thank you!

PR #1903, opened 2026-06-23
Disable modularity tests for DNF5 since RHEL 11

ppisar:
DNF5 will be compiled without modularity support there.

This patch skips building modular packages, adding modulemd.yaml into a repository, and performing modularity tests.

For and requires: https://github.com/rpm-software-management/dnf5/pull/2786

Test results are affected by this <https://github.com/rpm-software-management/dnf5/issues/2797> RPM-6.1/DNF5 regression.

Question1: Do we want to replace the os-release conditions (`@not.with_os=rhel__lt__11`) with a new Behave tag (e.g. `@with_modularity`). Because of dnf/option-hint.feature we would also need a negative form (`@without_modularity`).

Question2: Does `command specific long option used with root dnf command` scenario in dnf/option-hint.feature test explicit list of commands intentionally? Or should be replace the exact list of commands with a glob? That would allow us to shrink the two scenarios into one regardless of modularity support. We could have similar problem in the future when we enable/disable other features differently on different distributions.

### DIFF
--- a/dnf-behave-tests/common/lib/os_version.py
+++ b/dnf-behave-tests/common/lib/os_version.py
@@ -20,3 +20,10 @@ def detect_os_version():
         major_version = str(rhel_macro_version)
 
     return os_id + "__" + major_version
+
+# Whether the current operating systems is supposed to support modularity.
+def want_modularity():
+    os, version = detect_os_version().split('__')
+    if os == 'rhel' and int(version) >= 11:
+        return False;
+    return True;
--- a/dnf-behave-tests/dnf/command-aliases.feature
+++ b/dnf-behave-tests/dnf/command-aliases.feature
@@ -43,7 +43,6 @@ Examples:
         | list                | ls                           |
         | makecache           | mc                           |
         | mark                | mark                         |
-        | module              | module                       |
         | reinstall           | rei                          |
         | reinstall           | reinstall                    |
         | remove              | remove                       |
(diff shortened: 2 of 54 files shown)


ppisar:
I'm opening this pull request for a review because it basically works.
Current failures are caused by changes in RPM and by CI mixing F45 modular and nonmodular builds together.

ppisar:
I corrected the `@use.with_os=...` tag in dnf-behave-tests/dnf/option-hint.feature.

kontura:
> Question1: Do we want to replace the os-release conditions (`@not.with_os=rhel__lt__11`) with a new Behave tag (e.g. `@with_modularity`). Because of dnf/option-hint.feature we would also need a negative form (`@without_modularity`).

Personally I kind of like that all information is directly visible now however I also wouldn't mind a new `@with_modularity` tag. I am impartial about this.


> Question2: Does `command specific long option used with root dnf command` scenario in dnf/option-hint.feature test explicit list of commands intentionally? Or should be replace the exact list of commands with a glob? That would allow us to shrink the two scenarios into one regardless of modularity support. We could have similar problem in the future when we enable/disable other features differently on different distributions.

I would like to keep checking that at least some commands do get listed but otherwise I am fine with a glob.
Perhaps something like:
```gherkin
    And stderr matches line by line                                  # common/output.py:163
      """
      Unknown argument "--skip-broken" for command "dnf5". Add "--help" for more information about the arguments.
      The argument is available for commands: .*, install, do. \(It has to be placed after the command.\)
      """
```
since the `install` and `do` commands likely are not going away?

kontura (approved):

ppisar:
Thanks for your opinion. For now I will keep this pull request as it is and merge it. We can optimize it later.

PR #1911, opened 2026-07-09
Expect clean_command.lock in cachefiles

kontura:
Requires: https://github.com/rpm-software-management/dnf/pull/2340


m-blaha (approved):
Thanks!

PR #1915, opened 2026-07-15
scriptlets.feature: expect lua scriptlets output on stderr

kontura:
dnf5 prints scriptlet output to stderr, rpm no longer hard-codes lua scriptles output to stdout since:
https://github.com/rpm-software-management/rpm/pull/4022 (rpm-6.0.91 - f45)

### DIFF
--- a/dnf-behave-tests/dnf/scriptlets.feature
+++ b/dnf-behave-tests/dnf/scriptlets.feature
@@ -9,13 +9,16 @@ Background: Enable repository
     And I set environment variable "FORCE_COLUMNS" to "400"
 
 
+# Disable on older fedoras and rhels because they don't contain rpm-6.0.91
+@not.with_os=fedora__lt__45
+@not.with_os=rhel__lt__11
 Scenario Outline: Install a pkg with a successful scriptlet
    When I execute dnf with args "install <package>"
    Then the exit code is 0
     And Transaction is following
         | Action        | Package                  |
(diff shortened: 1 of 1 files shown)


ppisar:
Are you sure this what DNF5 should do? I would expect DNF5 to report scriptlet output the same way no matter whether the scriptlet is in Lua or in externally interpreted language.

ppisar:
See https://github.com/rpm-software-management/dnf5/issues/2797.

ppisar:
If you want to adapt the tests, you also need to extend the Fedora 45 condition to cover RHEL 11.

kontura:
> Are you sure this what DNF5 should do? I would expect DNF5 to report scriptlet output the same way no matter whether the scriptlet is in Lua or in externally interpreted language.

I would also expect that, isn't that the case with the fixed rpm? I thought dnf5 prints all scriptlet output to stderr.

Before rpm-6.0.91 it was inconsistent because bash scriptlets were redirected to stderr by dnf5 but lua went straight to stdout due to the rpm bug.

ppisar (approved):
I see. This test change corrects a descriptor where the scriptlet output is expected. That's indeed good with the new RPM.
I was more worried that the output was printed by RPM rather than by DNF5, mixing with the DNF5's progressbar rather than beeing properly quoted with ">>>" prefix by DNF5. But that's probably an independent bug.

PR #1920, opened 2026-07-17
add test for RHEL-80156

janblazek:
Add new scenario: updateinfo list should not show advisories for non-modular packages when modular version is installed

Assisted-by: Claude Sonnet 4.5


m-blaha (dismissed):
Thank you! I'm only wondering, whether the complexity in *.spec (like %install and %files sections) is needed. I think installing a file to %{_bindir} was the reason why AutoReq was needed. But I'm not blocking merge on it, feel free to keep it as it is.

janblazek:
You're right. The files are not needed in the test rpms. I have simplified the spec files.

m-blaha (approved):

PR #1922, opened 2026-07-23
dbus api: Arch changing upgrade scenario

m-blaha:
Add a test scenario for History::recent_changes() involving also arch changing upgrade.

For: https://github.com/rpm-software-management/dnf5/issues/2595


evan-goode (approved):
Thanks!

PR #1938, opened 2026-08-20
Add couple localpkg_gpgcheck tests

kontura:
Just a couple of simple tests I created when testing unrelated stuff.


ppisar (approved):
Good.

PR #1940, opened 2026-08-20
Dockerfile: Copy dnf5 repo configs for dnf4

jmarrero:
On systems where repo configs have moved to `/usr/share/dnf5/repos.d`, dnf4 cannot find them since it only looks in `/etc/yum.repos.d`. This causes the container build to fail with `There are no enabled repositories` when running on newer base images.

Copy the configs so the build steps that use dnf4 can find enabled repositories.

> Assisted-by: AI
> I am knowledgeable in this problem domain and reviewed it carefully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,13 @@ RUN set -x && \
         dnf5 -y install dnf4; \
     fi
 
+# Make dnf5 repo configs available to dnf4 on systems where repos
+# have moved to /usr/share/dnf5/repos.d (Fedora 45+)
+RUN set -x && \
+    if [ -d /usr/share/dnf5/repos.d ]; then \
+        cp -n /usr/share/dnf5/repos.d/*.repo /etc/yum.repos.d/ 2>/dev/null || true; \
+    fi
+
 # enable the test-utils repo
 RUN set -x && \
     dnf4 -y --refresh upgrade; \


Conan-Kudo (approved):

Dockerfile
```diff
@@ -26,6 +26,13 @@ RUN set -x && \
         dnf5 -y install dnf4; \
     fi
 
+# Make dnf5 repo configs available to dnf4 on systems where repos
```
ppisar, line 29: This block should go before `# Copy extra repo files` line so that the _extra repo files_ can replace system repositories if needed.

Dockerfile
```diff
@@ -26,6 +26,13 @@ RUN set -x && \
...
+# have moved to /usr/share/dnf5/repos.d (Fedora 45+)
+RUN set -x && \
+    if [ -d /usr/share/dnf5/repos.d ]; then \
+        cp -n /usr/share/dnf5/repos.d/*.repo /etc/yum.repos.d/ 2>/dev/null || true; \
```
ppisar, line 33: Use `--update=none` instead of `-n`. `-n` is deprecated.

ppisar (changes requested):
This fix is dubious. We should first decide whether DNF4 should support /usr/share/dnf5/repos.d or not. Or we should simply use system DNF in the Dockerfile exclusively instead of using DNF4. If will raise this issue on the team meeting.

ppisar:
An example of the failure this pull request tackles <https://artifacts.dev.testing-farm.io/c67ef616-0e69-4740-bf01-e1c78eb4344d/work-behave-dnfmntfly9d/log.txt>.

ppisar:
DNF team discussed this issue and decided to add a proper support for /usr/share/dnf5/repos.d into libdnf <https://github.com/rpm-software-management/libdnf/issues/1782>, rather then working around it like in this ci-dnf-stack pull request.

PR #1943, opened 2026-08-24
plugins-path: Remove dependency on particular plugin

m-blaha:
With new systemd-inhibit plugin, which config file orders alphabetically before the action plugin, this test started to fail:

Last Command stderr:
Cannot load libdnf plugin enabled from: /etc/dnf/libdnf5-plugins/00-systemd-inhibit.conf
 Cannot find plugin library "/tmp/dnf_ci_installroot_8mc9n1mt/test/plugins/systemd-inhibit.so"


kontura (approved):
Thank you.

PR #1944, opened 2026-08-31
Removing dnf5 integration tests

fhbash:
Removing `dnf5` integration tests since we moved all the tests to dnf5 repository: https://github.com/rpm-software-management/dnf5/tree/main/integration-tests


m-blaha (approved):
Thanks!

